PR #65, opened 2017-11-22
Add reference to mycete

qbit:


qbit:
Ah crud, looks like #59 does the same thing as my second commit.

richvdh:
thanks!

PR #80, opened 2018-01-12
Replace broken link to Jira with link to GitHub

jooert:
As far as I'm aware, Jira is no longer being used and the link to it in the FAQ returns 404. So, this removes the link and instead adds a link to the GitHub organization.


richvdh:
thanks!

PR #115, opened 2018-03-27
Update fractal client information

danigm:
Fractal has been moved to the gnome official group World, so the
repository has changed.

I've also updated the screenshot and the client maturity because the
client changes a little from the last update.


richvdh:
thanks!

PR #136, opened 2018-04-17
Changed Dockerfile location after handover

mvgorcum:
The dockerfile is now maintained by aventer and the images are at a different location on docker hub.


ara4n:
thanks!

PR #159, opened 2018-05-14
I'd like to be a bigproject as well

f0x52:
Benpa told the categorization was arbitrary, so here's my pr


benparsons:
I'll review the big list before long. (Just acknowledging that this PR is not simply being ignored!)

There are a couple of changes I want to make to that section. I also want to add, the list was not arbitrary, it was the best assessment I could make as of April 2018.

f0x52:
whoops, I should've branched these new commits

richvdh:
@benparsons @f0x52 what's the state of play here?

benparsons:
@f0x52 I'm going to close this without merging for now. Let's chat again when the new neo is ready to show.

PR #183, opened 2018-06-24
Synchronize project list

turt2live:
I've split them out into individual commits, although the diff isn't all that bad (just massive).

Notable changes:
* Updated metadata on my projects
* Added some unlisted projects
* Removed topic bot - This is a silly bot and in my opinion silly bots just clutter the page. Considering they add little value beyond comedy or simple demonstrations, I'm okay with them not being on Try Matrix Now.


Half-Shot:
...wow

benparsons:
wow is right. checking and will merge asap

PR #209, opened 2018-08-16
r0.1.0 release of the Push Gateway specification

turt2live:
Paired with https://github.com/matrix-org/matrix-doc/pull/1521


richvdh:
I'm assuming the massive gob of html is just what's generated from the current state of matrix-doc, in which case, this lgtm, modulo getting matrix-org/matrix-doc#1521 landed

turt2live:
Yup, just a generated pile of html. Will fixup the matrix-doc side

PR #217, opened 2018-08-30
r0.1.0 release of the Application Service specification

turt2live:
Paired with https://github.com/matrix-org/matrix-doc/pull/1630


unstyled_docs/spec/application_service/latest
```diff
@@ -0,0 +1 @@
+r0.1.0.html
```
richvdh: should this not be `latest.html` rather than `latest`?

turt2live: oops, yup. Will fix.

PR #228, opened 2018-09-10
add warnings to old spec versions

benparsons:
this adds warning to old spec versions (currently the only "old" spec versions are CS)


ara4n:
this lgtm, but is there a way to automatically update older spec versions otherwise we're never going to remember to do it?

benparsons:
How are the older spec versions generated? I couldn't see and thought they were cut manually?

PR #232, opened 2018-09-15
invalid json

nico202:
2 quotes were missing


richvdh:
@nico202: thanks for this. It's also addressed in the larger rework in #201, so rather than merge this and cause conflicts, I'm going to close it and hope that #201 lands soon. Thanks though!

PR #235, opened 2018-09-21
Fix Title in SimpleMatrix project file

MTRNord:
It seems like it worked on Github but on the Matrix Page the Title did not get rendered as a Title. I think the reason is that I missed these 2 whitespaces. Sorry for that


benparsons:
Good spot

PR #238, opened 2018-09-23
Fix indentation of application services

aaronraimist:
The indentation in this file caused some pain https://matrix.to/#/!HsxjoYRFsDtWBgDQPh:matrix.org/$1537722418907037aOqoN:matrix.org

PR #241, opened 2018-09-24
Fix application services guide to use kramdown style code blocks

aaronraimist:
Better #238

### DIFF
--- a/jekyll/_posts/guides/2015-08-21-application_services.md
+++ b/jekyll/_posts/guides/2015-08-21-application_services.md
@@ -63,84 +63,94 @@ A very basic application service may want to log all messages in rooms which hav
 
 Here's an example of a very basic application service using Python (with Flask and Requests) which logs room activity:
 
-        # app_service.py:
+~~~ python
+# app_service.py:
 
-        import json, requests  # we will use this later
-        from flask import Flask, jsonify, request
-        app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


richvdh:
thanks, let's try it!

richvdh:
well, that looks a lot better! https://matrix.org/docs/guides/application_services.html

richvdh:
(and yes, it is a weird format. I rather wonder if we shouldn't instead be configuring kramdown to read GFM instead of its own funny dialect)

aaronraimist:
Looks better but now you can't copy the examples without also getting the line numbers.

richvdh:
ugh. yes.

aaronraimist:
You could change this: coderay_line_numbers

https://kramdown.gettalong.org/rdoc/Kramdown/Options.html

aaronraimist:
or just get rid of the syntax highlighting, although it think it looks good

richvdh:
I don't seem to be able to get any changes in _config.yml to work, and I've now spent more time on this than I really wanted to :/

aaronraimist:
Yeah just go ahead and remove the "yaml" and "python" bits from this patch and the syntax highlighting will go away along with the line numbers.

richvdh:
[I suspect that the fact that the versions of jekyll and kramdown on our build server are prehistoric doesn't help anything]

PR #257, opened 2018-10-23
Update template.md with all new fields

Matrixcoffee:


Matrixcoffee:
Hmm, on reviewing it, should the screenshot be uploaded and locally referenced instead of pointing to a remote resource?

Issue #221, opened 2018-09-02
Remove https://github.com/kamax-io/mxhsd/


dax-dev:
The project decided to abandon the Matrix protocol in favor of their own.

See their [README](https://github.com/kamax-io/mxhsd/blob/master/README.md#current-status)

benparsons:
@maxidor to confirm. Max you'd like mxhsd removed from try-matrix-now?

maxidor:
@benparsons Please don't remove it. It is currently still aligned with the latest reverse-engineering and we might still make it compatible with the first S2S spec release if accurate enough.

dax-dev:
Sorry for assuming then -- I was confused because the README seemed to imply that the project was moving away from Matrix.

ara4n:
@dax-dev np - thanks for trying to help. unfortunately the situation there is a mess.

PR #265, opened 2018-10-30
Mark mxhsd as no longer maintained

MayeulC:
The following commit makes that clear:
https://github.com/kamax-matrix/mxhsd/commit/43558deb59c6ed64b02ab61788bc33f54ceea6ce

It might still be interesting to preserve the "Early Beta" state somewhere, although I expect this to become less and less true with time.

### DIFF
--- a/jekyll/_posts/projects/2018-04-24-mxhsd.md
+++ b/jekyll/_posts/projects/2018-04-24-mxhsd.md
@@ -4,7 +4,7 @@ title: mxhsd
 categories: projects server
 description: mxhsd is Matrix Homeserver aimed towards entities who want to have in-depth control of their servers
 author: Max Dor
-maturity: Early Beta
+maturity: No longer maintained
 language: Java
 license: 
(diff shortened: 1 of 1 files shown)


MTRNord:
I think this comment by @maxidor still applies: https://github.com/matrix-org/matrix.org/issues/221#issuecomment-418074244

ara4n:
yeah, the situation there is messy and unconstructive; it's probably best to let the author decide whether his project should be listed or not. thanks for trying to assist.

MayeulC:
Oh, my bad, sorry, I didn't realise there had been a PR for this already. Ruma is another project I was following, for which I'm not completely sure of the status. But hopefully everything will end up being sorted out in the (near) future!

dax-dev:
just a note here that the README now literally says it has been abandoned, and the repo is archived

https://github.com/kamax-matrix/mxhsd/commit/710692052f2e349a1ab415246dc5bf04bb46af98

PR #274, opened 2018-11-14
Delete go-avbot

### LINKED ISSUE #208: Hide go-avbot

aaronraimist:
Fixes #208

There isn't a way to hide it but keep it somewhere behind the scenes right?


benparsons:
Let's just delete. The way to "hide" would have been to remove the second category

PR #326, opened 2019-02-01
Release r0.1.0 of the server-server specification

turt2live:
**Please review with https://github.com/matrix-org/matrix-doc/pull/1836**


ara4n:
My only concern here is the rather anticlimatic changelog of:
```
<p class="topic-title first">Version: r0.1.0</p>
<p>No significant changes.</p>
```
Shouldn't we have at least a brief explanation of what's gone into this release?

turt2live:
err, yes, we should. That should say this:

```
r0.1.0
======

This is the first release of the Server Server (Federation) specification.
It includes support for homeservers being able to interact with other
homeservers in a decentralized and standard way.
```

Will fix and merge.

PR #264, opened 2018-10-28
Remove OpenMarket SMS Gateway (#205)

aaronraimist:
Fixes #205

PR #329, opened 2019-02-05
Remove defunct project from try-matrix-now

non-Jedi:
OpenMarket SMS Gateway has not worked in many months. Blog post documents it for
posterity, but keeping it on the try-matrix-now page is misleading. Maybe the old blog post should be edited as well?


aaronraimist:
Note that #264 exists already, sorry

non-Jedi:
Oops

PR #343, opened 2019-03-03
Update 2018-12-27-fluffychat.md

ChristianPauly:
The FluffyChat website has moved to jekyll with localization so I have moved everything in an assets directory.


ara4n:
thanks!

PR #376, opened 2019-04-16
Nginx theme

nadonomy:
Nginx theme, forked from https://github.com/Naereen/Nginx-Fancyindex-Theme, fully responsive, styled to look similar to matrix.org without any external dependencies, or unecessary JS.

Looks a little like:

Screenshot 2019-04-16 at 16 35 07

Screenshot 2019-04-16 at 16 35 15


ara4n:
lgtm, other than removing the remaining .js file (and the search box) for security.

is it impractical to reuse (i.e. reference) precisely the same header+footer+css as gatsby uses? (to avoid them drifting out of sync)

nadonomy:
> is it impractical to reuse (i.e. reference) precisely the same header+footer+css as gatsby uses? (to avoid them drifting out of sync)

@benparsons is still integrating, so should be able to answer this question soon. The nav CSS classes are the same though so should be trivial to switch.

(Other JS removed!)

PR #385, opened 2019-04-23
Update repo link for Pattle

babolivier:
Pattle's sources now live at https://git.pattle.im/pattle/app


turt2live:
Looks like this happened, thanks @babolivier for the PR though!

PR #424, opened 2019-05-12
Add Dart to the language filter so Pattle shows up

aaronraimist:

#404 forgot this


ara4n:
thanks!

PR #426, opened 2019-05-12
Fix the RSS feed so it shows the full post (#408)

### LINKED ISSUE #408: rss feed needs full post contents
rather than just the headers

aaronraimist:
Apparently the html output was just broken on that version of gatsby-mdx. When I update it to version 0.6.2 the RSS feed works but if I update it to the latest version (0.6.3) I get some awful react error when building the site. No idea up with that but at least the RSS feed works.

Also had to remove a random character from one of the blog posts to get the XML to be valid. We may need some way of validating blog posts so those characters don't get added and silently break the feed.

Fixes #408

### DIFF
--- a/gatsby/blog/2019-04-11/security.mdx
+++ b/gatsby/blog/2019-04-11/security.mdx
@@ -122,4 +122,4 @@ was passphrased and based on our initial analysis of the release, we believe it 
 
 <h2 id="toc_9">What are we doing to prevent this in future?</h2>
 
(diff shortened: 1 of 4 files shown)


benparsons:
@aaronraimist thanks for taking a look at this issue! When running this, I see that the `blog/feed` contains HTML-entities inside cdata, rather than the characters themselves. I think we need to have the original characters - right?

benparsons:
Actually, just noticed `<![CDATA[` is missing from the `<content:encoded>` element - could this be the problem?

benparsons:
(previous comments from me are incorrect re XML - see https://stackoverflow.com/a/10099772/384316 for why)

PR #452, opened 2019-06-11
Fix footer links to GitLab and YouTube

aaronraimist:


aaronraimist:
https://github.com/matrix-org/matrix.org/commit/ad8169b864986ef059734999cc3846261d7d21c9#diff-1a4970dc6761f27f640a25aff2095f6e also changed the link to GitHub which causes it to not work when built locally because I guess the redirect is like a 302 redirect instead of a redirect from Gatsby. I left that unchanged for now but it might be good to make those work when built locally as well.

PR #458, opened 2019-06-12
Fix several things

aaronraimist:
Got tired of creating a whole bunch of PRs, sorry there are several things in here. Each commit should be reviewable on its own.

- Twitter cards now look like:
Screen Shot 2019-06-11 at 7 50 05 PM
Screen Shot 2019-06-11 at 7 49 56 PM
instead of
Screen Shot 2019-06-11 at 7 54 35 PM
Still room for improvement for the link to the homepage but it is better than a massive logo with a transparent background.

- Seaglass TMN page fixed up
- Updated README.md for new Gatsby site and included instructions on how to submit your project to TMN
- Deleted old FAQs so they stop being confusingly duplicated in this repo
- Fix one FAQ that I noticed was incorrect. Someone probably needs to go through them all and make sure they are still up to date and none of the formatting has broken.


ara4n:
Thanks for these! On the twitter cards, i'm anxious that if an explicit image is set, we keep it as large as possible for visibility.  However, if no image is set, totally agreed that the written content should take priority (and the arbitrary pages need to do better at presenting that content - e.g. matrix.org/foundation currently just says "Blog, blog").

aaronraimist:
Yeah I agree but I don't know how to do this. Feel free to revert that commit if you know how to implement that.

benparsons:
@aaronraimist thank you very much for this PR - I made some small fixes on top, also incorporated Matthew's suggestion to use different card modes.

Separately, we need to restore the "featured image" for a lot of old content

ara4n:
hm, should this have fixed this?

Screenshot 2019-06-12 at 17 13 38

aaronraimist:
Should have. https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started.html says it is cached for 7 days.

benparsons:
I just forced it using https://cards-dev.twitter.com/validator - the card type should be be fixed now. The content is still not updated

PR #534, opened 2019-07-30
Push r0.1.1, Appservice r0.1.2, Server-server r0.1.3, updated index.html

turt2live:
Spec clarification releases. See:
* https://github.com/matrix-org/matrix-doc/pull/2201
* https://github.com/matrix-org/matrix-doc/pull/2202
* https://github.com/matrix-org/matrix-doc/pull/2203

Note: A client-server release is not included here as a more major release is expected in the very near future (making it somewhat pointless to cut now).


turt2live:
> I didn't read through all 10,600 changed lines

I also didn't do this, that sounds like a bad day. They are generated and dumped in here though - checkmarks on the matrix.org side are mostly procedural rather than practical.

PR #545, opened 2019-08-15
add projects from Jojii

benparsons:


benparsons:
@JojiiOfficial, this is a "Pull Request" - basically you can look at the "Files changed" tab to get an idea of the new code (which in this case is just markdown), and comment/review it

Issue #549, opened 2019-08-19
matrix-bifrost has the wrong room link


Half-Shot:
It should be #bifrost:half-shot.uk

PR #554, opened 2019-08-19
Edited the file to the new room as requested.

joakimvonanka:
#549


benparsons:
@Half-Shot can you confirm this is the correct room for Anton?

PR #599, opened 2019-11-05
Update spec index.html for Room v5 as the default

turt2live:
Also some other minor change which never made it to matrix.org apparently.

**Please review with https://github.com/matrix-org/matrix-doc/pull/2347**


turt2live:
(I'm going to merge this without CI because the CI takes 20 minutes to do stuff)

PR #604, opened 2019-11-14
Update footer.js

shivam1192:
added Github link to Github logo


benparsons:
Hi, we're quite happy with the direct from `/code` - makes it easier to change the redirect in future if we ever choose to. Also, the link is intended to go to the matrix-org organisation, not just the website repo.

PR #697, opened 2020-04-14
Add Maelstrom

dr-bonez:


aaronraimist:
That room appears to be invite only

aaronraimist:
The room mentioned in the README is `#maelstrom-server:matrix.org`

PR #703, opened 2020-04-22
Ben/bridging faq

benparsons:
This text is based on comments made by @Half-Shot in the public bridges room [earlier in the week](https://matrix.to/#/!ViClwbclMrgQwyPuvH:matrix.org/$K1Q_6A8VT0Rw4pvNeYz5WujpjPw9D4M2Xj7TVKedbFI?via=half-shot.uk&via=matrix.org&via=matrix.allmende.io). Seems to be a frequent topic.

### DIFF
--- a/gatsby/content/faq/bridging.mdx
+++ b/gatsby/content/faq/bridging.mdx
@@ -9,8 +9,8 @@ For example:
 
 > Is there any merit in a proposal that the current matrix.org Freenode bridge could be taken off matrix.org and put on a new HS (at a new server-name domain such as freenode.matrix.org) so as to not be bogged down in performance?
 
-In this case you are now maintaining two high traffic homeservers rather than one: Freenode handles enough traffic that it's intensive just by itself.
+The Matrix.org Foundation currently runs one homeserver: matrix.org. This homeserver is operationally intensive and can sometimes suffer from slow response times. However running two homeservers would not improve performance, because the work done to maintain one homeserver would need to be duplicated. Also, the traffic on the bridge instance would be intensive just by itself. Running two intensive homeservers, where optimisations such as caching and federation traffic batching would need to be duplicated would have a marked decrease in overall performance.
 
 Running the bridge on another homeserver means the same traffic will be hitting matrix.org, except it now has to be handled by the federation inbound workers too. Now you have two hosts handling the same amount of traffic. You might see a benefit to federated homeservers, but the new homeserver will struggle just as much with inbound/outbound federation slowness as it does now with bridging.
 
-Further, it may turn out that the ongoing work on Synapse to introduce shared master processes actually makes it so much more efficent that whole benefit of creating two homeservers is now moot.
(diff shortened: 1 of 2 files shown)


gatsby/content/faq/bridging.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+For example:
+
+> Is there any merit in a proposal that the current matrix.org Freenode bridge could be taken off matrix.org and put on a new HS (at a new server-name domain such as freenode.matrix.org) so as to not be bogged down in performance?
+
+In this case you are now maintaining two high traffic homeservers rather than one: Freenode handles enough traffic that it's intensive just by itself.
+
+Running the bridge on another homeserver means the same traffic will be hitting matrix.org, except it now has to be handled by the federation inbound workers too. Now you have two hosts handling the same amount of traffic. You might see a benefit to federated homeservers, but the new homeserver will struggle just as much with inbound/outbound federation slowness as it does now with bridging.
+
+Further, it may turn out that the ongoing work on Synapse to introduce shared master processes actually makes it so much more efficent that whole benefit of creating two homeservers is now moot.
```
Half-Shot, line 16: I think we should link to an article or something about sharding, or even Erik's amazing video so people don't come asking about sharding :)

gatsby/content/faq/bridging.mdx
```diff
@@ -0,0 +1,16 @@
...
+faq_section: bridging
+---
+
+#### Why don't you use separate homeserver instances to improve bridging performance?
+
+For example:
+
+> Is there any merit in a proposal that the current matrix.org Freenode bridge could be taken off matrix.org and put on a new HS (at a new server-name domain such as freenode.matrix.org) so as to not be bogged down in performance?
+
+In this case you are now maintaining two high traffic homeservers rather than one: Freenode handles enough traffic that it's intensive just by itself.
```
Half-Shot, line 12: Something about this statement feels wrong in tone. I know I wrote it, but maybe it needs a more positive tone somehow...

Something along the lines of "The matrix.org foundation currently runs one homeserver, matrix.org. This homeserver is operationally intensive and can be slow. However running two homeservers would not improve performance, because the work done to maintain one homeserver would need to be duplicated. Also, the traffic on the bridge instance would be intensive just by itself. Running two intensive homeservers, where optimisations such as caching and federation traffic batching would need to be duplicated would have an marked decrease in overall performance, to say nothing of the cost of running them."

Half-Shot: Maybe not verbatim, but maybe you can wordsmith it?

gatsby/content/faq/bridging.mdx
benparsons: Good idea, added. We will have more to show on this after it's released.

gatsby/content/faq/bridging.mdx
benparsons: Very good - added and edited.

PR #712, opened 2020-04-26
Replace link to nginx-rtmp-module with a maintained fork

anoadragon453:
So I had to set up livestreaming this weekend for an event and decide to use our blog post!

It worked out in the end, but the suggested repo for nginx-rtmp-module just wouldn't compile with the latest nginx. Plus seeing the last commit being in 2017 made me think hard about turning around. I did try it, but it wouldn't even compile with the latest stable release of nginx.

I looked up some guides for the process and found everyone recommending a fork of this module at https://github.com/sergey-dryabzhinsky/nginx-rtmp-module. This one seems to be actively maintained, and sure enough compiled fine.

I think we should recommend people use this up-to-date version instead of the current one.


benparsons:
good idea

PR #724, opened 2020-05-07
Update license identifier in all projects to use SPDX

tulir:
Intended to fix this mess:

image

Also a few other fixes, like marking all of kamax's projects, [synapse_scripts](https://github.com/xwiki-labs/synapse_scripts) and matrixstats.org as not maintained and updating neo's links

Maelstrom has Apache/MIT dual license, so I made it use the `(Apache-2.0 OR MIT)` like SPDX says dual licenses should be. The license filtering probably doesn't support it, but could be updated to follow the SPDX spec fully.

### DIFF
--- a/gatsby/content/projects/2014-06-09-riot.mdx
+++ b/gatsby/content/projects/2014-06-09-riot.mdx
@@ -9,7 +9,7 @@ description: Riot is a glossy web client with an emphasis on performance and usa
 author: Riot.im
 maturity: Stable
 language: JavaScript
-license: Apache
+license: Apache-2.0
 repo: https://github.com/vector-im/riot-web/
 home: https://riot.im/
 screenshot: /docs/projects/images/riot-web-large.png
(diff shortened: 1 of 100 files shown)


benparsons:
Great job - thank you @tulir

jaller94:
My project [silvy-matrix](https://github.com/matrix-org/matrix.org/blob/master/gatsby/content/projects/2018-04-24-silvy-matrix.mdx) got identified as `AGPL-3.0-only`.
The license I specified (only in the [package.json](https://gitlab.com/jaller94/silvy-matrix/-/blob/master/package.json)) is `AGPL-3.0-or-later`.

Mainly mentioning this to point out that there could have been other misidentifications like this.

tulir:
That's possible, I think I only checked license headers in source files to determine if it's `-or-later` (and defaulted to `-only` when there were no license headers)

jaller94:
Also, if you want the consistency to last, we would need some CI tests to enforce SPDX names.

[ligase](https://github.com/matrix-org/matrix.org/blob/master/gatsby/content/projects/2020-05-28-ligase.mdx) got added after your PR and does not follow the SPDX convention for a AGPL license.

PR #727, opened 2020-05-08
Reflect recent work in Nhekos feature list

deepbluev7:
Not sure, if I should specify, that only some aspects of pushrules are supported (you can mute and unmute rooms, which is like 90% what you use them for, but you can't add custom highlight keywords yet, etc.) Also E2EE works, but no device verification yet.

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>


deepbluev7:
Thank you :3

PR #728, opened 2020-05-11
Add HLS livestream to open-tech-meetup (disabled)

dbkr:
With switch that can be toggled to enable it before the event
(will still need a site redeploy though).


gatsby/src/pages/open-tech-meetup.js
```diff
@@ -9,17 +9,28 @@ import config from '../../config'
...
 const title = `Open Tech Will Save Us | ${config.siteTitle}`;
 
+const SHOW_LIVE_STREAM = false;
+
+let liveStream;
+if (SHOW_LIVE_STREAM) {
+    liveStream = <div>
+        <video id="videoStream" style={{width: '640px', height: '360px'}} src="https://stream.matrix.org/hls/live.m3u8" controls></video>
+        <script src="/js/hls.light.min.js"></script>
+        <script src="/js/livestream.js"></script>
```
benparsons, line 19: does this need to be in the header? or happy in the body?

dbkr: They actually probably need to be here so they load after the video tag has loaded, or if in the head I think they'd have to listen for the `load` event and run after that.

PR #738, opened 2020-05-21
some updated stats

benparsons:


ptman:
time for another update, maybe?

MadLittleMods:
Going to merge this anyway as these stats are still more up to date than before.

Another update would be good though.

PR #740, opened 2020-05-24
Add guide to fancy e2ee stuffs

Sorunome:
This is a WIP guide for implementing cross-signing and all that fuzz. The sections on SAS and displaying verification status based on signatures are already complete.

Please say if any additional metafiles need to be changed!

[Rendered](https://github.com/Sorunome/matrix.org/blob/soru/guide/gatsby/content/docs/e2ee_cross_signing.mdx)


gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+
+TODO: This section doesn't list yet how to *create* keys, add this later on.
+
+### General Structure of SSSS storage
+SSSS depends on keys with which the actual secrets are being encrypted to be stored. These keys need
+to be provided by the client. There is, however, metadata information on the keys in a users account_data.
+For that, there are entries `m.secret_storage.key.<keyId>`, where the keyId is the unique ID of said
+key. In addition, there is an `m.secret_storage.default_key` which denotes the default key to use.
+
+The `m.secret_storage.<keyId>` objects contain a mac and an iv to easily verify if you have the correct
```
poljar, line 583:
```suggestion
The `m.secret_storage.<keyId>` objects contain a MAC and an IV to easily verify if you have the correct
```

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+key, and, in case if there is also a password associated with it, some information on how to generate
+the key out of the password. More specific on that later.
+
+Secrets, such as the private keys for cross-signing, can only be strings (not binary data) and are
+stored encrypted in the account data of your account. For example, the `m.cross_signing.master` entry
+in account data is the private key of your master cross-signing key. Each secret can be encrypted by
+multiple SSS keys. Inside the `encrypted` object there is a mapping of key ID to the ciphertext, iv
+and mac of the secret. A content could look as following:
```
poljar, line 591:
```suggestion
and MAC of the secret. A content could look as following:
```

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+[as seen in the client-server spec](https://matrix.org/docs/spec/client_server/latest#verification-messages-specific-to-sas).
+
+#### `m.key.verification.start`
+This sends a bunch of parameters that the sending device supports. Please note that you will have to
+remember this object while performing a commitment check (explained in the next section), of the one
+that *includes* the metadata of `transaction_id` and `from_device` keys, omitted in these examples.
+```json
+{
+  "method": "m.sas.v1",
+  "key_agreement_protocols": ["curve25519-hkdf-sha256", "curve25519"], // while "curve25519" is depricated, it is still wildly used and thus needs to be supported
```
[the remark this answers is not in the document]
Sorunome: Oh, that's great, at the point of implementing that was still needed. Can easily remove that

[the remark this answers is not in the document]
Sorunome: Might be best to wait a month or two before that, though, to give people time to update their riots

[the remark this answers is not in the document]
Sorunome: Could also add a note "This deprecated method will be removed from the guide starting September 2020, as it should really not be used. However, some old clients might still depend on it"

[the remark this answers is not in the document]
poljar: I mean by the time we publish this guide and someone uses it to implement stuff I would expect most people to have updated. We should also add the time it takes to implement this and do a release.

I would say a month or two would have already passed.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+   verified master key. This in practice, however, does not allow easy migration between device-device verification and
+   cross-signing verification. Recursing until you hit a verified device is thus preferred.
+
+## 3. Implementing SSSS
+SSSS is short for Secure Secret Storage and Sharing, which provides a method of storing secret data
+on the server, without the server administrator or anyone else having access to that data. As such,
+this is optimal for storing the private keys for cross-signing on the server so that all your devices
+can use them later on to generate signatures, and thus verify other people.
+
+Unfortunately there doesn't seem to be any help from libolm for implementing SSSS (yet?) so this will
```
uhoreg, line 562: > (yet?)

Probably not for the foreseeable future, unfortunately.  Adding support for SSSS would mostly end up being just olm providing some cryptographic primitives, and we don't really want olm to be a library of cryptographic primitives (especially as we're looking into changing olm so that you can use better implementations such as HW acceleration).

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+      "mac": "macthingy"
+    }
+  }
+}
+```
+
+### Caching of secrets
+You have to access some secrets comparatively often. As fetching from the store requires to get a key
+for said secret, that requires user interaction. As such, if a client can store a secret seasonably
+securely client-sided, it is recommended to cache some secrets locally. For cross-signing that would
```
uhoreg, line 608: I would say something along the lines that each client author should make a decision about which secrets to cache based on tradeoffs between the sensitivity of the information, its ability to keep things secret, and how much of an inconvenience it presents to the user.  As some clients may want to cache more or less than what's recommended.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+### Sharing
+The same limitations of caching apply to sharing: It only makes sense to request secrets or share
+secrets that are cacheable, as else potentially sensitive data is leaked, or it requires user interaction
+to load those secrets in the first place anyways.
+
+Secrets can only be requested from other devices of your account that you already have verified, be
+it via cross-signing or directly. As such, it might be a good idea to request all cacheable secrets
+after successfully completing verification with yourself, as far as that also verifies your own master
+key (else the other client likely doesn't have cross signing and thus requesting secrets doesn't make
+much sense).
```
uhoreg, line 630: If requesting secrets after completing verification, it may be worth introducing a small delay before requesting secrets, to make sure that the other client has had a chance to mark you as being trusted.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+
+You request a secret by sending an `m.secret.request`, and you receive it back via an `m.secret.send`
+to_device event. Please keep in mind that these all should be encrypted! To do your part in the sharing
```
uhoreg, line 633: > Please keep in mind that these all should be encrypted!

The `send` must be encrypted, but the `request` does not have to be.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+  "request_id": "<same request id as received in m.secret.request>",
+  "secret": "<contents of the secret>"
+}
+```
+
+#### Asking for a secret
+1. Only ask your own, verified devices for a secret. Be reasonably sure that they might have it (e.g.
+   the other device sent you your master key to verify).
+2. When receiving a secret, verify that it was encrypted
+3. When receiving a secret, verify that it was from one of the devices you sent the secret to (possible
```
uhoreg, line 661:
```suggestion
3. When receiving a secret, verify that it was from one of the devices you sent the request to (possible
```

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+account data events of the secrets, the mac, the IV and the ciphertext are all base64 encoded.
+
+Note: Encrypting and decrypting both uses *unpadded* AES-CTR. If your library does not have an option
+for unpadded AES-CTR en/decryption, simply append prior en/decryption 0's to your data to fill up the
+input to a full number of blocks. One block here is typically 32? bytes in size. After en/decrypting,
+strip the extra bytes again. Please note that this method will *not* work with all ciphers (e.g. XTS), it is
+something specific to CTR.
+
+#### Deriving the keys using HKDF
+As key deriving is needed for both encrypting and decrypting it makes sense to implement this first.
```
uhoreg, line 804: Maybe worth noting that implementing this is only needed if you don't have access to an HKDF implementation.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+  if (info.algorithm != 'm.pbkdf2') {
+    throw 'Unknown algorithm';
+  }
+  var generator = PBKDF2(hashAlgorithm: sha512);
+  return Uint8List.fromList(generator.generateKey(password, info.salt,
+      info.iterations, info.bits != null ? info.bits / 8 : 32));
+}
+```
+
+Important: Some PBKDF2 algorithms take as length the length in bytes (like here), while others take
```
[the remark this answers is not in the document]
Sorunome: ah, yes, thanks

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+
+### Encrypting and decrypting
+Both encrypting and decrypting is done with AES-CTR. Out of the secret key and the name of the secret
+using [HKDF](https://en.wikipedia.org/wiki/HKDF) with SHA-256 both the actual AES key and the key for
+the MAC integrity is generated. IVs are randomly generated 16 bytes, with bit 63 set to 0. In the
+account data events of the secrets, the mac, the IV and the ciphertext are all base64 encoded.
+
+Note: Encrypting and decrypting both uses *unpadded* AES-CTR. If your library does not have an option
+for unpadded AES-CTR en/decryption, simply append prior en/decryption 0's to your data to fill up the
+input to a full number of blocks. One block here is typically 32? bytes in size. After en/decrypting,
```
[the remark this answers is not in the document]
Sorunome: Ah yeah, thanks. Thus the questionmark ^-^

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+strip the extra bytes again. Please note that this method will *not* work with all ciphers (e.g. XTS), it is
+something specific to CTR.
```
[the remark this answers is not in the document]
Sorunome: It is specific to CTR mode in that it is a stream cipher and not a block cipher?

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+  "commitment": "the-commitment-calculated"
+}
+```
+
+#### `m.key.verification.key`
+This sends the ephemeral *public* key to each other. If you haven't generated it yet (due to not having
+calculated the commitment), do so now with `var sas = olm.SAS();` and read it with `sas.get_pubkey();`.
+Be sure that if you received the commitment, you use the received public key to verify the commitment
+you received earlier. The public key is both added to the `sas` object with `sas.set_their_key(payload["key"]);`
+and also saved in a separate variable, as you'll need it later, as unfortunately the `sas` object does
```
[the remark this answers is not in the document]
Sorunome: That is the generated public key, though, not the one you set via `sas.set_their_key`

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+var sasInfo = 'MATRIX_KEY_VERIFICATION_SAS' +
+    (request.startedVerification
+        ? ourInfo + theirInfo
+        : theirInfo + ourInfo) +
+    request.transactionId;
+```
+
+##### Generate bytes
+Once you have that sas info, you use it to generate bytes through `sas.generate_bytes(sasInfo, bytes)`.
+For emoji verification, 6 bytes are needed, for number verification 5 are needed.
```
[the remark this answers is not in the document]
Sorunome: Didn't know you could do that! TIL!

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,920 @@
...
+(logins) only once. For this, each *user* has three keys: A master key, a self-signing key and a
+user-signing key. The master key will sign both the self-signing and the user-signing key. The
+self-signing key will then sign your own device keys, and the user-signing key will sign other people's
+master keys. The purpose of splitting this up into three keys is that in the case of either the
+self-signing key or the user-signing key getting compromised, you can easily exchange them, while
+still keeping the master key. As the master key is *only* used for signing your *own* user-signing and
+self-signing key, it is rarely used, so the attack surface to obtain it is minimal. Optionally, device
+keys themselves can sign their own master key.
+
+![](https://raw.githubusercontent.com/uhoreg/matrix-doc/cross-signing2/proposals/images/1756-graph2.dot.png)
```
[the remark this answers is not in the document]
Sorunome: Haahaa, should soru use the one of JCs guide instead? That didn't exist yet when writing that part ^^

gatsby/content/docs/e2ee_cross_signing.mdx
Sorunome: Good point! *adds that to own implementation*

gatsby/content/docs/e2ee_cross_signing.mdx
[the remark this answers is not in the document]
Sorunome: How is this re-wording?

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -110,6 +117,9 @@ set to the ID of the sending device. So the above payload becomes:
 
 For simplicity, all payloads mentioned below will be shown without those extra keys.
 
+While the to_device messages are sent encrypted, the room messages don't *have* to be encrypted - it
```
poljar, line 120: The to_device messages are not sent encrypted. They are always sent in the clear.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -604,11 +627,14 @@ and mac of the secret. A content could look as following:
...
-for said secret, that requires user interaction. As such, if a client can store a secret seasonably
-securely client-sided, it is recommended to cache some secrets locally. For cross-signing that would
-be `m.cross_signing.self_signing`, `m.cross_signing.user_signing` and for online key backup that would
-be `m.megolm_backup.v1`. Blindly caching all secrets is a security risk, as some secrets such as
-`m.cross_signing.master` are highly sensitive and shouldn't be stored in plaintext on any disk, ever.
+for said secret, that requires user interaction. As such, if a client can store a secret reasonably
+securely client-sided, the client could cache some secrets locally. As this introduces an attack surface
+of a third party manually reading out the clients cache, but also improves user experience to not have
+to ask the user for their secret store passphrase, each client will have to decide which secrets to
+cache and which not, based on what kind of security it can offer. For cross-signing that needed secrets
```
poljar, line 634:
```suggestion
cache and which not, based on what kind of security it can offer. For cross-signing the needed secrets
```

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -734,21 +761,21 @@ generate the key off of the password. It an look as follows:
...
   }
   var generator = PBKDF2(hashAlgorithm: sha512);
-  return Uint8List.fromList(generator.generateKey(password, info.salt,
+  return Uint8List.fromList(generator.generateKey(passphrase, info.salt,
       info.iterations, info.bits != null ? info.bits / 8 : 32));
 }
 ```
 
-Important: Some PBKDF2 algorithms take as length the length in bytes (like here), while others take
+Important: Some PBKDF2 implementations take as length the length in bytes (like here), while others take
```
poljar, line 778:
```suggestion
Important: Some PBKDF2 implementations take the length in bytes (like here), while others take
```

gatsby/content/docs/e2ee_cross_signing.mdx
Sorunome: Huh, riot at least works if you send them encrypted. So maybe change that they *can* be encrypted, then?

poljar: Clients will usually parse the decrypted event the same way it would parse an unencrypted one (except that they will remember that the event was encrypted).

So they will mostly accept encrypted versions as well. Riot doesn't send them encrypted and I don't think any of the mobile Riots do so either. The spec doesn't mention anywhere anything about them being encrypted either, though it doesn't mention that they can't be either.

There is not much to gain by sending those events encrypted, the server won't know that you're currently verifying the device that should receive the message.

That being said there is not much downside either, it's a bit more complex since you need to setup Olm a session before you can start verifying a device.

Sorunome: > That being said there is not much downside either, it's a bit more complex since you need to setup Olm a session before you can start verifying a device.

Ideally you'll have a method that sends a to_device method that does all that stuff for you anyways, so for the client developer it'd be zero-effort to add encryption. That being said, soru will update that section accordingly.

poljar: By more complex I didn't mean that a developer needs to do this, I mean that a client needs to go through a more complex path before it can do the verification dance which might introduce more delays that end up being noticeable to the user.

It's not that important since the same problem can happen with in-room verifications.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,949 @@
...
+  } else {
+    // if we don't have one, generate a random 16-byte IV
+    iv = Uint8List.fromList(SecureRandom(16).bytes);
+  }
+  // And don't forget to set bit 63 of the IV to zero
+  iv[8] &= 0x7f;
+
+  // derive our AES and HMAC keys using the before written function
+  var keys = deriveKeys(key, name);
+ciphertext
```
deepbluev7, line 909:
```suggestion

```

gatsby/content/docs/e2ee_cross_signing.mdx
Sorunome: Alright, adjusted the paragraph!

gatsby/content/docs/e2ee_cross_signing.mdx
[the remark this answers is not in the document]
uhoreg: Yeah, the one from JC's guide is fine (credit as needed)

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,1219 @@
...
+var theirInfo = '${request.userId}|${request.deviceId}|${theirPublicKey}|';
+var sasInfo = 'MATRIX_KEY_VERIFICATION_SAS|' +
+    (request.startedVerification
+        ? ourInfo + theirInfo
+        : theirInfo + ourInfo) +
+    request.transactionId;
+```
+
+##### `curve25519`
+IMPORTANT! This section will be removed before the guide is published. If it is after September 2020,
```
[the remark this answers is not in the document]
Sorunome: done, added a short paragraph that it is deprecated and shouldn't be implemented, in case people get confused.

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,1198 @@
+---
+layout: post
+title: Implementing more advanced e2ee features, such as cross-signing
+section: Encryption
+---
+
+PLEASE NOTE THAT THIS GUIDE IS STILL A WIP!
+
+# A guide to more advanced e2ee features of matrix
```
benparsons, line 9: The title from the front matter above gets inserted as an h1, used as the title and the slug. This is for consistency. You don't need to have a second title. Pick one to use as the title, and if you prefer you could keep the second one just as a subhead (can be regular text)

gatsby/content/docs/e2ee_cross_signing.mdx
```diff
@@ -0,0 +1,1198 @@
...
+(logins) only once. For this, each *user* has three keys: A master key, a self-signing key and a
+user-signing key. The master key will sign both the self-signing and the user-signing key. The
+self-signing key will then sign your own device keys, and the user-signing key will sign other people's
+master keys. The purpose of splitting this up into three keys is that in the case of either the
+self-signing key or the user-signing key getting compromised, you can easily exchange them, while
+still keeping the master key. As the master key is *only* used for signing your *own* user-signing and
+self-signing key, it is rarely used, so the attack surface to obtain it is minimal. Optionally, device
+keys themselves can sign their own master key.
+
+![](./matrix-cross-signing-key-diagram.svg)
```
benparsons, line 472: The image itself needs to be moved to `./gatsby/static/docs/guides/img`, and then this link can be updated to `/docs/guides/img/matrix-cross-signing-key-diagram.svg`.

Issue #747, opened 2020-05-26
Refactor UI: Clean up discover page


ludwigbald:
There are a lot of buttons, a lot of them are duplicates of each other.

ludwigbald:
On the discover/Try Matrix page, project thumbnails should be uniform height.

thibaultamartin:
Made obsolete by the new website

PR #756, opened 2020-05-28
Clean up and Partly Merge: Try Matrix / Discover

ludwigbald:
- I merged the confusing Try Matrix (project browser) page with the Discover page The project selector now lives on the Discover page.
- I cleaned up the Discover page, removing duplicate links (#747)


benparsons:
@ludwigbald thank you for this PR - can you break it out into separate PRs so we can discuss what's being proposed here?

ludwigbald:
> @ludwigbald thank you for this PR - can you break it out into separate PRs so we can discuss what's being proposed here?

I'll see what I can do. I agree that would make things easier, but it seems complicated :D

MadLittleMods:
Closing this as it has gone a bit stale and the review last review points are still outstanding.

If the changes are still relevant, it seems like it would be easy to copy the last few into a new branch. As Ben mentioned, can also split them out individually.

PR #760, opened 2020-05-29
Remove Stale Link to shop.matrix.org from the navbar

ludwigbald:
Maybe we should not have a dead link in such a prominent place


benparsons:
The shop isn't dead, we'll have it back running in a couple of weeks, with an annoucement that it's running.

ludwigbald:
I mean, then we can just put the button back again, and put it in a more meaningful spot.

PR #773, opened 2020-06-26
Add stub implementation guides

erikjohnston:
The idea is to create some stub implementation guides for client/servers, so that we have an easy place for people to dump small snippets of advice without the overhead of writing a full blown guide/blog post.

This uses [mdbook](https://rust-lang.github.io/mdBook/).

See https://jki.re/book/ for a demo of this PR.

Things to figure out:

- [ ] What do we want it to look like? It currently just uses the default themes, but we probably want to add the matrix.org header at the least
- [ ] How do we hook this up to the CI?


benparsons:
I merged this, but there are questions around:

* where in the website should this live? I wonder about using a subdomain - ideally we'd have something like spec.matrix.org for the spec itself and then this could sit at spec.matrix.org/implementation or similar.
* CI

erikjohnston:
So Matthew has said he doesn't care that much and we should just pick a url ftr

PR #786, opened 2020-07-17
Add building the implementation guides to CI

erikjohnston:
I think this works?


benparsons:
This will result in the output being on the same domain?

erikjohnston:
There's a PR up to provision a spec.matrix.org which will have `/book` and `/guides` point at this, but yeah, it'll also be around on matrix.org. This just seemed like the easiest way to get going

erikjohnston:
Woo, looks like it finally is doing the right thing!

erikjohnston:
It worked! https://matrix.org/implementation-guides/server/ :tada:

PR #787, opened 2020-07-20
Update Nheko client info

strct:
See description of [Nheko](https://github.com/Nheko-Reborn/nheko) since https://github.com/Nheko-Reborn/nheko/commit/6b16616e3a646a5fe3046434a54f9e673c1c0db6


deepbluev7:
Thanks for picking up my slack :D

Could you maybe also add, that nheko supports editing tags and reactions now? Then I don't have to open a separate PR for it. :3

strct:
Done

deepbluev7:
Thank you!

PR #788, opened 2020-07-20
Wording: In the next in this series -> Next in this series,

jaller94:
It feels like there's a word missing.. like next *post* or next *article*.

But I like the short version much better, so I'm suggesting that.


benparsons:
> In the next in this series

Is rather colloquial, but I'm not sure it's wrong. Anyway I like your change. 👍

PR #797, opened 2020-07-31
Add links to recordings for Open Tech Will Save Us.

clokep:
I had missed the fourth one and was trying to find the recording. Had to do some URL hacking from the 1st/2nd event to find it. 😄


clokep:
Looks like f552308825c4d727047c0aadbd0810251e8cde1a did the same thing as this.

PR #806, opened 2020-08-06
Add language to bots overview

Pamplemousse:
As a beginner looking for inspiration to write a bot in language X, this
allows me to filter bots of interests without changing page (clicking on
each bot individually).

Signed-off-by: Pamplemousse <xav.maso@gmail.com>

### DIFF
--- a/gatsby/src/components/MXProjectCard.js
+++ b/gatsby/src/components/MXProjectCard.js
@@ -3,6 +3,10 @@ const MXProjectCard = ({ project, imageSize }) => {
     imageSize = imageSize ? imageSize : 300;
     return <div className="mxclientcard">
         <h3><a href={project.slug}>{project.title}</a></h3>
+        {project.categories && project.categories[0] === "bot" && <>
+        <small>{project.language}</small>
+        <br />
+        </>}
         {project.thumbnail && <a href={project.slug}>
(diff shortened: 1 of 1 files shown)


gatsby/src/components/MXProjectCard.js
```diff
@@ -3,6 +3,7 @@ const MXProjectCard = ({ project, imageSize }) => {
     imageSize = imageSize ? imageSize : 300;
     return <div className="mxclientcard">
         <h3><a href={project.slug}>{project.title}</a></h3>
+        <h6>{project.language}</h6>
```
benparsons, line 6: some points:

* `h6` looks good but feels just wrong here, using `small` would be ok though.
* MXProjectCard is designed to be consistent across pages, and I think adding the language makes sense for bots, it's not right for clients. So, we need to check what type of project we're rendering before rendering the line.

I came up with the following to replace what you have, what do you think?

```
{project.categories && project.categories[0] === "bot" &&
  <>
    <small>{project.language}</small>
    <br />
   </>}
```

Pamplemousse: Oh, I did not realize the component was used elsewhere...
I corrected with the snippet you gave, and it seems alright!
Thanks :)

benparsons (approved):
lgtm

PR #836, opened 2020-09-21
Add Cloudron as installation method

gramakri:
We (Cloudron team) have been maintaining packages for Synapse and Element for quite a while now. We also have many users of Synapse on our platform (https://forum.cloudron.io/category/50/matrix-synapse-element). Would be great to have Cloudron listed as an installation option alongside others.


ara4n:
lgtm, although it'd be better if the Element link was on the Element page. I'll merge it for now though.

PR #843, opened 2020-09-30
added german matrix introduction for users

samuel-p:
I started an introduction in matrix on my German blog. Maybe this is interesting to add to the list of guides for users on matrix.org.


benparsons (approved):
lgtm

Issue #333, opened 2019-02-07
Search by programming language with wrong results


mttkrb:
When I'm searching on https://matrix.org/docs/projects/try-matrix-now.html for clients with C# as language the following results are shown:
- Nebuchadnezzar
- matrix-client.el 
- Emacs Maelstrom 
- FluffyChat 
- AgentSmith

In all these projects the language is not C#

mrdemc:
Nebuchadnezzar is a C# client: https://github.com/MartijnBraam/Nebuchadnezzar
If I try the same I only get this project, the others are filtered out.

mttkrb:
This is my search result. 
![grafik](https://user-images.githubusercontent.com/21170293/52478075-7f301800-2ba4-11e9-95fa-e4e366e806e9.png)

Ok Nebuchadnezzar is really a C# client. I was a little bit confused, because its description on the matrix page is "Linux native chat client (GTK2)"

benparsons:
The result you're seeing is still wrong. Am taking a look.

benparsons:
@mttkrb can you tell me if you're still seeing this result? Made various unrelated fixes and can no longer repro

mttkrb:
Hi,
at a first look there no wrong results. 
thanks a lot.

PR #852, opened 2020-10-07
Tweak Jekyll styles for readability

wbamberg:
Over in https://github.com/matrix-org/matrix-doc/issues/2805 I suggested making some changes to the spec CSS to make it more readable. But some of the relevant CSS is coming from the Jekyll builder over here. Specifically:

* `font-weight: 300` makes the text harder to read than a normal weight
* `color: #666` is worse for contrast than `#333`, and `#333` is the [style guide recommendation](https://matrix-org.webflow.io/style-guide)

Before these changes, the spec looks like:

before1

After these changes, the spec should look like:

after1

Unfortunately I'm not sure how to build the Jekyll site locally to test this properly, but I'm happy to do so if you can give me some pointers.

### DIFF
--- a/jekyll/_sass/_base.scss
+++ b/jekyll/_sass/_base.scss
@@ -17,7 +17,6 @@ body {
     font-family: $base-font-family;
     font-size: $base-font-size;
     line-height: $base-line-height;
-    font-weight: 300;
     color: $text-color;
     background-color: $background-color;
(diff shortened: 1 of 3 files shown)


ara4n:
this looks very sane to me. i vote we merge it and see what breaks :D

ara4n (approved):

PR #864, opened 2020-10-24
Replace Riot with Element

### LINKED ISSUE #832: https://matrix.org/docs/guides/moderation mentions Riot
we should probably s/riot/element across the repo

moritzdietz:
The first commit 222c8f86b61e342bd47790445a5f9fe4e056a989 closes #832 
A suggested in the issue I am planning on adding another commit which replaces Riot throughout the entire Repository.
But now that I am thinking about this, it might not be a good idea to do it globally.
Maybe exclude everything in `gatsby/content/blog/` Any thoughts?

I might do multiple commits targeting different areas of the repository so it is easier to review and gives more choice what can be merged.

I am thinking of the following:

- Ignore all linked URLs links from outside sources (other Matrix owned repositories)
- Ignore past blog blogposts for now (maybe we can add something like `Riot (now Element)` as it's been done now the other way around
- Rename files with `riot` in the name as well

---
### Findings

1. `https://matomo.riot.im` resolves but `https://matomo.element.io` doesn't so this needs to continue to exist


aaronraimist:
Probably not a good idea to do it globally since that would change all of the old blog posts. Doing in to the docs and other timeless stuff makes sense though.

gatsby/content/docs/2019-04-29-moderation.mdx
```diff
@@ -116,7 +116,7 @@ For further detailed info about server ACLs, see [https://matrix.org/docs/spec/c
...
+Prior to Matrix 1.0, a bug existed which caused room state (e.g. membership, power levels, bans/kick data) to intermittently be reset to a historical value due to a bad merge conflict resolution - a so-called ‘state reset’, or the ‘hotel california bug’.  This was improved in v2 and later of the room algorithm, which will be used as the default when creating new rooms in Matrix 1.0.  However, as of April 2019 most existing rooms still use the original buggy v1 algorithm.  The best way for now to handle a state reset is to manually re-apply any bans & ACLs if they are reset.  However, if state resets keep happening, it is possible to upgrade the room algorithm via Element/Web with the undocumented `/upgraderoom 3` command - but please note this is **not** recommended, as we are still testing the final room algorithm and upgrade mechanism for Matrix 1.0.
```
aaronraimist, line 119:
```suggestion
Prior to Matrix 1.0, a bug existed which caused room state (e.g. membership, power levels, bans/kick data) to intermittently be reset to a historical value due to a bad merge conflict resolution - a so-called ‘state reset’, or the ‘hotel california bug’.  This was improved in v2 and later of the room algorithm, which will be used as the default when creating new rooms in Matrix 1.0.  However, as of April 2019 most existing rooms still use the original buggy v1 algorithm.  The best way for now to handle a state reset is to manually re-apply any bans & ACLs if they are reset.  However, if state resets keep happening, it is possible to upgrade the room algorithm via Element/Web with the undocumented `/upgraderoom 6` command - but please note this is **not** recommended, as we are still testing the final room algorithm and upgrade mechanism for Matrix 1.0.
```

unrelated but while your editing this file might as well update this

moritzdietz:
> Probably not a good idea to do it globally since that would change all of the old blog posts. Doing in to the docs and other timeless stuff makes sense though.

This was suggested in the linked issue :) I am trying to come up with reasons to **not** do it.
But if the Matrix team thinks this is a good idea I am not the one to disagree ¯\_(ツ)_/¯ 

Also PR is still a draft and not ready for review.

gatsby/content/docs/2019-04-29-moderation.mdx
moritzdietz: IMO this change would be out of scope of what this PR is trying to achieve. Even though it's trivial.

benparsons (approved):
This lgtm, it just changes the moderation page?

moritzdietz:
Yup as of now I just changed the moderation page. So feel free to merge.
I was/am still working on checking where this can be applied as well on other pages - if this is still something you'd need.
As I know some sections shouldn't be a simple `s/riot/element/` but maybe something like `Element (previously Riot)`.
This could be something we could discuss in this PR.

moritzdietz:
Just fixed the conflict.

moritzdietz:
With https://github.com/matrix-org/matrix.org/pull/864/commits/bee3081af48c3531f5711e1e87625fc90a864fec I just have pushed some changes to the blog section. This is what I highlighted earlier. If folks agree that this would make sense and the wording is ok, I'll continue with the "refactoring".

I looked at element-web's repo where it says "Element (formerly known as Vector and Riot)" in the README and applied the same style here as well.
Looking for feedback now :)

benparsons (approved):
lgtm

benparsons:
I'm going to merge this and we can start a new PR going forward. Thanks @moritzdietz for your work!

PR #865, opened 2020-10-25
Fix markdown in /op example

hugopeixoto:
Small fix. Found this while browsing https://matrix.org/docs/guides/moderation#power-levels, the backticks were out of control.


benparsons (approved):
lgtm

PR #870, opened 2020-11-02
Update synapse federation docs link

Amar1729:
Changed previous outdated link (looks like they moved federation docs from their README to a separate file).


benparsons (approved):
lgtm

PR #877, opened 2020-11-11
Update mm repo and description

meutraa:
I moved the source code from gitlab to a gitea instance a while ago.

Updated the description while I was there.


benparsons (approved):
lgtm

Issue #873, opened 2020-11-04
https://matrix.org/bridges/ Signal bridge is missing


Thorgrimson:
on the clickable icons and in the list you cant click on signal.
maybe you could add https://github.com/tulir/mautrix-signal there? or remove signal from the list

tobymurray:
Agreed - tried this on Firefox and Chromium on Ubuntu, the other anchors seem to work e.g. https://matrix.org/bridges/#mumble but for Signal it flashes up and then disappears.

PR #886, opened 2020-11-22
Fix Signal not working on bridges page, update Slack logo

aaronraimist:
Fixes https://github.com/matrix-org/matrix.org/issues/873


benparsons (approved):
lgtm

PR #888, opened 2020-11-25
Update FluffyChat screenshot

ChristianPauly:
fluffychat-screenshot

Just a fancy new screenshot for FluffyChat for which I have paid 7 bucks, so something around one starbucks coffee, which is however not possible at the moment because of the lockdown :-(

Best regards and stay healthy


benparsons:
@ChristianPauly the context of this graphic helped, thanks!

PR #924, opened 2021-01-08
reduce number of clicks required by "Try Now"

Evidlo:
The `Try Now` button requires clicking through four pages before being presented with a login.  

    Try Now > Start Element Web > Try Element > Open in Browser

This simplifies the process somewhat:

    Try Now > Open in Browser


benparsons (approved):
I agree, lgtm

PR #938, opened 2021-01-28
Jekyll: Run `bundle update`

callahad:
Resolves a dependabot warning, produces semantically identical output.

Signed-off-by: Dan Callahan <danc@element.io>


callahad:
...do we even use anything in the Jekyll folder anymore?

callahad:
After a little more reading of the Buildkite pipeline and the Dockerfile we're using... I don't think we're actually using the Gemfile or Gemfile.lock at all. 🙃

MadLittleMods:
I don't have the definite source of truth but it feels like we only use the `gatsby` directory nowadays for the main site. Going to close this PR out.

PR #943, opened 2021-02-03
Remove Daydream

MTRNord:
I am officially facing the truth and calling this project dead due to missing time, enjoyment and I decided to not any further develop on daydream at all.


aaronraimist:
With the maturity set to Not actively maintained it is hidden by default but is still there if someone is looking for code to build off of or something. I think it is worth keeping.

MTRNord:
> With the maturity set to Not actively maintained it is hidden by default but is still there if someone is looking for code to build off of or something. I think it is worth keeping.

tbh I am thinking about removing the code too as I am really not happy about the project :/

benparsons (approved):
lgtm, though sadly

PR #952, opened 2021-02-12
Add Michaël to HoF.

dkasak:
This is related to https://github.com/matrix-org/internal-config/issues/947.


clokep:
I think this looks fine, but not sure I can approve! 👍

PR #956, opened 2021-02-25
Remove expired Github profile link.

dkasak:
The mentioned user deleted their profile and it was subsequently registered by another as a broken link hijacking POC. No risk, but may confuse someone in the unlikely event that they follow it.


benparsons (approved):
lgtm

PR #959, opened 2021-03-03
docs: mention up-to-date tutorial more aggressively

jdreichmann:
Currently, there is only a small note at the bottom of the article, which also is a dead link, that there is an up-to-date version of this tutorial.
This commits adds a more aggressive note at the beginning of the tutorial, as very few of the configs in the article stayed that way until today,
but the blog post itself shows up in search engines pretty high up


benparsons (approved):
lgtm

PR #964, opened 2021-03-16
Freeze the current appendices.html file for the legacy spec

anoadragon453:
This PR is part of unfreezing the spec as a whole after the spec redesign has launched: https://github.com/matrix-org/matrix-doc/issues/3055

Essentially we'd like to redirect the unstable versions of the old spec to the new spec site (https://spec.matrix.org). Unfortunately, we can't do that for the appendices page as there is no stable vs. unstable appendices page. It's never been versioned. As such, if we redirected it then any links to the appendices in the stable versions of the legacy spec would take you to the appendices page on the new spec - which would be very weird.

Thus, this PR freezes the appendices page in time. We'll be setting up redirects for all of the `/unstable` spec pages, as well as the `/proposals` page, as with the former the user consciously chooses to view the unstable spec, and in the latter the proposals page is a living document which isn't deep-linked into by the spec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@
 /projects-gen/
 /unstyled_docs/api
 /unstyled_docs/howtos/client-server.html
-/unstyled_docs/spec/appendices.html
 /unstyled_docs/spec/*/unstable.html
 
 /node_modules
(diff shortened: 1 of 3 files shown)


turt2live (approved):
I ctrl+f'd the appendices.html for `steal` and `hack`

anoadragon453:
One could also download the generated `styled-docs.tar.gz` from the build and check to see if appendices is in the there. But that works too.

PR #966, opened 2021-03-18
Fix Riot.im mention and link in Slack bridging page

jaywink:


MadLittleMods:
Thanks @jaywink! Looks sane 🐎

PR #968, opened 2021-03-23
Update matrix-appservice-bridge

jaller94:
* It's Stable (v2.5.0) and we use it in our stable Slack and IRC bridges.
* The language changed to TypeScript
* [ ] It could get moved into the `sdk` sub-folder. (does that mess with its URL?)


MadLittleMods:
This looks like a good iterative improvement. We can move the `sdk` subdirectory in another PR if we want.


Tested locally and looks good :+1:

http://localhost:8000/docs/projects/as/matrix-appservice-bridge

MadLittleMods:
Thanks for for the update @jaller94 🦜

PR #978, opened 2021-04-22
Fix broken links to MSC1777

anoadragon453:
The foundation page and a blog post was linking to `github.com/.../pulls/1777`, which attempts to find PRs authored by user `1777`. Looks like a typo of `pull/1777`.

PR #980, opened 2021-04-22
Return swagger-ui to 64dc306

anoadragon453:
#978 accidentally included a submodule update to the `swagger-ui` directory, which our usual CSS patch then failed to apply against - breaking CI.

This PR returns the submodule to the previous commit.


anoadragon453:
The spec CI step has passed, so looks like things are fixed.

PR #984, opened 2021-05-07
Remove need to install `gatsby` as a global dependency

MadLittleMods:
Remove need to install `gatsby` as a global dependency

Use the built-in npm scripts which already have `gatsby`
already on the path. Already in place at [`gatsby/package.json#L58`](https://github.com/matrix-org/matrix.org/blob/1397f009d4474a7001071bec5648a05cb58e0684/gatsby/package.json#L58)

Relevant docs, https://docs.npmjs.com/cli/v7/using-npm/scripts#path

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,11 @@ To discuss maintenance of this site, please come talk to the team in [#matrix.or
 ### How to build
 
 0. Make sure you have the prerequisites:
-   - npm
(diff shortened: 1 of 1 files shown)


ara4n:
lgtm

MadLittleMods:
Thanks for taking a look @ara4n 🐵

PR #986, opened 2021-05-08
2021-05-07-twim.mdx: New IRC command is !unlink

jaller94:
I had gotten that wrong. It should say `!unlink` instead of `!unbridge`.

https://github.com/matrix-org/matrix-appservice-irc/pull/1298


benparsons (approved):
lgtm

PR #987, opened 2021-05-12
Update "Riot->Element" in article title and remove backslashes

jaller94:
* Travis has renamed his article after the Riot->Element rebranding.

## Also changed
* In the article "Client Server API" the backslashes are visible.

## Unrelated notes
* I find the Feedback overlay to be incredibly annoying. Dismissing it brings it back several times even when staying on the same article. As a reader on a mobile phone, I wish it wasn't a floating overlay but a form at the bottom of the page.


benparsons (approved):
lgtm

PR #988, opened 2021-05-14
Hang link anchors off to the side so heading text aligns with body

MadLittleMods:
Hang link anchors off to the side so heading text aligns with body

Before | After
--- | ---
Screen Shot 2021-05-13 at 8 02 52 PM | Screen Shot 2021-05-13 at 8 01 53 PM

Responsive:


ara4n:
lgtm!

PR #989, opened 2021-05-18
Add Server_Stats Project

MTRNord:
Adding my 3 in 1 Project :) The linked FAQ is being written while I am making this PR and should be online shortly


MadLittleMods:
Tested locally and looks good 👍 

http://localhost:8000/docs/projects/other/server-stats-project

PR #1003, opened 2021-06-02
Update some VoIP info in FAQ

SimonBrandner:


gatsby/src/pages/faq.js
```diff
@@ -568,9 +568,8 @@ const Faq = ({ data }) => {
                 provides a simple chatroom interface to Matrix - letting the
                 user interact with users and rooms anywhere within the Matrix{" "}
                 <a href="#definitions">federation</a>. Text and image messages
-                are supported, and basic voice-only VoIP calling via WebRTC is
-                supported in one-to-one rooms. (As of October 2015, experimental
-                multi-way calling is also available on Element Web).
+                are supported, and VoIP calling via WebRTC is supported in
```
MadLittleMods, line 571: Should we mention video and voice calling? Or is VoIP sufficiently descriptive?

gatsby/src/pages/faq.js
```diff
@@ -568,9 +568,8 @@ const Faq = ({ data }) => {
                 provides a simple chatroom interface to Matrix - letting the
                 user interact with users and rooms anywhere within the Matrix{" "}
                 <a href="#definitions">federation</a>. Text and image messages
-                are supported, and basic voice-only VoIP calling via WebRTC is
-                supported in one-to-one rooms. (As of October 2015, experimental
-                multi-way calling is also available on Element Web).
```
MadLittleMods, line 573: I can start a video or voice call in any room and multiple people can join. I assume this was what the multi-way calling was?

Seems like we could remove the "in one-to-one rooms" qualifying statement and cover everything.

SimonBrandner: The reason I removed it was that it's not Matrix-based - it's Jitsi. And it is not in the spec, so it didn't seem entirely appropriate. Though maybe something like _Group/conference calls are possible using widgets such as Jitsi at the moment_ would be ok?

SimonBrandner: > I assume this was what the multi-way calling was?

I think the author actually meant the FreeSWITCH bridge or the mesh solution, both of which were removed due to issues

gatsby/src/pages/faq.js
```diff
@@ -568,9 +568,8 @@ const Faq = ({ data }) => {
...
+                are supported, and voice and video calling via WebRTC is supported in 
+                one-to-one rooms.
```
MadLittleMods, line 572:
```suggestion
                are supported, as well as voice and video calling via WebRTC in
                one-to-one rooms.
```

gatsby/src/pages/faq.js
MadLittleMods: @SimonBrandner Thanks for the context!


---

> And it is not in the spec, so it didn't seem entirely appropriate.

Fair enough, widgets are still in MSC phase :+1:

My thinking was we already call out that the voice/video is "via WebRTC" (signaling over Matrix). We could do the same and write "via Jitsi" (widget signaling happens over Matrix) for the rest of the rooms.

> Text and image messages are supported, as well as voice/video calling via WebRTC in one-to-one rooms and via Jitsi elsewhere.

MadLittleMods (commented):
Suggested a small change but feel free to reject. The current phrasing has a lot of "ands" and "supported".

gatsby/src/pages/faq.js
SimonBrandner: @MadLittleMods, in the end, I've decided to add it, it probably doesn't hurt anything. Thanks for all the suggestions!

MadLittleMods:
Looks good :+1: Thanks for the updates @SimonBrandner 🐔

PR #1006, opened 2021-06-10
Add low bandwidth tutorial

kegsay:
cc @neilalexander @ara4n

### DIFF
--- a/gatsby/content/blog/2021/06/2021-06-10-low-bandwidth-matrix-tutorial.mdx
+++ b/gatsby/content/blog/2021/06/2021-06-10-low-bandwidth-matrix-tutorial.mdx
@@ -1,0 +1,162 @@
+---
+  date: '2021-06-10'
+  title: 'Low Bandwidth Matrix: An implementation guide'
+  categories:
+    - Tutorials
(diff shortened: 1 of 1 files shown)


callahad:
Looks good to me. From an editorial standpoint I think it would be nice to have a _little_ more high-level exposition about what this is doing, for a lay audience before diving into the tutorial. Even just a paragraph explaining the motivation and answering architectural questions like what _is_ CoAP, why are we doing a proxy, will homeservers support this natively in the future, why is this setup useful right now, etc. A small diagram of the Matrix network with a CoAP proxy in front of one of the servers could also be nice to show the end result of the tutorial.

At one point, you say "Otherwise, congratulations! You now have a low bandwidth Matrix homeserver!" -- but you don't, really, you have a low-bandwidth proxy in front of someone else's homeserver (namely, matrix.org). It could reduce confusion to specifically address this in the text. E.g., you can connect to your proxy just like you would to matrix.org, etc. Explaining how the proxy does / doesn't look like its own device to m.o, etc.

kegsay:
@callahad PTAL

callahad:
Looks great!

PR #1009, opened 2021-06-14
Un-feature Pattle and Matrix Dart SDK.

dkasak:
These are no longer maintained and frequently confuse people, so it's best not to have them linked up front. Pattle is also no longer available in app stores.


benparsons (approved):
lgtm

Issue #138, opened 2018-04-19
add timepath's xonotic chat client


ara4n:
https://xonotic.gitlab.io/-/xonotic-data.pk3dir/-/jobs/1145499/artifacts/public/d0/d44/lib_2matrix_2matrix_8qc.html

benparsons:
Getting more info on what's been done.

ara4n:
for stuff like this i'd probably just get at least a placeholder on the site.  In this instance I'd link to Xonotic (http://www.xonotic.org), describe it as in-game chat, and probably alpha grade.  We can then flesh it out with more details as it comes in. It'd be a shame to have the entry lost entirely due to lack of details. (ping @TimePath to provide some details though - istr there was a really fun screenshot of matrix embedded in the gameplay :)

benparsons:
yep got a chat going :D

TimePath:
Bit of an update, there's some ongoing work to pull the usual chat UI out of the engine and into the game code. Once that's complete, it might start being used more seriously once we can figure out how we want to map our users to matrix users

aaronraimist:
This got added to the website at some point https://matrix.org/docs/projects/client/xonotic but it has basically no information on what it is

PR #1014, opened 2021-06-15
Update Xonotic to include some more juicy details

MadLittleMods:
Update Xonotic to include some more juicy details

Fix https://github.com/matrix-org/matrix.org/issues/138

---

http://localhost:8000/docs/projects/try-matrix-now
Screen Shot 2021-06-15 at 5 01 50 PM

http://localhost:8000/docs/projects/other/xonotic
Screen Shot 2021-06-15 at 5 01 56 PM


gatsby/content/projects/2018-04-24-xonotic.mdx
```diff
@@ -1,12 +1,17 @@
 ---
 layout: project
 title: xonotic
-categories:
- - client
-description: In-game chat
+categories: 
+ - other
```
MadLittleMods, line 5: Updated to `other` so it shows up on http://localhost:8000/docs/projects/try-matrix-now under the other category

Currently, this page doesn't appear anywhere AFAIK

PR #1016, opened 2021-06-21
Remove two old guides for installing Synapse

aaronraimist:
These both use Python 2 and link to out of date Riot links. They also make bad recommendations, the first one tells you to  use a matrix subdomain in your server name and the second one doesn't use Postgres.


richvdh (approved):
thanks!

PR #1021, opened 2021-07-05
Change nickname in HoF

0x1a8510f2:
I'd like to, if possible, change my Hall of Fame entry as I've changed my online nickname. Will reply to this from my old GitHub account to confirm.


TR-SLimey:
Can confirm, that is indeed me :P

MadLittleMods:
Seems reasonable :+1:

Checked it out locally:

http://localhost:8000/security-disclosure-policy/
Screen Shot 2021-07-06 at 6 11 50 PM

0x1a8510f2:
Thank you! :3

PR #1025, opened 2021-07-17
Remove page for Ruma homeserver

jplatte:
It's been dead for a while.


gatsby/content/projects/2015-11-29-ruma.mdx
```diff
@@ -1,19 +0,0 @@
...
-license: MIT
-repo: https://github.com/ruma/ruma
-home: https://www.ruma.io/
-room: "#ruma:matrix.org"
----
-
-The Ruma Rust-language homeserver project is [currently not maintained](https://github.com/ruma/ruma), but the *Ruma project*,
-including [ruma-api](https://github.com/ruma/ruma-api]) (framework for building Matrix APIs in Rust) and
-[ruma-client-api](https://github.com/ruma/ruma-client-api) (request and response types for the Matrix client API)
-are still being actively worked on.
```
MadLittleMods, line 19: Let's also remove it from https://github.com/matrix-org/matrix.org/blob/b93d07ab61ad670aed1fcbb6ea3348a424524d48/gatsby/src/pages/faq.js#L1389

gatsby/content/projects/2015-11-29-ruma.mdx
```diff
@@ -1,19 +0,0 @@
----
-layout: project
-title: Ruma
```
MadLittleMods, line 3: For reference on Ruma being dead: https://github.com/ruma/homeserver has been archived and links to https://www.ruma.io/news/ruma-is-dead-long-live-ruma-2020-04-10/


https://matrix.org/docs/projects/try-matrix-now has a lot of other "Not actively maintained" projects but we just have that box unchecked by default.

Seems like it would be ok to remove though 🤔

MadLittleMods (changes requested):

gatsby/content/projects/2015-11-29-ruma.mdx
jplatte: Yeah, that should be removed / replaced with a link to [Conduit](https://conduit.rs/).

gatsby/content/projects/2015-11-29-ruma.mdx
jplatte: I don't know the status of the other ones, not actively maintained could also mean that it might come back at some point, which is definitely not the case for the Ruma HS (there's too much confusion with the overall project, which is also why I posted this PR).

aaronraimist: Imo it is better to just set it as not actively maintained rather than deleting it even if you plan to never do anything with it. It’ll be hidden by default but is still findable if someone is looking for Rust code that does Matrix things that they can use for inspiration or just as an archive of what previously Matrix projects have existed.

jplatte: I'd really prefer it to be deleted, given the somewhat common confusion with the overall project of the same name.

MadLittleMods: For reference, @jplatte is a maintainer of Ruma. I'm inclined to go with your wishes although leaving it in "Not actively maintained" seems perfectly fine as well. I'll leave this for another day in case you want to respond, then merge.

The change to `gatsby/src/pages/faq.js` is good in any case.

PR #1039, opened 2021-08-08
Update instructions for using Slack bridge with Element

aaronraimist:


gatsby/content/projects/bridges/matrix-appservice-slack.mdx
```diff
@@ -36,6 +36,6 @@ The bridge is regularly hacked on and is accepting new PRs for bugfixes and feat
 ✓ Formatted Text Content  
 ✓ Media Content  
 
-When using [Element on app.element.io](https://app.element.io), you can enable Slack integration per-room. Click the “Manage Integrations” (four squares) button at the top of the page.
+When using [Element](https://element.io) with the default https://scalar.vector.im integrations manager, you can enable the Slack integration per-room. Click the “Room info” button in the top right corner of the page and then click "Add widgets, bridges & bots".
```
MadLittleMods, line 39: https://scalar.vector.im seems like a very weird thing to link. I just get a `Unable to initialise application. HTTP 403 Invalid token` response when trying to visit and learn more about Scalar. Is this supposed to go somewhere?

Where do I learn more about this? All I see is a private repo, https://github.com/matrix-org/scalar

If I setup my own homeserver and Element client, can I still use Scalar? My guess is this is a private EMS thing since that's what it shows if I try to set something up and the project repo is private.

aaronraimist: It is not a private EMS thing. https://scalar.vector.im is the public integrations manager hosted by you guys which anyone on homeserver can use.

Instead of linking to it I could just say "the default integrations manager" or "the default Scalar integrations manager". In the previous text it said you specifically had to be using https://app.element.io which is untrue. I just wanted to clarify that you can use any instance of Element as long as you are using that public integrations manager.

It isn't a completely foreign thing to mention though, it shows up in Element's settings for example:
screenshot

MadLittleMods: Thanks! It was unclear to me how this works and whether this was a client or homeserver thing. It looks like a client thing where it uses your credentials to do a bunch of actions. And I also just looked on my local Element instance which has `scalar.vector.im` set as the default 👍

```suggestion
When using [Element](https://element.io) with the default integrations manager (`scalar.vector.im`), you can enable the Slack integration per-room. Click the "Room info" button in the top right corner of the page and then click "Add widgets, bridges & bots".
```

(also got rid of the magic quotes in favor of normal quotes)

MadLittleMods:
@aaronraimist Thanks for the updates and the help understanding Scalar! The new instructions got me straight to the right place 👍

PR #1044, opened 2021-08-10
Added listing for matrix-faq bot

hwittenborn:
This bot serves the same function as [Matrixcoffee/FAQBot](https://matrix.org/docs/projects/bot/faq-bot) from an end-user perspective of actually interacting with the bot, but the backend for configuration has some notable differences, such as allowing the specification of a prefix (not sure if FAQBot allows that or not) and specifying the FAQ messages in Markdown instead of Org Mode.


afranke:
LGTM, cheers.

PR #1045, opened 2021-08-11
Fix Try Now link

### LINKED ISSUE #1027: 'Try now' button should link to a client picker these days

afranke:
Fixes #1027


afranke:
I wasn’t able to test the redirection locally because of https://github.com/gatsbyjs/gatsby/issues/13072, but it’s a simple enough page and the worst case is it doesn’t work, in which case we can just decide it’s not worth spending more time on this, remove it and move on.

The PR is now ready for a quick review and a merge.

gatsby/static/css/matrix-org.webflow.css
```diff
@@ -1358,9 +1354,6 @@ a {
   .mxnavbar__navlink.mxnavbar__navlink--primary {
     display: none;
   }
-  .mxnavbar__navlink.mxnavbar__navlink--try {
-    display: block;
-  }
```
MadLittleMods, line 1363: We also need to remove the [`display: none` rule for `.mxnavbar__navlink.mxnavbar__navlink--primary`](https://github.com/matrix-org/matrix.org/blob/8a2a43545a19e841542635d9c97ecae18e8f1cca/gatsby/static/css/matrix-org.webflow.css#L1358-L1360) so it shows up in the hamburger mobile menu.

It's currently missing with that rule in place:

Screen Shot 2021-08-16 at 9 49 15 PM

MadLittleMods (changes requested):
The simplification looks good. Looks like just one more thing to remove to get this going in all environments ⏩

gatsby/static/css/matrix-org.webflow.css
afranke: Good catch, fixed.

PR #1046, opened 2021-08-13
Add gatsby clean as npm script

afranke:
This way one can cleanup with `npm run clean` even though they don’t have gatsby installed system wide (as suggested by the README).

### DIFF
--- a/gatsby/package.json
+++ b/gatsby/package.json
@@ -54,6 +54,7 @@
   "main": "n/a",
   "scripts": {
     "build": "gatsby build",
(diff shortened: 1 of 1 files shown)


MadLittleMods:
@afranke We can also add `"gatsby": "gatsby",` so we can also easily run whatever other command as a one-off, `npm run gatsby -- someothercomamnd` for example

afranke:
Nice! I added it in #1049 which I’ll merge as soon as CI is done.

Issue #1060, opened 2021-08-23
Is upgrading a room still not recommended?


MadLittleMods:
Currently, the note about State Resets mentions that upgrading a room is not recommended,

> `/upgraderoom 6` command - but please note this is not recommended, as we are still testing the final room algorithm and upgrade mechanism for Matrix 1.0.
>
> *-- [`gatsby/content/docs/2019-04-29-moderation.mdx`](https://github.com/matrix-org/matrix.org/blob/c01cd843eb108106df2a236ec33969437a66d69e/gatsby/content/docs/2019-04-29-moderation.mdx#a-note-on-state-resets)*

Is this still the case nowadays as we seem to be doing it a lot without a problem for Spaces stuff in room versions, v7 and v8?

---

*Spawned from recommending this section to someone: https://gitlab.com/gitterHQ/webapp/-/issues/2792#note_658288382*

aaronraimist:
Element definitely still has bugs where it doesn't notify you that a room upgrade has occurred https://github.com/vector-im/element-web/issues/11932

MadLittleMods:
Thanks @aaronraimist!

From some out of band discussion with @turt2live as well, the upgrade mechanism itself is fine nowadays but as you mentioned the UX around upgrades has many bugs, https://github.com/vector-im/element-web/labels/A-Room-Upgrades

And the goal being to make it seamless: stitching timelines together, autojoin people, etc.

I think it's probably worth updating the qualifying note around why it's not totally perfect, vs still in the testing stage.

PR #1061, opened 2021-08-24
Update qualifying statement why upgrading a room is probably fine if you need to

MadLittleMods:
Update qualifying statement about why upgrading a room is probably fine if you need to. The upgrade process itself is not in the testing stage anymore but does still have some rough edges when it comes to the user experience.

Fix https://github.com/matrix-org/matrix.org/issues/1060

### DIFF
--- a/gatsby/content/docs/2019-04-29-moderation.mdx
+++ b/gatsby/content/docs/2019-04-29-moderation.mdx
@@ -117,7 +117,16 @@ For further detailed info about server ACLs, see [https://matrix.org/docs/spec/c
 
 ### A note on state resets
 
-Prior to Matrix 1.0, a bug existed which caused room state (e.g. membership, power levels, bans/kick data) to intermittently be reset to a historical value due to a bad merge conflict resolution - a so-called ‘state reset’, or the ‘hotel california bug’.  This was improved in v2 and later of the room algorithm, which will be used as the default when creating new rooms in Matrix 1.0.  However, as of April 2019 most existing rooms still use the original buggy v1 algorithm.  The best way for now to handle a state reset is to manually re-apply any bans & ACLs if they are reset.  However, if state resets keep happening, it is possible to upgrade the room algorithm via Element/Web with the undocumented `/upgraderoom 6` command - but please note this is **not** recommended, as we are still testing the final room algorithm and upgrade mechanism for Matrix 1.0.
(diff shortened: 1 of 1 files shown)


gatsby/content/docs/2019-04-29-moderation.mdx
```diff
@@ -117,7 +117,16 @@ For further detailed info about server ACLs, see [https://matrix.org/docs/spec/c
 
 ### A note on state resets
 
-Prior to Matrix 1.0, a bug existed which caused room state (e.g. membership, power levels, bans/kick data) to intermittently be reset to a historical value due to a bad merge conflict resolution - a so-called ‘state reset’, or the ‘hotel california bug’.  This was improved in v2 and later of the room algorithm, which will be used as the default when creating new rooms in Matrix 1.0.  However, as of April 2019 most existing rooms still use the original buggy v1 algorithm.  The best way for now to handle a state reset is to manually re-apply any bans & ACLs if they are reset.  However, if state resets keep happening, it is possible to upgrade the room algorithm via Element/Web with the undocumented `/upgraderoom 6` command - but please note this is **not** recommended, as we are still testing the final room algorithm and upgrade mechanism for Matrix 1.0.
+Prior to Matrix 1.0 in room version 1, a bug existed which caused room state
```
aaronraimist, line 120: This wording makes it sound like Matrix 1.0 may have fixed the bug even in room version 1 which isn't true. You can probably just remove "Prior to Matrix 1.0".

```suggestion
In room version 1, a bug existed which caused room state
```

gatsby/content/docs/2019-04-29-moderation.mdx
```diff
@@ -117,7 +117,16 @@ For further detailed info about server ACLs, see [https://matrix.org/docs/spec/c
...
+(e.g. membership, power levels, bans/kick data) to intermittently be reset to
+a historical value due to a bad merge conflict resolution - a so-called
+'state reset', or the 'hotel california bug'. This was improved in room
+version 2 onwards and room state should no longer diverge. However, as of
+April 2019, most existing rooms still use the original buggy v1 algorithm.
```
aaronraimist, line 125:
```suggestion
version 2 onwards and room state should no longer diverge. However, many
old rooms still use the original buggy v1 algorithm.
```

MadLittleMods:
Thanks for the review @aaronraimist 🐅

PR #1063, opened 2021-08-25
👷 Port CI to Github Actions

afranke:
This is intended to replace the existing Buildkite one, but the latter can be removed at a latter point when all the bits are in place. The workflow can be seen working for the gh-actions branch at [the Actions tab](https://github.com/matrix-org/matrix.org/actions).

### DIFF
--- a/.github/workflows/build-matrix.org.yml
+++ b/.github/workflows/build-matrix.org.yml
@@ -7,6 +7,8 @@ on:
   workflow_dispatch:
 
 jobs:
+  # Fetch the latest spec build, and apply styling to it.
+  # This step also sets up the OpenAPI UI.
   build-spec:
     name: "📖 Spec"
     runs-on: ubuntu-latest
@@ -37,6 +39,8 @@ jobs:
           name: styled-docs-artifact
           path: styled-docs.tar.gz
 
+  # Run jekyll. We run this in a custom docker image.
+  # See dockerfile in the `.buildkite` directory.
   build-jekyll:
     name: "🌐 Jekyll"
     runs-on: ubuntu-latest
@@ -58,6 +62,8 @@ jobs:
           name: spec-artifact
           path: jekyll-site.tar.gz
 
+  # Run gatsby. We also use a custom docker image for this, to avoid
+  # installing the entirety of npmjs for each build.
   build-gatsby:
     name: "🌐 Gatsby"
     runs-on: ubuntu-latest
@@ -104,6 +110,8 @@ jobs:
           name: guides-artifact
           path: implementation-guides/implementation-guides.tar.gz
 
+  # Merge the three sets of outputs together, together
+  # with the existing static stuff in `content`.
   package:
     name: "📦 Package"
     runs-on: ubuntu-latest
--- a/.github/workflows/build-matrix.org.yml
+++ b/.github/workflows/build-matrix.org.yml
@@ -1,6 +1,35 @@
 name: "🌐 Build the matrix.org website"
 on: [workflow_dispatch, push, pull_request]
 jobs:
+  build-spec:
+    name: "📖 Spec"
+    runs-on: ubuntu-latest
+    container: "python:3.6"
+    steps:
+      - name: "📥 Source checkout"
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+      - name: "📥 matrix-doc OpenAPI artifact download"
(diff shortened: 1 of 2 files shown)


afranke:
It breaks the Buildkite CI because [I had to remove the part of `update_docs.sh` where the artifact is downloaded from Buildkite](https://github.com/matrix-org/matrix.org/pull/1063/commits/550064881787d48d807b0303ee01165191a592ef#diff-aaea39b0e0018c2647c12138816ce874adc0ffaa452c76668479503089e7fd01L26)

.github/workflows/build-matrix.org.yml
```diff
@@ -0,0 +1,133 @@
...
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+      - name: "📥 matrix-doc OpenAPI artifact download"
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: main.yml
+          workflow_conclusion: success
+          repo: ${{ github.repository_owner }}/matrix-doc
+          branch: ${{ github.ref }}
```
afranke, line 19: This attemps to download the OpenAPI definition from the matrix-doc repo, from a branch with the same name as the current branch. I.e. the gh-actions branch for matrix.org tries to download the artifact from the gh-actions branch in the matrix-doc repo. When we merge this, master will reach for master.

This breaks with pull requests though, as can currently be seen where `refs/pull/1063/merge` is only relevant in this repository and not the matrix-doc one. I reckon we can change this to master *as a subsequent PR once this is merged*.

.github/workflows/build-matrix.org.yml
```diff
@@ -0,0 +1,133 @@
+name: "🌐 Build the matrix.org website"
+on: [workflow_dispatch, push, pull_request]
```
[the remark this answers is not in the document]
afranke: Fixed.

.github/workflows/build-matrix.org.yml
afranke: And now that [matrix-doc moved to a `main` branch](https://github.com/matrix-org/matrix-doc/issues/3367) this is broken before it even happened.

Should I just hardcode `main`? This means it will be broken for the context of the pull request, but will work once everything is merged. For what it’s worth, there is evidence this worked in previous runs, so we’re not running blind here.

.github/workflows/build-matrix.org.yml
```diff
@@ -0,0 +1,137 @@
...
+          env/bin/pip install requests
+          . env/bin/activate; scripts/update_docs.sh
+      - name: "📦 Tarball creation"
+        run: tar -czf styled-docs.tar.gz content/docs
+      - name: "📤 Artifact upload"
+        uses: actions/upload-artifact@v2
+        with:
+          name: styled-docs-artifact
+          path: styled-docs.tar.gz
+  build-jekyll:
```
richvdh, line 37: please could you copy over the comments from the BK pipeline - https://github.com/matrix-org/matrix.org/blob/master/.buildkite/pipeline.yaml#L18 for example. Likewise the other comments in that file are worth keeping.

(indeed, there's a particular gotcha here in that the Dockerfile is in the `.buildkite` directory. We should move it eventually, but let's leave a comment for now)

richvdh (changes requested):

scripts/update_docs.sh
```diff
@@ -21,9 +21,13 @@ cd "$(dirname "$(dirname "${SELF}")")"
 
 SITE_BASE="$(pwd)"
 
-# grab and unpack the latest matrix-docs build from buildkite
-rm -rf assets.tar.gz assets
-scripts/fetch-buildkite-artifact matrix-dot-org matrix-doc assets.tar.gz
+if [ -n "$BUILDKITE" ]; then
```
afranke, line 24: I checked the documentation before applying your suggestion, @richvdh, but [Buildkite doesn’t seem to like that variable](https://buildkite.com/matrix-dot-org/matrix-dot-org/builds/3266#bbbeddd9-0eea-47c5-919d-d0ca5722a18f): it claims it is “unbound”, even though the “Environment” tab shows it first in the list. Any clue what’s going on?

afranke: Updated the test, now it runs again both on BK and GHA.

richvdh (approved):
lgtm. Thanks!

PR #1066, opened 2021-08-31
Remove outdated Gitter bridge reference

MadLittleMods:
Remove outdated [`matrix-appservice-gitter`](https://github.com/matrix-org/matrix-appservice-gitter) reference.

Don't want to confuse people on how the new Gitter bridge works as this old bridge doesn't even run anymore.

### DIFF
--- a/gatsby/content/docs/2017-03-11-types-of-bridging.mdx
+++ b/gatsby/content/docs/2017-03-11-types-of-bridging.mdx
@@ -50,7 +50,7 @@ Several obstacles exist to the proper implementation of double-puppeted bridges.
 
 ### Hybrid Relaybot Puppet Bridge
 
-This type of bridge is a combination single or double puppet bridge which tries to solve the problem of representing other users by means of the bridgebot technique.
+This type of bridge is a combination single or double puppet bridge which tries to solve the problem of representing other users by means of the bridgebot technique. [mautrix/telegram](https://github.com/mautrix/telegram) works in this way.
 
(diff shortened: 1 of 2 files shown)


gatsby/content/docs/2017-03-11-types-of-bridging.mdx
```diff
@@ -50,7 +50,7 @@ Several obstacles exist to the proper implementation of double-puppeted bridges.
 
 ### Hybrid Relaybot Puppet Bridge
 
-This type of bridge is a combination single or double puppet bridge which tries to solve the problem of representing other users by means of the bridgebot technique. [matrix-appservice-gitter](https://github.com/matrix-org/matrix-appservice-gitter) works in this way.
```
MadLittleMods, line 53: Is there another bridge that works this way that we can reference here instead?

MadLittleMods: It looks like https://github.com/mautrix/telegramcould fits the bill.

MadLittleMods:
Thanks @afranke! Updated the FAQ 👍

I've left `gatsby/content/projects/2016-08-10-matrix-appservice-gitter-twisted.mdx` alone as I don't know what that authors plans are (if they actually want to deprecate in favor of the native bridge). It's already marked as `maturity: Not actively maintained` and not `featured` so it doesn't show up. I've never even heard of it for example.

afranke:
Fair enough.

PR #1072, opened 2021-09-07
Tidy projects/ files and languages

### LINKED ISSUE #1035: Duplicated languages on Try Matrix Now
The “Let’s go deeper” section of [Try Matrix Now](https://matrix.org/docs/projects/try-matrix-now) page has duplicated languages (`JavaScript` vs `javascript`, `Python 3` vs `Python3`). The list is already long as it is, some cleanup to remove duplicates would be welcome.

Zopieux:
This PR tidies up the "try matrix now" project repository. Changes are split in atomic commits, please see descriptions. Fixes #1035.

Language noise:
* there was three (3) different spellings for Go
* there was two different spellings for JavaScript
* the Python / Python 3 distinction makes little sense in 2021

File management:
* there was no subfolder for application services (`as`)
* most projects were stored in the root level instead of using subfolders; renamed everything, without affecting slugs (stable URLs)

Other:
* one project (opsdroid) incorrectly provided two categories in one line
* the README was not up-to-date with all available categories
* refactored the filtering logic so any category can appear at any position in the category array


afranke:
Thank you for the cleanup!

I had started doing this locally, but it was not as advanced yet. I’ll check if you missed anything that I had while reviewing your PR and will let you know if it’s the case.

afranke (changes requested):
Nitpick for your commit messages: please don’t add full stops at the end.

We had almost the same content for your language fixes. The only thing I had that you didn’t change is in `gatsby/content/projects/2016-12-11-riotchat-ansible.mdx` where I changed `Playbook` (not a language name) to `Ansible`. Technically it’s YAML, but that wouldn’t be useful to anyone. Can you add that change?

You claim to be moving projects with multiple categories to the directory of the most relevant ones, but I’d argue that appservice bridges belong in the bridges category more than they belong in the as one.

afranke:
Please edit the PR description to inclued `Fixes #1035` as well, so it gets closed when this is merged. See [the relevant documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

Zopieux:
Thanks for the quick review.

> You claim to be moving projects with multiple categories to the directory of the most relevant ones, but I’d argue that appservice bridges belong in the bridges category more than they belong in the as one.

I'm confused, none of the files moved to the new `as/` folder have a category other than `as`, so I did not think this could be controversial. The only files with more than one category for which I made a judgment call are: 

```
bots/feedbot.mdx
bots/maubot.mdx
bots/opsdroid-connector.mdx
bridges/matrix-keybase.mdx
bridges/matrix-xmpp-bridge.mdx
bridges/matrix-zulip-bot.mdx
```

Please review these and provide a final decision. I have no strong opinion. We could also not categorize them at all and leave them at root level -- your call.

You might be confused by the fact I put eg. `matrix-xmpp-bridge` (`as`, `bridge`) under `bridges` but since `as` is its *first* category, its slug computes to `as/matrix-xmpp-bridge`. I did not think of that aspect in this PR. Should I further make sure the *first* category matches the subfolder?

tulir:
> I'm confused, none of the files moved to the new as/ folder have a category other than as, so I did not think this could be controversial.

They were just misclassified to begin with. Looking at the directory, all of them seem to be either bridges or SDKs.

Zopieux:
I see. I reclassified all `as` into either `bridge` or `sdk`. PTAL.

afranke:
Indeed, I was looking at the paths and saw bridges with appservice in their names being moved to the `as` directory. I didn’t check what the categories in the files were actually.

So… now if I go to http://localhost:8000/docs/projects/try-matrix-now and check only Application Services but none of the other project types, I get no result ([matrix.org](https://matrix.org/docs/projects/try-matrix-now) gives me 3). Shouldn’t I get at least matrix-keybase and matrix-xmpp-bridge which both still have the `as` category?

Sorry for being such a PITA about this, but I don’t think removing `as` and putting `bridge` *instead* is correct. I think all these appservice-bridges belong in both categories. The `bridge` one should be *added*, not replacing `as`.

… unless we get rid of the `as` category altogether? @benparsons any thoughts on this? It seems bots+bridges are enough to cover all AS today, but that may not be very future-proof.

afranke (approved):

benparsons:
my 2c: I would be happy to see the AS category removed. That page doesn't need to show what *technology* the project uses (other than language), so much as what the project *is*.

afranke:
Sounds good. @Zopieux can you go ahead and remove the AS category completely? Forget what I said about having both `bridge` and `as` for the files you changed.

Zopieux:
Better?

afranke:
`git grep` in `content/projects` still shows me some with the as category (matrix-keybase for instance).

Zopieux:
Indeed, fixed.

afranke:
Awesome, thank you very much!

PR #1073, opened 2021-09-08
Update the free matrix server article/guide

ptman:
Is shell or sh the right choice for syntax highlighting shell code? How do you feel about removing the 'small' from the name? Oracle now offers enough resources to run a quite beefy matrix server.

### DIFF
--- a/gatsby/content/docs/free-small-matrix-server.mdx
+++ b/gatsby/content/docs/free-small-matrix-server.mdx
@@ -1,5 +1,5 @@
 ---
-title: 'Free Small Matrix Server'
+title: 'Free Matrix Server'
 author: Paul Tötterman <paul.totterman@iki.fi>
 layout: post
 section: Server Setup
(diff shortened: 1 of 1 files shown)


gatsby/content/docs/free-small-matrix-server.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: 'Free Small Matrix Server'
```
benparsons, line 2: Maybe we could change the title to "Free Matrix Server using Oracle Cloud", or similar? Suggesting this because "Free Matrix Server" feels ambiguous - matrix.org is a "Free[ to use] Matrix Server"

benparsons (changes requested):
Only one point for me, the title

gatsby/content/docs/free-small-matrix-server.mdx
ptman: sure

gatsby/content/docs/free-small-matrix-server.mdx
```diff
@@ -113,18 +115,6 @@ snap remove oracle-cloud-agent
 apt purge snapd open-iscsi lxd lxcfs
 ```
 
-## Tune server (optional)
-
-I suggest enabling swap, since there's only 1 GB of RAM.
-
-```shell
```
MadLittleMods, line 120: > Is shell or sh the right choice for syntax highlighting shell code?

It looks like we use `shell` a lot more than `sh`

gatsby/content/docs/free-small-matrix-server.mdx
```diff
@@ -1,5 +1,5 @@
 ---
```
MadLittleMods, line 1: Looks good on http://localhost:8000/docs/guides/free-small-matrix-server ✅

We could also add a link to it from http://localhost:8000/docs/guides/ which seems to be manually curated https://github.com/matrix-org/matrix.org/blob/9255200369fab9fbff211aa871cf993773139e70/gatsby/src/pages/docs/guides.js#L138-L156

MadLittleMods:
Thanks for refreshing the guide with the latest and greatest 🐎

PR #1074, opened 2021-09-10
Pre-announce coordinated release.

dkasak:


afranke (approved):
From a stylistic point of view, I’d rather have **strong content** than the strong tag used, but it’s not a big deal. Apart from that nitpick, it looks good to me.

PR #1079, opened 2021-09-21
Warn about deprecated article and stop linking to it

jplatte:
Came across this today because sbd. posted about trying some of the curl requests from the article against Conduit and that failed because a lot of it is deprecated and Conduit doesn't implement it.

Only just noticing that my editor also removing some trailing spaces, I can remove those if you want.


afranke:
Do you know of an up to date guide that we could link to from the warning? I hate it when docs tell me not to use them but don’t tell me what to use instead.

jplatte:
No, unfortunately I don't.

PR #1081, opened 2021-09-24
💄 Switch from Swagger UI to RapiDoc

afranke:
For a more modern interface to explore the API.


afranke:
# Before

image

# After

image

# Notes

* RapiDoc supports custom API servers, but getting an entry like the one we had before requires adding metadata to the spec definition. I’ll take care of that in a separate PR to [the matrix-doc repository](https://github.com/matrix-org/matrix-doc/). I don’t consider this a blocker.
* I didn’t style RapiDoc with the website header/navbar **on purpose** because quite frankly it’s ugly in this context, gets in the way, and I doubt someone really needs to jump from the playground **directly** to the blog… The `[matrix]` logo in the top left corner is a link back to [the Develop section](https://matrix.org/docs/develop), where visitors can then find the navbar again.
* RapiDoc supports both Swagger 2.0 and OpenAPI 3.x.x, so we won’t have to change anything when we make that move eventually.

scripts/swagger-ui.patch
```diff
@@ -1,80 +0,0 @@
---- api/client-server/index.html	2017-08-03 11:16:08.456850097 +0100
```
anoadragon453, line 1: Looks like [content/swagger.css](https://github.com/matrix-org/matrix.org/blob/master/content/swagger.css) can go as well.

scripts/update_docs.sh
```diff
@@ -4,8 +4,8 @@
 # At present this includes:
```
anoadragon453, line 4: Should we change [scripts/update_docs_local.sh](https://github.com/matrix-org/matrix.org/blob/master/scripts/update_docs_local.sh) as well? Or does that file just seem redundant in general (and we should remove it)?

rapidoc/index.html
```diff
@@ -0,0 +1,30 @@
+<!doctype html> <!-- Important: must specify -->
```
anoadragon453, line 1: For instances like:

https://github.com/matrix-org/matrix.org/blob/a53ecfc9dad7afc939aab96ff5a903bb8c64f51e/unstyled_docs/spec/index.html#L1111

Would we still call RapiDoc a "Swagger Viewer", or should we switch entirely to the terms OpenAPI/RapiDoc?

anoadragon453 (changes requested):
Some minor clarifying points, but otherwise looks good to me.

Design may want to modify the colour scheme at some point, but from your screenshot I think it looks pretty slick for a first pass (though I have no idea about accessibility for instance).

rapidoc/index.html
afranke: I’d say OpenAPI indeed. Lemme fix it real quick.

scripts/update_docs.sh
afranke: I have actually never used it, and I suspect other things in the scripts are not in sync either. Whether we remove it or get it more in line with the non local one, we can do this in another PR?

scripts/swagger-ui.patch
afranke: Gone.

scripts/update_docs.sh
afranke: I had a quick look, it’s indeed out of sync and is basically a(n old) copy with some commented out steps. I reckon a better solution would be to remove the script altogether and make some steps optional in the main one.

Out of scope here.

anoadragon453: Yep, that was my hunch and thanks for checking. Yes, we can clean it up separately.

anoadragon453 (approved):

PR #1088, opened 2021-10-05
🚀 Publish previews on netlify

afranke:
This *should* publish a preview on netlify for every Actions run, with a link to it.

### DIFF
--- a/.github/workflows/build-matrix.org.yml
+++ b/.github/workflows/build-matrix.org.yml
@@ -163,4 +163,17 @@ jobs:
         with:
           name: merged-content-artifact
           path: content.tar.gz
+      - name: "ℹ Save PR metadata"
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/pr.json', JSON.stringify(context.payload.pull_request));
+      - name: Upload PR Info
+        uses: actions/upload-artifact@v2
+        with:
+          name: pr.json
+          path: pr.json
+            # We'll only use this in a triggered job, then we're done with it
+          retention-days: 1
 
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -5,7 +5,7 @@ on:
     types:
       - completed
 jobs:
-  build:
+  deploy-preview:
     runs-on: ubuntu-latest
     if: >
       ${{ github.event.workflow_run.conclusion == 'success' }}
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -21,30 +21,30 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               run_id: ${{github.event.workflow_run.id }},
-              });
-              var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
-                return artifact.name == "merged-content-artifact"
-                })[0];
-                var download = await github.actions.downloadArtifact({
-                  owner: context.repo.owner,
-                  repo: context.repo.repo,
-                  artifact_id: matchArtifact.id,
-                  archive_format: 'zip',
-                });
-                var fs = require('fs');
-                fs.writeFileSync('${{github.workspace}}/content.zip', Buffer.from(download.data));
+            });
+            var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "merged-content-artifact"
+            })[0];
+            var download = await github.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+            });
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/content.zip', Buffer.from(download.data));
 
-                var prInfoArtifact = artifacts.data.artifacts.filter((artifact) => {
-                  return artifact.name == "pr.json"
-                })[0];
-                var download = await github.actions.downloadArtifact({
-                  owner: context.repo.owner,
-                  repo: context.repo.repo,
-                  artifact_id: prInfoArtifact.id,
-                  archive_format: 'zip',
-                });
-                var fs = require('fs');
-                fs.writeFileSync('${{github.workspace}}/pr.json.zip', Buffer.from(download.data));
+            var prInfoArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "pr.json"
+            })[0];
+            var download = await github.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: prInfoArtifact.id,
+              archive_format: 'zip',
+            });
+            var fs = require('fs');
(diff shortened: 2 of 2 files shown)


afranke:
The [second workflow will only be run if the file is in the default branch](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run), so we can’t rely on this PR to check whether it works correctly. The changes to the build workflow have not broken it. @richvdh do you reckon I can merge this and fix it afterwards if it turns out not to work as expected?

.github/workflows/build-matrix.org.yml
```diff
@@ -163,4 +163,17 @@ jobs:
         with:
           name: merged-content-artifact
           path: content.tar.gz
+      - name: "ℹ Save PR metadata"
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/pr.json', JSON.stringify(context.payload.pull_request));
```
richvdh, line 171: how will this behave for builds not associated with a PR?

.github/workflows/netlify.yml
```diff
@@ -0,0 +1,83 @@
...
+      - completed
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      # There's a 'download artifact' action but it hasn't been updated for the
+      # workflow_run action (https://github.com/actions/download-artifact/issues/60)
+      # so instead we get this mess:
```
richvdh, line 15: wouldn't it be easier just to put these steps in the other workflow?

.github/workflows/netlify.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: "📥 Download artifact"
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var artifacts = await github.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: ${{github.event.workflow_run.id }},
+              });
+              var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
```
richvdh, line 25: there's something weird going on with the indentation here.

.github/workflows/netlify.yml
```diff
@@ -0,0 +1,83 @@
...
+                var download = await github.actions.downloadArtifact({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  artifact_id: prInfoArtifact.id,
+                  archive_format: 'zip',
+                });
+                var fs = require('fs');
+                fs.writeFileSync('${{github.workspace}}/pr.json.zip', Buffer.from(download.data));
+      - name: "📦 Extract Artifacts"
+        run: unzip content.zip && rm content.zip && tar -xzvf content.tar.gz && rm content.tar.gz && unzip pr.json.zip && rm pr.json.zip
```
richvdh, line 49: maybe put these on different lines?

richvdh (changes requested):
do we not need to inhibit the netlify upload for non-PR builds?

.github/workflows/netlify.yml
```diff
@@ -0,0 +1,83 @@
...
+                  return artifact.name == "pr.json"
+                })[0];
+                var download = await github.actions.downloadArtifact({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  artifact_id: prInfoArtifact.id,
+                  archive_format: 'zip',
+                });
+                var fs = require('fs');
+                fs.writeFileSync('${{github.workspace}}/pr.json.zip', Buffer.from(download.data));
```
richvdh, line 47: if we really need this stuff.. maybe wee should write our own action for it.

.github/workflows/netlify.yml
afranke: If we do that and for some reason netlify fails, then the build workflow status won’t be `success`.

.github/workflows/netlify.yml
afranke: Should be fixed.

.github/workflows/netlify.yml
afranke: Probably, although the ideal scenario really would be for [the download-artifact action](https://github.com/actions/download-artifact) to be fixed, as pointed out by the comment. This comes from [matrix-react-sdk](https://github.com/matrix-org/matrix-react-sdk/blob/develop/.github/workflows/netlify.yaml) BTW.

Can we consider this out of scope for this PR though?

.github/workflows/netlify.yml
richvdh: well yes, but how much is that really a problem?

.github/workflows/netlify.yml
richvdh: > Probably, although the ideal scenario really would be for [the download-artifact action](https://github.com/actions/download-artifact) to be fixed, as pointed out by the comment. 

Sure, but given the issue has been open a year with no reply from the maintainers, let's not hold our breath on that one.

> This comes from [matrix-react-sdk](https://github.com/matrix-org/matrix-react-sdk/blob/develop/.github/workflows/netlify.yaml) BTW.

ok, so this is at least our second copy of it!

> Can we consider this out of scope for this PR though?

I guess so. I do think we should do something else before we make a third copy.

My understanding is that once you've got the javascript, writing an action is pretty easy.

afranke:
> do we not need to inhibit the netlify upload for non-PR builds?

I would love to. I really can’t tell from the GHA documentation whether it’s possible though.

There is no documented `on.workflow_run.branches-ignore` as far I can see, only [`on.<push|pull_request>.branches-ignore`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) but then the examples for [`workflow_run`](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run) use `branches` which is not documented either. Shall I just try to use `on.workflow_run.branches-ignore` and :crossed_fingers: ?

.github/workflows/build-matrix.org.yml
afranke: It seems this will fail because `prInfoArtifact.id` is not defined (https://github.com/matrix-org/matrix-react-sdk/runs/3805077607?check_suite_focus=true would be an example of that).

We should probably handle this more gracefully.

There is no way to exit a job early, so that means that after the script step we would need to have an `if` for every subsequent step. Not great (but not necessarily an issue, just something to be aware of).

If [`on.workflow_run.branches-ignore`](https://github.com/matrix-org/matrix.org/pull/1088#issuecomment-934611174) does not exist and work, I reckon we have to run the script step in all cases as I don’t see a way to determine whether there was a PR before we try to download the PR metadata artifact.

afranke: I misread the code this was referring to and thought we were talking about the script step of the `netlify.yml` workflow. Will read the documentation further.

afranke:
Considering the concerns that were raised here (and which I share), I’ll write another version of this that does everything in the initial workflow. I will do it on a different branch and open a new PR, and we can decide which one to run with once we have both.

afranke:
We went with #1089 instead.

PR #1092, opened 2021-10-14
Stop trying to publish the unstable spec

richvdh:
This hasn't really worked properly since matrix-doc was redesigned to use hugo, but we still had a few vestiges of the old code hanging around. In particular, there is no need to fetch the matrix-doc assets, and no need to attempt to style them.

I've also removed `update_docs_local.sh` since its purpose was unclear.

Preview: https://pr1092--matrix-org-previews.netlify.app


afranke:
If `build-spec` is effectively going to do the same thing to the same committed files over and over again, considering any future version of the spec goes through a different process now, is there a point in keeping this at all? Couldn’t we just commit the styled files once and for all?

richvdh:
> If build-spec is effectively going to do the same thing to the same committed files over and over again, considering any future version of the spec goes through a different process now, is there a point in keeping this at all? Couldn’t we just commit the styled files once and for all?

I'm reluctant to do this, as it ties the spec to the same rendering forever more, even if the matrix.org homeserver gets a rebrand. And yes, in future, maybe we'll find alternative hosting for old versions of the spec, but we aren't there yet, so for now let's keep it as-is.

PR #1108, opened 2021-11-03
Jekyll cleanup

afranke:
THIS IS NOT READY YET (but I want CI to check the result).


afranke:
Not what I meant to do. *jedi hand pass* this didn’t happen

PR #1120, opened 2021-11-19
Update Nheko data

deepbluev7:
Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>

I have no idea, what dimension the screenshots should be.


afranke:
The files you sent look fine. I am not aware of a defined standard and looking at the ones we have currently, you’re in the ballpark. Thanks for the update.

PR #1122, opened 2021-11-26
Update icon for matrix-puppeteer-line

AndrewFerr:
Use an unmodified version of the LINE logo instead of the "in matrix"
version, to match how most bridge icons use the logo of the service
being bridged (rather than a custom icon for the bridge itself).


afranke:
Thank you.

PR #1126, opened 2021-11-26
Jekyll cleanup

### LINKED ISSUE #1047: Remove leftovers from Jekyll
[`jekyll/`](../tree/f363d9536bac0526144cdc3efcd04bc79433376a/jekyll) is still around even though the site has switched to gatsby and (most of) the content was moved.

We should check whether any bits still need to be moved, and then finish by cleaning up.

* [ ] check if anything needs to be moved
* [ ] remove content that’s not relevant anymore (e.g. [the old version of Try Matrix Now](../blob/f363d9536bac0526144cdc3efcd04bc79433376a/jekyll/_posts/projects/2016-02-05-try-matrix-now.md) which is superseded by [the current version](../blob/f363d9536bac0526144cdc3efcd04bc79433376a/gatsby/src/pages/docs/projects/try-matrix-now.js))
* [ ] clean up (e.g. remove jekyll from Buildkite)

afranke:
This is now ready to be merged.
I tried to have clear steps as much as I could, but pulling a thread on one end of this makes a knot on the other end, so some things may not be in the most obvious order at first. Let me know if you have any doubts about something I made.

Given the potential for breakage, I did thoroughly check whenever I removed anything and made sure that either the resource was already not accessible (e.g. an apache redirect takes over the URL) or a new redirect was created in gatsby.

Fixes #1047


afranke:
Do note that the before/after artifact will differ, because some files were included but were useless.

callahad:
Post-merging we should delete / clean up / archive obsolete container images on Docker Hub

PR #1131, opened 2021-11-30
Update FluffyChat screenshots

krillefear:


afranke:
I’m not really comfortable with the text at the top. Can we have screenshots that are actually just screenshots, like the previous ones were?

krillefear:
> I’m not really comfortable with the text at the top. Can we have screenshots that are actually just screenshots, like the previous ones were?

Yes! Sure. Sorry I had no special intention with this text. I was just lazy and have copy pasted two of the screenshots after some hours of creating those (I don't like creating screenshots). I will change this asap

krillefear:
Would the colored background still be okay? I can remove the background as well if it would fit better in the overall design of the clients page (just being lazy again after all the work)

afranke:
I couldn’t have asked for anything better than that. Thank you so much!

PR #1136, opened 2021-12-02
👷 Split out Netlify deployment

afranke:
This is based on the file available in the matrix-doc repository. The only changes are the artifact name and the comments to match what we deploy.

A quick sanity check would be welcome (but we know that CI won’t use this until it is merged, so don’t trust it to tell us if anything is wrong).


richvdh (approved):
:shrug: seems reasonable to me

PR #1141, opened 2021-12-06
👽️ Use new public definition file

afranke:
This depends on https://github.com/matrix-org/matrix-doc/pull/3542 and should not be merged before it lands.

This will also solve issues with broken links.

Fixes https://github.com/matrix-org/matrix-doc/issues/3507

Preview: https://pr1141--matrix-org-previews.netlify.app

### DIFF
--- a/rapidoc/index.html
+++ b/rapidoc/index.html
@@ -6,7 +6,7 @@
   </head>
   <body>
     <rapi-doc
-      spec-url = "json/api-docs.json"
+      spec-url = "https://spec.matrix.org/client-server-api/api.json"
       show-header="false"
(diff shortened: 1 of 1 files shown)


psrpinto:
Thanks for working on this @afranke.

I noticed that https://matrix.org/docs/api/client-server/ now fails to load https://spec.matrix.org/client-server-api/api.json, apparently due to missing CORS headers:

```
CORS Missing Allow Origin
```

afranke:
The file has not been backported for the already deployed stable spec *yet*. I noticed it after merging this. Should be done soon.

PR #1149, opened 2021-12-12
New .NET menu category and yet another C# SDK

m-kus:


afranke:
I broke the CI part that provides Netlify previews right before you rebased. It should be fixed now, so rebasing once more would be great, if you don’t mind. Sorry.

PR #1216, opened 2021-12-31
Privacy Notice: Correct Element Privacy Policy URL

jaller94:
Without the schema this is seen as a relative link: https://matrix.org/legal/element.io/privacy

## Note
In the [original document](https://github.com/vector-im/policies/blob/master/docs/matrix-org/privacy_notice.md) this is not a link.

Preview: https://pr1216--matrix-org-previews.netlify.app


afranke:
Thanks.

PR #1226, opened 2022-01-15
Add GitHub Action to check spelling

aaronraimist:
Works the same way as https://github.com/matrix-org/matrix-doc/pull/3591

> This uses https://github.com/crate-ci/typos so it will only find known misspellings in their dictionary. It won't know about or correct Matrix specific terminology but I think this should find most of the misspellings.

> When it finds a typo it looks like this https://github.com/aaronraimist/matrix-doc/runs/4621564484?check_suite_focus=true

> When there is a false positive, like a room ID, it can be added to the exclude list in https://github.com/matrix-org/matrix-doc/blob/1f41c6db6c943f19643dbeba31fb045d81da9728/.github/_typos.toml

Preview: https://pr1226--matrix-org-previews.netlify.app


gatsby/content/projects/sdks/simple-matrix-bot-lib.mdx
```diff
@@ -17,56 +17,47 @@ screenshot: /docs/projects/images/python.png
 maturity: Released
 ---
 
-(Version 2.4.1)
-
-Simple-Matrix-Bot-Lib is a Python bot library for the Matrix ecosystem built on [matrix-nio](https://github.com/poljar/matrix-nio).
+Simple-Matrix-Bot-Lib is a Python bot library for the Matrix ecosystem built on [matrix-nio](https://github.com/poljar/matrix-nio)
```
afranke, line 20: Changes to this file look like they have nothing to do with the topic of this PR. Did you really mean to include that?

.github/_typos.toml
```diff
@@ -0,0 +1,32 @@
+[files]
+extend-exclude = [
+	"*.js*", "*.json", "*.svg",
+	"/pre-generated", "/jekyll", # can be removed once https://github.com/matrix-org/matrix.org/pull/1126 is merged
```
afranke, line 4: This happened. :slightly_smiling_face:

gatsby/content/projects/sdks/simple-matrix-bot-lib.mdx
aaronraimist: 🤔 Good spot. I'm not sure what happened here. I have undone these changes.

.github/_typos.toml
aaronraimist: Thanks! Removed that line.

PR #1234, opened 2022-01-18
fix link in 2022-01-18-synapse-1.50-released.mdx

HarHarLinks:
@babolivier


babolivier (approved):
Gah, thanks for noticing and fixing!

Issue #1235, opened 2022-01-19
Add unstable spec warnings to the top of all specs at matrix.org/docs


turt2live:
https://matrix.org/docs/spec/client_server/r0.6.1 and r0.5.0 are observed ones. Should appear like r0.4.0: https://matrix.org/docs/spec/client_server/r0.4.0

PR #1239, opened 2022-01-26
Add "not latest" warning to all specs and remove latest.html

turt2live:
The latest.html redirects were handled back in Matrix 1.1's release.

Fixes https://github.com/matrix-org/matrix.org/issues/1235

Preview: https://pr1239--matrix-org-previews.netlify.app


turt2live:
I'm not fixing typos on this - suspect it's fallout from https://github.com/crate-ci/typos/issues/411

aaronraimist:
Those words are actually misspelled though. Typos just didn't find them due to that bug.

afranke:
LGTM.

@aaronraimist would you mind submitting a PR to fix those typos that fell through?

afranke (approved):

afranke (changes requested):
Actually `latest.json` is still being linked from `scripts/update_docs.sh`. Can you remove that line from the script?

afranke:
Make that two lines, the mkdir becomes moot as well.

afranke:
Can you please also add a rule for `latest.html` alongside [the existing “latest” rules](https://github.com/matrix-org/matrix-ansible-private/blob/main/roles/apache/templates/sites/matrix.org.conf.j2#L114)?

afranke:
Huh, the ones that exist should actually already cover this. Sorry, I was mislead by the 404 on the Netlify preview.

turt2live:
> Those words are actually misspelled though. Typos just didn't find them previously due to that bug.

They're valid, but strictly out of scope of this PR.

turt2live:
> Actually latest.json is still being linked from scripts/update_docs.sh. Can you remove that line from the script?

@afranke I don't think this is the same `latest.json`? I'm working in `unstyled_docs/spec` whereas the script uses `unstyled_docs/api` (which doesn't exist, but I assume it exists at build time).

afranke:
No, `unstyled_docs/api/` is where the link ends up. It **is** that same `latest.json` file that is linked *to*. In any case, just like the rest of this MR, this is about ancient references to the latest version of the spec that has become irrelevant with the new spec website. More specifically, this was for Rapidoc to use, but Rapidoc now points at the new website. It needs to go.

turt2live:
or can we just remove the whole 4 lines?

afranke:
The `mkdir` needs to be tweaked, and the `ln` line removed. I can do it, if you're ok with me pushing to your branch. We are not removing more as the rest is still needed for Rapidoc setup. It’s this:

```
diff --git a/scripts/update_docs.sh b/scripts/update_docs.sh
index c15b184f7..29157180a 100755
--- a/scripts/update_docs.sh
+++ b/scripts/update_docs.sh
@@ -25,8 +25,7 @@ cd "$(dirname "$(dirname "${SELF}")")"
 
 # Set up foundations for RapiDoc API playground
 rm -fr unstyled_docs/api/client-server
-mkdir -p unstyled_docs/api/client-server/json
-ln -s ../../../spec/client_server/latest.json unstyled_docs/api/client-server/json/api-docs.json
+mkdir -p unstyled_docs/api/client-server
 
 # copy the unstyled docs and add the styling
 rm -rf content/docs
```

turt2live:
yes, please do push to the branch :D 

(I still don't really understand the changes needed)

afranke (approved):

afranke:
We’re all good now. If that works for you, you can squash & merge.

PR #1241, opened 2022-01-27
♿️ Increase playground font size

afranke:
I use the available API in Rapidoc here to tweak the font for improved readability. The weight cannot be changed this way, but I reckon this does the trick.

@turt2live hope that works for you.

Preview: https://pr1241--matrix-org-previews.netlify.app

### DIFF
--- a/rapidoc/index.html
+++ b/rapidoc/index.html
@@ -3,12 +3,15 @@
   <head>
     <meta charset="utf-8"> <!-- Important: rapi-doc uses utf8 characters -->
     <script type="module" src="./rapidoc-9.1.2+9364198.min.js"></script>
+    <link rel="stylesheet" type="text/css" href="/OpenSans.css">
   </head>
   <body>
     <rapi-doc
       spec-url = "https://spec.matrix.org/latest/client-server-api/api.json"
       show-header="false"
       render-style="focused" 
+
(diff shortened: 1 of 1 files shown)


afranke:
https://mrin9.github.io/RapiDoc/examples/font-size.html#overview provides a handy way to compare the available options for `font-size`. I went with `largest`.

afranke:
[Before](https://matrix.org/docs/api/client-server/#get-/account/3pid) | [After](https://pr1241--matrix-org-previews.netlify.app/docs/api/client-server/#get-/account/3pid)

turt2live (commented):
This unfortunately doesn't help. It's honestly a weight issue, not a size issue.

This is what it looks like for me: 
image

Those letters are so thin the antialiasing doesn't know what to do with it (like on the `G`)

afranke:
The problem was that the proper weights weren’t loaded. Only the semibold and light weights of OpenSans were. I added the necessary stylesheet and the text is indeed more legible now. I am still bumping the size up a notch.

Please have another look at the before/after links and confirm that this works for you.

afranke:
This solves the issue and is a definitive improvement. I’ll go ahead and merge. Let us know if you still notice a problem with fonts.

PR #1246, opened 2022-02-02
Make other API available in the playground

afranke:
This adds a combo that allows selection of API amongst the ones we generate.

Depends on https://github.com/matrix-org/matrix-doc/pull/3684

Fixes https://github.com/matrix-org/matrix-doc/issues/1458

### DIFF
--- a/rapidoc/index.html
+++ b/rapidoc/index.html
@@ -4,9 +4,29 @@
     <meta charset="utf-8"> <!-- Important: rapi-doc uses utf8 characters -->
     <script type="module" src="./rapidoc-9.1.2+9364198.min.js"></script>
     <link rel="stylesheet" type="text/css" href="/OpenSans.css">
+    <script>
+      function getRapiDoc() {
+        return document.getElementById("rapidoc");
+      }
+      function switchSpec() {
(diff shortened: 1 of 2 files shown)


richvdh (approved):
seems generally sensible to me

afranke:
Something’s amiss.

* https://matrix.org/docs/api/ gives a forbidden
* The nav bar with icons at https://matrix.org/docs/develop still points to https://matrix.org/docs/api/client-server/ even though it should be fixed (see the diff for `[gatsby/src/components/Navigation.js](https://github.com/matrix-org/matrix.org/pull/1246/files#diff-f10b3fcf2d7b3d202f13b235c7375c1814c66c2c817b3826c6de33d598355719) line 38)
* There is still a Rapidoc behind that URL, without the combo box, like before.

Looks like it didn’t deploy? 17 hours after it’s been merged, doesn’t seem right…

richvdh:
hrm, thanks @afranke; will take a look

richvdh:
so it turns out the webhook listener was ooming. Hopefully that is now improved with https://github.com/matrix-org/gha-webhook-listener/pull/8.

PR #1249, opened 2022-02-02
Add registration bot

moan0s:
With the v1.2 and the now official registration API I want to introduce a bot that makes it much easier to obtain an manage tokens.

Preview: https://pr1249--matrix-org-previews.netlify.app


afranke (approved):
LGTM

PR #1251, opened 2022-02-03
⬆️ Upgrade lockfile

afranke:
This is the result of “just” doing an `npm install`, which complained that the lockfile version was old and needed to be upgraded, then proceeded to do said upgrade.

Preview: https://pr1251--matrix-org-previews.netlify.app


thibaultamartin:
Was there anything blocking this PR? Could this impact other projects than the matrix.org website?

richvdh:
I don't think so.

Looks like it just got forgotten about.

But it also looks like it's now outdated.

afranke:
> Was there anything blocking this PR?

This didn’t break anything for me, who ran the command, and it didn’t seem to break the CI, but it has potential to break stuff for anyone else who had a clone before, as this changes the `lockfileVersion` from `1` to `2`. When I pushed this, I asked @thibaultamartin to please pull it and see if it broke anything for him, as he was a perfect candidate for someone with an existing clone who might run into problems.

> Could this impact other projects than the matrix.org website?

I don’t see how this could affect anything else.

> Looks like it just got forgotten about.

Well it’s not as if much work was done on the website since Feb 3rd, apart from content.

> But it also looks like it's now outdated.

Yes, if someone has a bit of time to work on the website, a good approach would possibly be to just run `npm install` again and upgrade as prompted. Otherwise, if we can get the confirmation I asked for above, we could also merge this for now and do a less disruptive update after that.

PR #1273, opened 2022-02-22
Synapse 1.53

babolivier:
And it's a chunky one.

To be merged only once 1.53 is released.

Preview: https://pr1273--matrix-org-previews.netlify.app


gatsby/content/blog/2022/02/2022-02-22-synapse-1.53.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+## Time-based cache expiry now enabled by default
+
+To avoid being overly intensive on resources by making too many queries to the database, Synapse maintains several in-memory caches to store data it needs to use frequently. However, this comes with the inconvenience that, if Synapse needs to store too much data, these caches can become fairly big and occupy too much space in the host's memory.
+
+Historically, Synapse has been dealing with this issue by having set sizes for each cache, either hardcoded or set in the configuration, and evicting the oldest items when exceeding this size. But relying just on this solution alone comes with its own issues. For example, adding an entry to the cache that Synapse will only rarely use might cause it to evict other entries that it needs to use more often, thus forcing Synapse to query the database for them again.
```
richvdh, line 21:
```suggestion
Historically, Synapse has dealt with this issue by having set sizes for each cache, either hardcoded or set in the configuration, and evicting the oldest items when exceeding this size. But relying just on this solution alone comes with its own issues. For example, adding an entry to the cache that Synapse will only rarely use might cause it to evict other entries that it needs to use more often, thus forcing Synapse to query the database for them again.
```

richvdh: we haven't done anything to address the fact that a new cache entry might evict more useful entries, so the last sentence here is wrong.

richvdh (changes requested):
looks great in general!

gatsby/content/blog/2022/02/2022-02-22-synapse-1.53.mdx
babolivier: I've removed that sentence entirely since it's not needed and a bit confusing anyway.

richvdh (approved):

thibaultamartin:
Can I get a tweetable summary in the advocacy room when this is merged :) ?

PR #1292, opened 2022-03-21
Update types of bridging link

Aeyk:


gatsby/src/pages/bridges.js
```diff
@@ -97,7 +97,7 @@ const Bridges = ({ data }) => {
...
-          <p>For a thorough examination of the different methods of bridging, and a discussion of the terminology involved, check out <em><a href="https://matrix.org/blog/2017/03/11/how-do-i-bridge-thee-let-me-count-the-ways">How do I bridge thee? Let me count the ways...</a></em></p>
+          <p>For a thorough examination of the different methods of bridging, and a discussion of the terminology involved, check out <em><a href="https://matrix.org/docs/guides/types-of-bridging">How do I bridge thee? Let me count the ways...</a></em></p>
 
           <p><strong>Portal rooms</strong>: these control chunks of room aliases namespace. For example, <code>#freenode_<i>#channelname</i>:matrix.org</code> corresponds to <i>#channelname</i> on Freenode. In this way, Matrix users can transparently join IRC channels on Freenode. Portal rooms are typically managed by the remote network's side of the room.</p>
```
thibaultamartin, line 102: Thanks for this PR! Can you update the example from free node to match OFTC, as the freenode bridge was shut down (see https://matrix-org.github.io/matrix-appservice-irc/latest/bridged_networks.html)?

thibaultamartin, line 102: Thanks for this PR! Can you update the example from free node to match OFTC, as the freenode bridge was shut down (see https://matrix-org.github.io/matrix-appservice-irc/latest/bridged_networks.html)?

Aeyk: I replaced mention of freenode with libera.chat, as that's the spiritual successor of freenode, if you prefer it to be OFTC I can replace it.

gatsby/src/pages/bridges.js
```diff
@@ -97,11 +97,11 @@ const Bridges = ({ data }) => {
       <div className="mxgrid">
         <div className="mxgrid__item50">
           <p>An important idea in Matrix is <em>Interoperability</em>. This means that Matrix is open to exchanging data and messages with other platforms using an <a href="https://matrix.org/docs/spec">Open Standard</a>. We refer to the connection to other platforms as <em>bridging</em>.</p>
-          <p>For a thorough examination of the different methods of bridging, and a discussion of the terminology involved, check out <em><a href="https://matrix.org/blog/2017/03/11/how-do-i-bridge-thee-let-me-count-the-ways">How do I bridge thee? Let me count the ways...</a></em></p>
+          <p>For a thorough examination of the different methods of bridging, and a discussion of the terminology involved, check out <em><a href="https://matrix.org/docs/guides/types-of-bridging">How do I bridge thee? Let me count the ways...</a></em></p>
```
afranke, line 100: The link text should be updated as well to match the title of the new page.

gatsby/src/pages/bridges.js
thibaultamartin: I believe keeping OFTC makes sense, as it has a prefix (the `#OFTC_` par of the room alias) Libera Chat doesn't have. I would appreciate if you could make the change, but can do it if you got busy in the meantime :)

Issue #1109, opened 2021-11-04
Streamline the redirect situation


afranke:
We have too many different ways to set up redirects (so that legacy URLs still work but land you on a newer page):

* in the Apache config (in another repo managed by ops)
* in a handful of leftover jekyll pages (those won’t be a problem for long)
* in ~50 gatsby pages (`$ git grep 'http-equiv.*refresh' -- "./gatsby/*"`)
* and worst of all we have [a script that generates another 50](https://github.com/matrix-org/matrix.org/blob/master/scripts/generate-redirects.js).

We don’t need to settle on a single one (I reckon some belong in Apache config and some don’t), but we probably want to merge the last three methods.

afranke:
The script seems to be what created most of the gatsby ones actually. Nothing in the repo calls it and it must have been called manually, as a one-off. I’m a bit confused now.

richvdh:
> We have too many different ways to set up redirects (so that legacy URLs still work but land you on a newer page):
> 
>     * in the Apache config (in another repo managed by ops)

Internal users can see these at https://github.com/matrix-org/matrix-ansible-private/blob/main/roles/apache/templates/sites/matrix.org.conf.j2#L78

>     * in a handful of leftover jekyll pages (those won’t be a problem for long)
> 
>     * in ~50 gatsby pages (`$ git grep 'http-equiv.*refresh' -- "./gatsby/*"`)
>     * and worst of all we have [a script that generates another 50](https://github.com/matrix-org/matrix.org/blob/master/scripts/generate-redirects.js).

per the above, these are the same thing. It sounds like the script has served its purpose and can be ignored or removed.

Sounds like the jekyll ones are going away too.

So that leaves us with two places:
 * Apache
 * Gatsby

> I reckon some belong in Apache config and some don’t

yes, this. Each of Apache/Gatsby has its own attraction:

 * Importantly, the Apache redirects are "server-side" (they return a 30x code rather than relying on the HTML engine) - which means they also work for things like the federation tester which serves a lump of json. They are also generally lighter weight.
 * The Gatsby ones allow the redirects to be versioned alongside the content and don't require management from the ops team.

In short, I think there is a place for both. I'm sure there are some that could do with being moved from one to the other, though.

richvdh:
I don't think there's anything to do here.

afranke:
[...]

PR #1293, opened 2022-03-22
Remove redundant `.htaccess` file

richvdh:
We need to wait for https://github.com/matrix-org/matrix-ansible-private/pull/4658 to be deployed, but once it has been, we should remove the `.htaccess` file.

Preview: https://pr1293--matrix-org-previews.netlify.app


afranke:
See https://github.com/matrix-org/matrix.org/issues/1109#issuecomment-1076212880

PR #1301, opened 2022-03-29
Fix unable to select text on animation bug (#1053)

### LINKED ISSUE #1053: Unable to select text in `How does it work?` animation section
In the text panels explaining each part of the `How does it work?` animation, each text panel is positioned/stacked on top of each other and is only hidden by `opacity: 0;`. So you're unable to select any text on any of the panels except the last one.

https://matrix.org/

## Potential solutions

We should update it to actually hide the panels from the document flow `display: none;` or at least have `pointer-events: none` for hidden panels.

tawandamoyo:
Fixes #1053 , used the suggestion from @MadLittleMods

### DIFF
--- a/gatsby/static/css/how-it-works.css
+++ b/gatsby/static/css/how-it-works.css
@@ -13,8 +13,9 @@
 .legend {
     position: absolute;
     padding-top: 20px;
-    opacity: 0;
     max-width: 570px;
+    opacity: 0;
+    pointer-events: none;
 }
 
 .legend pre {
--- a/gatsby/static/js/how-it-works.js
+++ b/gatsby/static/js/how-it-works.js
@@ -490,8 +490,8 @@ function nextStage() {
     d3.select("#legend" + (stageIndex + 1)).style("opacity", 1e-6)
         .transition()
         .duration(dissolveTime)
-        .style("opacity", 1);
-
+        .style({ "opacity": 1, "pointer-events": "auto" });
+        
     if (pendingStages.length > 0) {
         var c = pendingStages.length;
         console.log("starting c=" + c);


gatsby/static/js/how-it-works.js
```diff
@@ -483,14 +483,14 @@ function nextStage() {
 
     var dissolveTime = 500;
 
-    d3.select("#legend" + stageIndex).style("opacity", 1)
+    d3.select("#legend" + stageIndex).style("display", "block")
         .transition()
         .duration(dissolveTime)
-        .style("opacity", 1e-6);
-    d3.select("#legend" + (stageIndex + 1)).style("opacity", 1e-6)
+        .style("display", "none");
```
MadLittleMods, line 489: The opacity transition is lost here. We can still have the best of both worlds though.

Or if you can't figure it out, we can always use `pointer-events: none;` on hidden panels so you can click through them.

tawandamoyo: HI @MadLittleMods , thanks for the feedback. I have used the second option of `pointer-events: none`. If you could kindly review, thanks.

gatsby/static/css/how-it-works.css
```diff
@@ -13,8 +13,9 @@
 .legend {
     position: absolute;
     padding-top: 20px;
-    opacity: 0;
     max-width: 570px;
+    opacity: 0;
+    pointer-events: none;
```
MadLittleMods, line 18: Looks good to me 👍

https://user-images.githubusercontent.com/558581/167490575-a9279ffb-6bec-412b-a494-414c094344df.mp4

MadLittleMods (changes requested):
Remove changes to `gatsby/package-lock.json`

gatsby/static/css/how-it-works.css
```diff
@@ -13,8 +13,9 @@
 .legend {
```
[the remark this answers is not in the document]
tawandamoyo: Oh thanks, done.

MadLittleMods:
Thanks for the contribution and fix @tawandamoyo! Now the demo is even more slick 🦦

tawandamoyo:
Thanks for the reviews and suggestions @MadLittleMods , I learnt a lot. And happy to contribute in any way to this revolutionary project!

PR #1304, opened 2022-03-30
Update Thunderbird metadata and add a Screenshot

freaktechnik:
Prepare the Thunderbird metadata for what will likely be released later on this year and is already available in Daily and Beta.

CC @clokep 

Preview: https://pr1304--matrix-org-previews.netlify.app


gatsby/content/projects/clients/thunderbird.mdx
```diff
@@ -14,6 +15,7 @@ platforms:
     - Mac
     - Windows
 repo: https://hg.mozilla.org/comm-central
+screenshot: /docs/projects/images/thunderbird.png
 home: https://www.thunderbird.net/
 room: "#maildev:mozilla.org"
```
freaktechnik, line 20: It might be more appropriate to point to our space at `#thunderbird-community:mozilla.org`? Or at the very least at `#thunderbird:mozilla.org`?

clokep: I think this info is meant for users, so I think either of those makes more sense. 👍

thibaultamartin: I would favour a room to welcome people over a Space, as some Matrix clients don't support them yet. @freaktechnik do you mind updating the room to `#thunderbird:mozilla.org` before I merge this PR :) ?

freaktechnik:
I think this broke the site, because it redirects from https://matrix.org/docs/projects/client/thunderbird to https://matrix.org/docs/projects/client/mozilla-thunderbird, which now no longer exists. I think the redirect is generated in https://github.com/matrix-org/matrix.org/blob/master/scripts/generate-redirects.js

PR #1306, opened 2022-04-03
Remove confusing "{project} on matrix.org" text

tulir:
It looks like it was originally added as a part of the new SDKs page, but later leaked into bridges when that page was updated. For bridges, it's very confusing as there's no link and it can be interpreted as the bridge being hosted on matrix.org. For SDKs, it just links to a page that has the exact same content as the current page, so I don't think it's necessary at all (e.g. on https://matrix.org/sdks/#mautrix-go it links to https://matrix.org/docs/projects/sdk/mautrix-go)

image
image

Preview: https://pr1306--matrix-org-previews.netlify.app


thibaultamartin:
Thanks, it did look confusing indeed

Issue #1311, opened 2022-04-12
Rename libpurple to XMPP/Jabber


JokerGermany:
**Is your feature request related to a problem? Please describe.**
The XMPP Bridge is hard to find under libpurple for endusers.

**Describe the solution you'd like**
Would be nice to rename libpurple in XMPP/Jabber
https://matrix.org/bridges/#libpurple

afranke:
It is not limited to XMPP though.

JokerGermany:
push

jugendhacker:
> It is not limited to XMPP though.

It's not, but currently it's the main purpose and people actually searching for a XMPP bridge will have a hard time finding it hidden behind libpurple.

JokerGermany:
push

JokerGermany:
> @JokerGermany feel free to open a PR
If i would fully understand how this code works, i would have done this before... but i will try.

I would change this in this file libpurple to xmpp
https://raw.githubusercontent.com/matrix-org/matrix.org/5c828263bf38a9e99cd370fc9a37ae1217b592fb/gatsby/content/projects/bridges/matrix-bifrost.mdx

and in this file i would change
https://github.com/matrix-org/matrix.org/blob/1606ade277e2b7e30a91eb43a71a005d7b960e05/gatsby/src/pages/bridges.js
`["libpurple", "libpurple"],`
to
`["Jabber/XMPPP", "xmpp"],`
Is this correct?

MadLittleMods:
@JokerGermany It would be best if you checked out the project locally and just tried it. There is no project owner here who probably knows for sure and we would have to validate it just like you. It sounds like you're on the right track. Instructions here for how to build the project locally: https://github.com/matrix-org/matrix.org#how-to-build

richvdh:
> Rename libpurple in XMPP/Jabber

I think you mean "Rename libpurple *to* XMPP/Jabber" ?

JokerGermany:
Looks like it will be changed with the new version of the Homepage:
https://matrix.turbo.fish/ecosystem/bridges/

juniorlegion123:
can you assign this issue to me. I am coonfident that i will be able to solve this issue.

thibaultamartin:
@juniorlegion123 this is already fixed in the upcoming matrix.org website

PR #1340, opened 2022-05-20
rename libpurple in XMPP

JokerGermany:
solves #1311


afranke:
@Half-Shot should probably be aware of this.

PR #1343, opened 2022-05-25
updated project info for matrix-commander

8go:
Please update our project info on your webpage. Thank you. 

Kind regards!  :heart: and thanks!

Preview: https://pr1343--matrix-org-previews.netlify.app


thibaultamartin:
Hej 8go, thanks for updating the project information. I have mixed feelings about having a good chunk of the documentation of the client on this page. I think it'd be better to route people to the project's information so they have the latest freshest information, but given there was already some documentation previously I'm going to merge it. We may revisit that together eventually if we bring changes to matrix.org :)

PR #1346, opened 2022-06-01
Switch to using matrix-org/netlify-pr-preview action

t3chguy:
Preview: https://pr1346--matrix-org-previews.netlify.app


thibaultamartin:
@t3chguy I think it's safe to close that PR?

PR #1347, opened 2022-06-01
Use matrix-org/pr-details-action@v1.1 for Netlify deploys

t3chguy:
Preview: https://pr1347--matrix-org-previews.netlify.app

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -17,22 +17,12 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request'
     steps:
-      - name: "dump context data"
-        run: |
-          jq . < $GITHUB_EVENT_PATH
-
       - name: "🔍 Read PR number"
-        id: readctx
-        # we need to find the PR number that corresponds to the branch, which we do by
-        # searching the GH API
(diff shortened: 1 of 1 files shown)


.github/workflows/netlify.yml
```diff
@@ -53,7 +43,7 @@ jobs:
           deploy-message: "Deploy from GitHub Actions"
           enable-pull-request-comment: false
           enable-commit-comment: false
-          alias: pr${{ steps.readctx.outputs.prnumber }}
+          alias: pr${{ steps.prdetails.outputs.pr_id }}
```
richvdh, line 46: this is maybe more of a comment on the action, but this doesn't seem to be defined in the action's documentation. I'd really rather the action be properly documented before we switch things over to it.

richvdh (changes requested):

.github/workflows/netlify.yml
t3chguy: Good call, I tend to rely on documenting within actions.yml but duplicating to README helps - https://github.com/matrix-org/pr-details-action/blob/main/README.md#inputs

thibaultamartin:
@t3chguy same here, this one can safely be closed I presume?

thibaultamartin:
Err, got a bit trigger happy here, sorry. Please reopen if needed

PR #1354, opened 2022-06-05
Fix 'previous' link in categories on pages ≥ 3

reivilibre:
It goes to /posts/ rather than the actual category right now.

See https://matrix.org/blog/category/this-week-in-matrix/33 for example.

Preview: https://pr1354--matrix-org-previews.netlify.app


thibaultamartin:
Good catch, thanks!

PR #1367, opened 2022-06-27
update fluffychat features

Raatty:
hopefully this is the right way to update https://matrix.org/clients-matrix/ there was some missing features also got rid of a couple of platforms that aren't supported anymore


thibaultamartin:
Thanks for helping us keeping it up to date :)

PR #1391, opened 2022-07-26
Cinny is available on desktop platforms

1bcb:
Preview: https://pr1391--matrix-org-previews.netlify.app


thibaultamartin:
It might be a misunderstanding here: desktop clients support means you can install it as a desktop app on your desktop, in a way supported by upstream.

@ajbura is Cinny actually available as desktop clients (.exe on Windows, .dmg on macOS and e.g. Flatpak on Linux)?

thibaultamartin:
Thanks for helping us keeping it up to date :tada:

PR #1399, opened 2022-08-12
Publish TWIM 20220812

anoadragon453:


anoadragon453:
The other spell checking error (`tha'`) is a false positive.

anoadragon453:
In which I remember that this will now cause CI to fail for all eternity :facepalm:

aaronraimist:
@anoadragon453 https://github.com/matrix-org/matrix.org/blob/master/.github/_typos.toml#L5-L10

anoadragon453:
@aaronraimist [Fixed](https://github.com/matrix-org/matrix.org/pull/1400), thanks!

PR #1403, opened 2022-08-16
Fix thirdroom link Matrix Summer Special

uranuspucksaxophone:
Preview: https://pr1403--matrix-org-previews.netlify.app


babolivier (approved):
Thanks for fixing it! :)

PR #1411, opened 2022-08-23
Fix Synapse support room reference in FAQ

ilmari:
The sentence was missing several words, and using a non-published
address for the room.  Make it a complete sentence and change the link
to use the canonical address.


babolivier (approved):
Thanks for PR!

Issue #1416, opened 2022-08-24
Upgrade to gatsby v4 - This is not an issue but more a hint on "how far did I get"


MTRNord:
I tried to migrate this. I ended up not finishing it after 2 days of work as it seems that multiple MDX files in a single page (FAQ page) is not any more supported, and queries in templates are not executed anymore which makes all posts fail. Also, `MDXRenderer` was removed. Making `allMdx` not usable for rendering pages like used in here. After talking to the gatsby people it seems like this is and never was officially supported anyway. So I deem this migration impossible.

Either way, here is how far I got in case this will be of any use ever: https://github.com/matrix-org/matrix.org/compare/master...MTRNord:matrix.org:MTRNord/upgrade-gatsby

The commits do explain some stuff on what happens there.

MTRNord:
(Feel free to close this. This is just to record this in case it ever gets useful for someone. So just that it does not get lost)

MadLittleMods:
@MTRNord Create a pull request so it's easier to look at and then we can close that

MTRNord:
> @MTRNord Create a pull request so it's easier to look at, reference, and comment and then we can close that.

Sure I can do that :)

MTRNord:
@MadLittleMods Done at https://github.com/matrix-org/matrix.org/pull/1417

MadLittleMods:
Thanks @MTRNord! I think the PR will track this better 👍

PR #1417, opened 2022-08-24
Gatsby V4 - Do not merge as this is not (yet?) working

MTRNord:
This is my journey on trying to migrate this.

Mostly the thing is that the current structure is not migratable as is, as some things have entirely been removed by gatsby or mdx.

I ended up not finishing it after 2 days of work as it seems that multiple MDX files in a single page (FAQ page) is not any more supported, and queries in templates are not executed anymore which makes all posts fail. Also, `MDXRenderer` was removed. Making `allMdx` not usable for rendering pages like used in here. After talking to the gatsby people it seems like this is and never was officially supported anyway. So I deem this migration impossible.

(originally shared in #1416 which has some details)


MTRNord:
Oh and yes this will build and also show the front page and a lot of the guides. Just the blog for example doesnt work at all.

MTRNord:
Turns out at least the blog posts are possible to be fixed. FAQ is then left as the major issue. This may end up actually being not as impossible as I assumed

MTRNord:
Seems like currently the code block design is broken, MDX tables are broken and faq is broken

PR #1435, opened 2022-09-07
Add page for matrix-rss-bridge

imbev:
Preview: https://pr1435--matrix-org-previews.netlify.app


gatsby/content/projects/bridges/matrix-rss-bridge.mdx
```diff
@@ -0,0 +1,53 @@
+---
+layout: project
+title: matrix-rss-bridge
+categories:
+ - bot
+ - bridge
```
thibaultamartin, line 6:
```suggestion
```

This seem to be more of a bot than a bridge

gatsby/content/projects/bridges/matrix-rss-bridge.mdx
```diff
@@ -0,0 +1,53 @@
...
+description: matrix-rss-bridge is a bridge for reading RSS feeds in Matrix rooms.
```
thibaultamartin, line 7:
```suggestion
description: matrix-rss-bridge is a bot for reading RSS feeds in Matrix rooms.
```

gatsby/content/projects/bridges/matrix-rss-bridge.mdx
imbev: It is both a bot and a bridge, similar to feedbot which is categorized as both.

thibaultamartin: Ah good catch, this is a mistake: at first sight feedbot is not a bridge either. The major difference between a bridge and a bot is that bridges a bi-directional.

I might have missed it, but I don't think this is the case for this project, is it?

thibaultamartin: I suggest to remove the bridge section for now, and to add it in the future if you add features to make it bi-directional. Are you fine with me applying the suggestions and merging the PR?

imbev: Yes

PR #1437, opened 2022-09-09
Make the scroll targets adjust for the navigation bar

### LINKED ISSUE #1152: Clicking a section header link brings too far down
**Describe the bug**
When clicking a section link in an article on matrix.org/blog, the title of the section is hidden because the window scrolls too far to the bottom.

**To Reproduce**
Example:
1. Click on https://matrix.org/blog/posts#the-process-of-improving-coverage
2. "The process of improving coverage" is too far up
 
**Expected behavior**
It would be expected to show the title of the section.

**Additional context**

callahad suggested:

> Yep. Setting scroll-padding-top: 5em; on html in the blog works well enough for me. I'll leave it as an exercise for the reader to figure out that best actual selector and value for this.
> 
> Whole article on how to make this happen at https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/

MTRNord:
This should fix #1152

Apparently this css feature is rather (if not brand) new, but it is supported by all major browsers (including safari!): https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin#browser_compatibility

Preview: https://pr1437--matrix-org-previews.netlify.app

### DIFF
--- a/gatsby/static/css/matrix-org.webflow.css
+++ b/gatsby/static/css/matrix-org.webflow.css
@@ -46,6 +46,15 @@ h6 {
   font-weight: 500;
 }
 
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+    scroll-margin-top: 74px;
+}
+
 a {
   color: #0098d4;
   text-decoration: none;
--- a/gatsby/static/css/matrix-org.webflow.css
+++ b/gatsby/static/css/matrix-org.webflow.css
@@ -63,9 +63,9 @@ h6 {
   h4,
   h5,
   h6 {
-    /* Offsets the anchor scroll position by the navbar height + 10px buffer.
-    The mobile navbar is 4px smaller in height than the regular navbar. The breakpoint the navbar changes at is 991px */
-    scroll-margin-top: 70px;
+    /* The mobile version of the navbar correctly pushes the elements down.
+       Therefor we dont need to have the offset. We however need to unset this as otherwise the 74px is applied. */
+    scroll-margin-top: inherit;
   }
 }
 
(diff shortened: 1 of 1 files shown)


MTRNord:
Example for a scroll target affected: https://pr1437--matrix-org-previews.netlify.app/blog/2022/09/09/this-week-in-matrix-2022-09-09#dept-of-servers-

gatsby/static/css/matrix-org.webflow.css
```diff
@@ -46,6 +46,15 @@ h6 {
   font-weight: 500;
 }
 
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+    scroll-margin-top: 74px;
```
Costava, line 55: The `.mxnavbar` class does not appear to have a fixed height.
When I resize the window, I see that the navbar is either `60px` or `64px` tall.
Is `74px` an arbitrary size that will hopefully always be larger than the navbar?

MTRNord: It was mostly a size that looks fine on my monitor 😅

But it's easy to use some media breakpoints to fix that scaling issue

Costava: As it is currently, I think it would be valuable to add a comment noting how the value was chosen.
e.g. `/* An arbitrary size that has a ~10px buffer above the expected size of the navbar */`
But I agree that `@media` can be a better solution.

MTRNord: Added the `@media` and the comment to it now :)

thibaultamartin:
I hadn't realised this PR would fix #1152 when jotting a quick fix. I think this PR can be closed?

PR #1442, opened 2022-09-14
Remove noscript banner

progval:
The website works perfectly fine without Javascript


MTRNord:
Doesn't the "how works matrix" (Alice, Bob and Charlie thingy) animation use d3js?

progval:
Oh, you're right; I never noticed it. I could move the `<noscript>` there

PR #1469, opened 2022-09-29
Add security.txt

### LINKED ISSUE #1468: Consider adding a security.txt
While matrix.org has a security page, https://securitytxt.org/ seems to become a standard and helps to have the key points in a structured format which people can easily access and integrate in whatever tools people may be using.

It would be great if it could be added to the page. :) Google and others use it, and it is recommended by many people to have it :)

MTRNord:
Fixes #1468

This adds a https://securitytxt.org/ file. This seemingly starts to get a common thing for security research and seems to be a nice to have along the security disclosure page.

Preview: https://pr1469--matrix-org-previews.netlify.app

### DIFF
--- a/content/.well-known/security.txt
+++ b/content/.well-known/security.txt
@@ -1,0 +1,7 @@
+Contact: mailto:security@matrix.org
+Expires: 2022-01-30T23:00:00.000Z
+Encryption: https://matrix.org/.well-known/pgp-key.txt
+Acknowledgments: https://matrix.org/hall-of-fame
+Preferred-Languages: en
+Canonical: https://matrix.org/.well-known/security.txt
+Policy: https://matrix.org/security-disclosure-policy/
--- a/content/.well-known/security.txt
+++ b/content/.well-known/security.txt
@@ -1,3 +1,6 @@
+-----BEGIN PGP SIGNED MESSAGE-----
+Hash: SHA256
+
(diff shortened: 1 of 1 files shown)


content/.well-known/security.txt
```diff
@@ -0,0 +1,7 @@
+Contact: mailto:security@matrix.org
+Expires: 2022-01-30T23:00:00.000Z
```
aaronraimist, line 2:
```suggestion
Expires: 2024-01-30T23:00:00.000Z
```

I'm assuming this shouldn't already be expired. Could do 2023 but there are only a few months left until 2023.

MTRNord: Oops it should have been 2023 mot 2022 :) the rfc says its recommended to be less than 1 year to ensure it not getting stale. Should I go with 2024 anyway? :)

aaronraimist: None of the information in here has changed since 2019 so imo there isn't much chance of it getting stale.

MTRNord: ok :) Then I just apply your suggestion 👍

dkasak:
Looks good, apart from concerns [outlined here](https://github.com/matrix-org/matrix.org/issues/1468#issuecomment-1265360060). I'm adding a signature from security@matrix.org and then rolling it out.

PR #1474, opened 2022-10-01
Fix typos in the TWIM announcement for the matrix-widget-api

dhenneke:
Just some minor fixes to improve the news.

Preview: https://pr1474--matrix-org-previews.netlify.app


dhenneke:
@thibaultamartin could you have a quick look if we can update the announcement?

PR #1477, opened 2022-10-03
Add NewsArticle schema data to the post template

MTRNord:
(Note this is against zola not against gatsby)

This implements https://developers.google.com/search/docs/appearance/structured-data/article for the blog articles.

### DIFF
--- a/templates/post.html
+++ b/templates/post.html
@@ -12,11 +12,15 @@
     "headline": "{{ page.title }}",
     "datePublished": "{{ page.date | date(format="%Y-%m-%dT%H:%M+00:00") }}",
     "dateModified": "{{ page.updated | date(format="%Y-%m-%dT%H:%M+00:00") }}",
-    "author": [{
+    "author": [
+    {% for author in page.taxonomies.authors %}
+    {
         "@type": "Person",
-        "name": "{{ page.taxonomies.authors | default (value=["unknown author"]) | first }}",
-        "url": "{{ config.base_url | safe }}/authors/{{ page.taxonomies.authors | default (value=['unknown author']) | first | slugify }}"
-    }]
+        "name": "{{ author }}",
+        "url": "{{ config.base_url | safe }}/authors/{{ author | slugify }}"
+    }{% if not loop.last %},{% endif %}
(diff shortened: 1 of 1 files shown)


templates/post.html
```diff
@@ -1,5 +1,25 @@
...
+{
+    "@context": "https://schema.org",
+    "@type": "NewsArticle",
+    "headline": "{{ page.title }}",
+    "datePublished": "{{ page.date | date(format="%Y-%m-%dT%H:%M+00:00") }}",
+    "dateModified": "{{ page.updated | date(format="%Y-%m-%dT%H:%M+00:00") }}",
+    "author": [{
+        "@type": "Person",
+        "name": "{{ page.taxonomies.authors | default (value=["unknown author"]) | first }}",
+        "url": "{{ config.base_url | safe }}/authors/{{ page.taxonomies.authors | default (value=['unknown author']) | first | slugify }}"
```
jplatte, line 18: Why `| first` instead of using all authors from the taxonomy to build the `author` array? Also I don't think taxonomies are usually in plural, i.e. it should probably be named just `author` as well.

MTRNord: Hm, the first indeed doesnt really make sense here. It's simply because we used it as a singular thing in the template.

The reason it is named plural is that it made (to me) more sense to have an array of "authors" in the frontmatter opposed to having an array of "author" in the frontmatter.

MTRNord: For example here the plural made to me more sense than the singular form https://github.com/matrix-org/matrix.org/blob/1ed7ec109afb289ca37e4cd6c2f9758140312fd7/content/blog/2022/09/2022-09-02-synapse-1.66.md?plain=1#L8-L9

jplatte: I think since in Zola the name of the taxonomy is linked to how you set it in the frontmatter, it still makes most sense for it to be singular. But ultimately I don't care that much.

MTRNord: (Just for record, the decision was made to first make here the authors generate the full objects for multiple authors and then change `authors` to `author` in another PR)

PR #1501, opened 2022-10-12
Template & pagination improvements

jplatte:


templates/includes/header.html
```diff
@@ -1,14 +1,5 @@
...
-{% elif section %}
-{% set path = section.path %}
-{% elif taxonomy %}
-{% set path = current_path %}
-{% else %}
-{% set path = "" %}
-{% endif %}
 <header class="site-header">
-    <a href="/" class="brand{% if path == '/' %} current{% endif %}">
+    <a href="/" class="brand {{ macros::current_class(if_path_starts_with='/') }}">
```
thibaultamartin, line 2: The CSS only applies a different text colour when there's a `current` class on a link. Do we really need it here, given the link only consists of the Matrix logo?

jplatte: I think I just copied what was being done before. I agree it's not actually necessary.

thibaultamartin (approved):

PR #1510, opened 2022-10-18
Cancel concurrent GHA builds

richvdh:
This is a thing we do on Synapse, and it makes things build faster by not tying up GHA workers with builds we don't care about.

Preview: https://pr1510--matrix-org-previews.netlify.app

### DIFF
--- a/.github/workflows/build-matrix.org.yml
+++ b/.github/workflows/build-matrix.org.yml
@@ -6,6 +6,11 @@ on:
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


thibaultamartin:
Will the default branch always have preference? Or can a PR cancel a build triggered from merging something into `master`?

richvdh:
oh, sorry, I explained this badly. Only builds on the same branch will be cancelled. So each PR is independent, as is `master`.

thibaultamartin (approved):
Lgtm then :)

PR #1511, opened 2022-10-19
Import 2014 blog posts from Gatsby

jplatte:


richvdh:
is there any way to do this with a `git mv`, rather than a copy?

when https://github.com/matrix-org/matrix.org/matrix-spec was migrated to a new system, all the files got copied, which makes it a nightmare if you ever want to follow the history of a file (because `git log` stops at the copy).

Mind you, that's probably way less important for a bunch of static blog posts than for the spec, so don't worry if it's non-trivial.

jplatte:
We intentionally started from a clean history. I think git-subtree or git-filter-repo might allow us to pull in the history of just the blog posts, but it's not going to be easy.

thibaultamartin:
I think the cost of preserving the git history outweighs the benefits of moving files like this :)
Additionally and if it's really needed, the information of who had created the original post is not lost, since it's still on the `master` branch

jplatte:
Cherry-picked the commit for 2022 `_index.md` files onto the zola branch and squashed the remaining two commits.

thibaultamartin (approved):
Built locally: everything seems to build and almost render correctly.

As agreed separately we're first importing all the posts in a way that doesn't make zola scream and we will clean up the html and oddities in the posts afterwards (see #1520)

PR #1519, opened 2022-10-21
Import 2022 blog posts from Gatsby

jplatte:
~~Doesn't build right now, need to fix..~~


thibaultamartin (approved):
Builds and looks fine, I added an entry in #1520 since the YouTube embedded player looks a little weird (it's square-ish with two fat bands above and below the video)

PR #1527, opened 2022-10-25
Docs for Developers

thibaultamartin:
Left to do:

- [ ] Table of Content?
- [x] Elements of Matrix
    - [x] Homeserver
    - [x] Client
    - [x] AppService (bots and bridges)
    - [x] The Spec defining interactions between all those

Out of scope:
- Tutorials (they will never be updated)
- Project specific docs (they will never be updated and could mislead people)

Will be split in another PR:
- An introduction for the page, a general overview of how Matrix works: there are several clients connected to different homeservers, federated together, and they send events.

### DIFF
--- a/content/docs/developers/_index.md
+++ b/content/docs/developers/_index.md
@@ -13,20 +13,20 @@ last_updated = "2022-10-18T16:00:00Z"
 
 A homeserver is a piece of software hosting accounts of Matrix users. It is
 bound to a single domain that cannot change over time. The accounts on a server
-have an identifier made of a local part, and a server part. A typical identifier
-would be
+have an identifier made of a local part (the user id), and a server part, which
+is the (vanity) domain of the homeserver. A typical identifier would be
 
 ```txt
 @username:example.com
 ```
 
-Users on a server can send `events` into `rooms`. An event is a particular json
+Users on a server can send *events* into *rooms*. An event is a particular json
 object, describing what a user is trying to do (join a room, send a message,
 update a specific value…). In the case of instant messaging, rooms are very
 similar to Slack, Discord or IRC channels. Most of the events in such rooms are
 about sending messages. Rooms have a unique technical identifier, and zero or
-several human-readable aliases or addresses. Aliases are made of a room name,
-and a server part. A typical room alias would be:
+more human-readable aliases. Aliases are made of a room name, and a server part,
+and are sometimes referred to as "addresses". A typical room alias would be:
 
 ```
 #mountain-bike:example.com
@@ -43,12 +43,13 @@ out to `example.com` to get a copy of the room. `outdoors.com` and `example.com`
 then stay in touch to synchronise their copy of the room.
 
 Whenever the homeserver receives new events, it's in charge of parsing them,
-perform checks on the event, and take action accordingly (e.g. add a specific 
-user to a specific room when it receives a room join event). The expected
-behaviour of homeservers is described fully in the [Matrix Specification](https://spec.matrix.org).
+perform checks on the event, and take action accordingly (e.g. sending messages
+from users on the homeserver to other participating homeservers, or distributing
+messages from other participating homeservers to users). The expected behaviour
+of homeservers is described fully in the [Matrix Specification](https://spec.matrix.org).
 
 You can find a list of existing homeserver implementations in the
-[Ecosystem > Servers](/ecosystem/servers) section of this website. Some of them
+[Ecosystem > Servers](/ecosystem/servers) section of this website. Most of them
 are open source, so you can explore how they work. Please also refer to their
 documentation if you want to deploy them either for testing or in production.
 
@@ -58,7 +59,7 @@ Homeservers communicate between each other with the Server-Server / Federation
 API, but they also communicate with clients in a standard way: the Client-Server
 API.
 
-Clients are pieces of software who can use a Matrix account to send and receive
+Clients are pieces of software that can use a Matrix account to send and receive
 events from a specific homeserver. The clients themselves only ever talk to the
 homeserver of the account they're using. If a client uses the
 `@alice:example.com` account, they will only talk to `example.com`.
--- a/content/docs/developers/_index.md
+++ b/content/docs/developers/_index.md
@@ -1,0 +1,8 @@
++++
+title = "Developers"
+weight = 300
+[extra]
+emoji = "⚒️"
+tile = "I want to develop a client, bridge or bot"
+last_updated = "2022-10-18T16:00:00Z"
++++
--- a/content/docs/developers/_index.md
+++ b/content/docs/developers/_index.md
@@ -13,7 +13,7 @@ last_updated = "2022-10-18T16:00:00Z"
 
 A homeserver is a piece of software hosting accounts of Matrix users. It is
 bound to a single domain that cannot change over time. The accounts on a server
-have an identifier made of a local part (the user id), and a server part, which
+have an identifier made of a local part (the username), and a server part, which
 is the (vanity) domain of the homeserver. A typical identifier would be
 
 ```txt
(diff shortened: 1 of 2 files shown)


content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+A homeserver is a piece of software hosting accounts of Matrix users. It is
+bound to a single domain that cannot change over time. The accounts on a server
+have an identifier made of a local part, and a server part. A typical identifier
+would be
+
+```txt
+@username:example.com
+```
+
+Users on a server can send `events` into `rooms`. An event is a particular json
```
jplatte, line 23:
```suggestion
Users on a server can send *events* into *rooms*. An event is a particular json
```

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+object, describing what a user is trying to do (join a room, send a message,
+update a specific value…). In the case of instant messaging, rooms are very
+similar to Slack, Discord or IRC channels. Most of the events in such rooms are
+about sending messages. Rooms have a unique technical identifier, and zero or
```
jplatte, line 27:
```suggestion
messages. Rooms have a unique technical identifier, and zero or
```

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+several human-readable aliases or addresses. Aliases are made of a room name,
```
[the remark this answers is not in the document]
jplatte:
```suggestion
more human-readable aliases. Aliases are made of a room name,
```

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+that room.
+
+For example, if `@alice:outdoors.com` is the first user from `outdoors.com` to
+try to join `#mountain-bike:example.com`, then her homeserver is going to reach
+out to `example.com` to get a copy of the room. `outdoors.com` and `example.com`
+then stay in touch to synchronise their copy of the room.
+
+Whenever the homeserver receives new events, it's in charge of parsing them,
+perform checks on the event, and take action accordingly (e.g. add a specific 
+user to a specific room when it receives a room join event). The expected
```
jplatte, line 47: Maybe it's just me, but what does "add a specific user to a specific room" mean if not sending a join event? Maybe we could just write something like "e.g. sending messages from users on the homeserver to other participating homeservers, or distributing messages from other participating homeservers to users"?

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+behaviour of homeservers is described fully in the [Matrix Specification](https://spec.matrix.org).
+
+You can find a list of existing homeserver implementations in the
+[Ecosystem > Servers](/ecosystem/servers) section of this website. Some of them
```
jplatte, line 51:
```suggestion
[Ecosystem > Servers](/ecosystem/servers) section of this website. Most of them
```

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+are open source, so you can explore how they work. Please also refer to their
+documentation if you want to deploy them either for testing or in production.
+
+### Client
+
+Homeservers communicate between each other with the Server-Server / Federation
+API, but they also communicate with clients in a standard way: the Client-Server
+API.
+
+Clients are pieces of software who can use a Matrix account to send and receive
```
jplatte, line 61:
```suggestion
Clients are pieces of software that can use a Matrix account to send and receive
```

jplatte: (I'm not sure the previous is gramatically wrong, but this sounds better to me)

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+
+If you're more interested in writing your own client to bring a new experience
+to users, you may want to rely on an existing SDK (see
+[Ecosystem > SDKs](/ecosystem/sdks)). Those will do significant part of the
+Matrix heavy-lifting and allow you to focus on the UX you want to build.
+Finally, if you're interested in learning more about the interactions between
+clients and servers, please head to
+[the Client-Server section of the Matrix Specification](https://spec.matrix.org/latest/client-server-api/).
+
+### AppService (bots and bridges)
```
jplatte, line 81: It's not actually that common for bots to be appservices, so this heading seems a bit misleading.

content/docs/developers/_index.md
```diff
@@ -0,0 +1,145 @@
...
+
+But sometimes you need to get a more global view of what is happening on your
+homeserver to take action. If you want to write an anti-spam module for example,
+you want to be able to read each and every message from public rooms to detect
+patterns and ring the alarm or take action directly.
+
+To do it with a bot, you would need to invite the bot in each and every room
+where you want the monitoring to happen. An appservice is able to monitor all
+the _unencrypted_ events (messages sent/edited/redacted, people joining or 
+leaving rooms).
```
jplatte, line 102: It is? 😄

Don't you need either the main user or a user from the appservice's user namespace in the room?

content/docs/users/_index.md
```diff
@@ -11,7 +11,6 @@ updated = "2022-10-18T16:00:00Z"
 ## Matrix for Instant Messaging
 
 Matrix works a little like email, but instantaneous and secure:
-
```
jplatte, line 14: Why?

content/docs/developers/_index.md
MTRNord: I would kinda disagree. While in the public that is true there are some non public projects using appservices to be able to do bots at scale. One public project may be mjolnir which seems to be looking into using appservices to provide itself to multiple users.

So I think it makes sense to list both there. However it might make sense to say that appservices for bots only make sense of you have to deal with a scale where using /sync or a single bot user may not work anymore.
As in: appservices for bots are mostly a thing you need for scaling and should not be the default for bots.

jplatte: I'm just saying that bots are not a kind of appservice, which the title reads like to me. I'm totally in favor of having client-like and appservice-like bots both documented.

content/docs/users/_index.md
thibaultamartin: Oh whopsey, that's a rebase that didn't go well 🤦

content/docs/developers/_index.md
thibaultamartin: Hm unsure about this one. Are the events themselves sending messages? Technically, the events _are_ the messages, aren't they?

jplatte: > Technically, the events *are* the messages, aren't they?

Yes, that is what this say if you remove those two words as suggested, no?

content/docs/developers/_index.md
```diff
@@ -13,20 +13,20 @@ last_updated = "2022-10-18T16:00:00Z"
 
 A homeserver is a piece of software hosting accounts of Matrix users. It is
 bound to a single domain that cannot change over time. The accounts on a server
-have an identifier made of a local part, and a server part. A typical identifier
-would be
+have an identifier made of a local part (the user id), and a server part, which
```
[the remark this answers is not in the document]
thibaultamartin: Eh coffee is not strong enough this morning…

content/docs/developers/_index.md
thibaultamartin: Ok I really need another coffee, for some reason I didn't see the "sending" was removed too.

content/docs/developers/_index.md
thibaultamartin: From the bridge crew:

>  For your appservice to monitor what's happening in a room, its HS needs to have a local copy of the room, so a local user needs to be said room, you don't need to have a user from within the namespace in the room

> if my appservice has the namespace foo-*, it will be able to monitor all the unencrypted events within all the rooms starting with #foo- that my HS has a local copy of, even if there is no @foo-*:mydomain.tld user in them

But you're right I need to say an appservice is bound to their namespace 👍

content/docs/developers/_index.md
thibaultamartin: I think it's worth keeping `(bots and bridges)` in the title because `appservice` is a very matrixy term. Newcomers will not necessarily know what appservices are. The first paragraph also says:

> Many Matrix bots are non-human clients. They can be built with the same SDKs as regular clients, and instead of showing a UI to display what is happening they will listen to events, parse them, and for example send automatic replies.

I'm wondering what's the best way to make it more obvious that regular clients can be bots too. Should I add (clients and bots) in the `### Clients` title?

jplatte: How about `Appservices (bridges and some bots)`?

thibaultamartin: I went with "some" bots :)

jplatte (approved):

Issue #1520, opened 2022-10-21
Fix the legacy posts imported in Zol


thibaultamartin:
Todo once #1511 #1512 #1513 #1514 #1515 #1516 #1517 #1518 and #1519 have been merged:

- [ ] Replace all `<code>` blocks in the markdown by actual code blocks
- [ ] Replace all `<pre>` blocks by actual code blocks
- [ ] Replace all `<a>` links by `[]()` links
- [ ] Download "legacy" images, add them next to the post markdown, and replace `<img>` tags by `![]()` 
- [ ] Replace all `<ul><li>` by `- ` or `* ` (I'm biased towards the former for the sake of lisibility)
- [ ] Remove `&nbsp;`
- [ ] Replace `<hX>` by X*`#` (from h1 to h5)
- [ ] Remove `<p>` tags
- [ ] Replace `[embed]` with https://regex101.com/r/nQOzcm/1
- [ ] Replace `<b>` tags with `*this is bold*`
- [ ] Replace `<em>` tags with `*this is italics*`
- [ ] Remove `<div>` (let's check none is super important first)

vedant-z:
@thibaultamartin May I work on any issue in this, like removing all <p> tags ?

thibaultamartin:
Hi vedant-z, feel free to give it a go and open a PR against the `zola` branch when you're ready :)

For the sake of expectations management, a disclaimer that the Matrix Foundation will not offer internships around the matrix.org website as part of GSoC.

PR #1528, opened 2022-10-25
Zola blog cleanup round 1

jplatte:
For posterity / as a hint for others who want to do similar things, I was able to automate large parts of this using regex-replacement in VSCode, e.g. `<h1 id="[^"]+">(.+)</h1>` => `\n## $1\n`.

Part of #1520.


thibaultamartin (approved):
Thanks for this first round of clean-up

PR #1554, opened 2022-11-05
Brand new frontpage

thibaultamartin:
Left to do:

- Use case personal messaging
  - [x] Take a screenshot of the same conversation on Element iOS and FluffyChat on Android
  - [x] Display them horizontally aligned on desktop, vertically aligned on mobile
- Use case community
  - [x] Take a screenshot of Cinny on desktop and FluffyChat on Android. FluffyChat displays all the rooms in the space. Cinny displays the timeline of one of the rooms visible in the FluffyChat screenshot 
  - [x] Display them horizontally aligned on desktop, vertically aligned on mobile
- Use case complex apps
  - [x] Showcase Element Call and Thirdroom?
- Learn how it works
  - [ ] Create a SVG of several clients connected to different homeservers, and animate the transmission of events between them. Most homeservers are connected to the public federation, some are in private federation
    - [ ] 1 client sending an event, the event propagates to another client on the same homeserver
    - [ ] 1 client sending an event, the event propagates to another client on a different homeserver
    - [ ] 1 client sending an event, the event propagates clients on the same homeserver and some clients on different homeservers
- Sponsors
  - [ ] Make sure we have an up to date list of corporate sponsors
  - [ ] Link to individual sponsors page as well?


thibaultamartin:
This is how the basic layout looks like on desktop

https://user-images.githubusercontent.com/1688309/200141276-a7527ffa-9280-40cb-814d-d3ccda58f114.mov

and this is how it looks like on mobile

https://user-images.githubusercontent.com/1688309/200141285-1239b98e-7c36-4a9c-9cc0-dc586ba132c9.mov

Obvious problems:
- On mobile the background of the first section looks bad, the white lines are getting on the text. Maybe offset the background on mobile to get less lines?
- The `⌵` symbol is a hack and should be replaced by a proper image
- In the second section, when "Lorem" and "Ipsum" are replaced with actual (tall) screenshots, they overflow beyond their (grand)parent's size. I'm not really sure how to make them shrink to fit
- On Firefox, when using the mouse wheel it's not possible to reach the footer. Scrolling with a touchpad works. On Safari it's the opposite.

thibaultamartin:
This is how it looks like so far

image

thibaultamartin:
Left to do: use the mobile backgrounds on mobile (which will be handed by design on Monday 23).

templates/index.html
```diff
@@ -3,262 +3,65 @@
...
+        <div id="personal_bottom"></div>
+    </div>
+    <div id="use-case-community" class="section">
+        <h2>A cozy, safe, supercharged place for your community</h2>
+        <div class="call-to-action">Reclaim your Space</div>
+        <img src="assets/frontpage/community-graphic.svg" alt="An artist view of a chat community." />
+    </div>
+    <div id="complex_apps" class="section">
+        <h2>Build advanced, rich communication applications on top of Matrix.</h2>
+        <img src="https://via.placeholder.com/536x320.png" alt="A diagram depicting federation." />
```
jplatte, line 30:
```suggestion
        <img src="https://via.placeholder.com/536x320.png" alt="A diagram depicting Matrix homeserver federation." />
```

sass/_index.scss
```diff
@@ -1,335 +1,424 @@
...
-        height: 600px;
-        background-image: url('/assets/hero-spiro.svg');
-        background-position: 100% 50%;
-        background-size: cover;
-        background-repeat: repeat;
-        background-attachment: scroll;
+#frontpage {
 
+    #hero {
+        height: calc(100vh - 60px);
```
jplatte, line 4: You updated some other `60px`es to `95px`, this one is okay at 60px?

sass/_index.scss
```diff
@@ -1,335 +1,424 @@
...
+
+        .hero_content {
+            height: 100%;
+            display: flex;
+            flex-direction: column;
+            justify-content: center;
+
+            h2 {
+                color: #fff;
+                font-size: 40px;
```
jplatte, line 23: I would probably declare this relative to the parent, but not super important.

sass/_index.scss
```diff
@@ -1,335 +1,424 @@
...
+            background-color: black;
+            border-radius: 100% 100%;
 
-.thisismatrix {
-    h1 {
-        text-align: center;
-        font-weight: bold;
-        padding: 0em;
+            left: 50%;
+            transform: translateX(-50%);
```
jplatte, line 155: What does this do actually? I wonder if a transform is needed.

sass/_index.scss
```diff
@@ -1,335 +1,424 @@
...
-    h2 {
-        margin-top: 20px;
-        margin-bottom: 20px;
-    }
+        .call-to-action {
+            background-color: #fff;
+            color: #000;
+            font-size: 18px;
+            font-weight: 600;
+            border-radius: 1000px;
```
jplatte, line 192: Is this making an element round? I think you can do `50%` instead.

sass/_index.scss
thibaultamartin: This is used to centre the circle at the bottom of the div. I couldn't find another way to do it than with the `transform` that would work at all screens width. Is `transform` that expensive? It seems to be [widely supported](https://caniuse.com/?search=transform) now

sass/_index.scss
thibaultamartin: Using `50%` doesn't seem to produce the expected result. It makes the whole div an ellipse, instead of just rounding the corners.

sass/_index.scss
thibaultamartin: I'm not sure which part would be relative to the parent :) ?
Do you mean the font-size, in `(r)em` instead of `px`?

sass/_index.scss
jplatte: Centering is certainly not something I would expect a transform to be needed for, does `display: flex` + `justify-content` / `align-items` not work? Either way, not that important.

sass/_index.scss
jplatte: Okay, nevermind then.

sass/_index.scss
thibaultamartin: It doesn't work because the item displayed is overflowing (with overflow hidden). In such a case `justify-content` and `align-items` seem to left-align the item, unfortunately. I've spent quite a bit of time scratching my head on this one :D

sass/_index.scss
thibaultamartin: Solved in 3809be9

sass/_index.scss
thibaultamartin: Solved in d1e56d3

sass/_index.scss
jplatte: Yes, I mean having font-sizes (except at the root) in `%` or `em`.

jplatte (approved):

PR #1555, opened 2022-11-07
Clients page template cleanup

jplatte:


thibaultamartin (approved):
Looks good to me, very elegant way of handling it

PR #1560, opened 2022-11-13
Remove Nio from clients

kiliankoe:
Unfortunately Nio is not actively maintained anymore. Maybe that'll change again in the future, maybe it'll turn into something else, I don't know. But lots of people keep stumbling across it on matrix.org and get their hopes and expectations up that I can't fulfill right now :/ 

Preview: https://pr1560--matrix-org-previews.netlify.app


kiliankoe:
@tulir has suggested to instead mark it as not actively maintained and not featured, perhaps that would make more sense?

PR #1563, opened 2022-11-14
Update clients/Nio metadata

kiliankoe:
Preview: https://pr1563--matrix-org-previews.netlify.app


thibaultamartin:
Thanks for keeping us posted Kilian. It's always a little sad to see a project going dormant, but it's been nice to have Nio around :)

PR #1571, opened 2022-11-22
Display subsections in header menu

thibaultamartin:


thibaultamartin:
Screenshot 2022-11-22 at 14 25 28

Menus and submenus will need a bit more love to look really good so I just added a tiny border radius to make the submenus look less rough. The point is mostly to display the docs submenu here.

jplatte (approved):

PR #1582, opened 2022-12-06
List all docs under the legacy section

thibaultamartin:
Note:
- Docs are in no special order, but I don't think that's a big deal since they are legacy docs
- There are no subsection, everything under `legacy` is a page


jplatte (commented):
I think the page order is actually going to be different from run to run if you don't specify an order at all. We should sort by title if we don't want something special

jplatte (approved):

PR #1583, opened 2022-12-07
Implement the SDKs page

thibaultamartin:
The structure for the SDKs page has been laid out, as well as the data structure.

I still need to import the actual data from current matrix.org

### DIFF
--- a/content/ecosystem/sdks/index.md
+++ b/content/ecosystem/sdks/index.md
@@ -5,11 +5,16 @@ weight = 3
 
 SDKs are pieces of software meant for developers, that take care of the
 technical Matrix protocol calls behind the scenes. They allow them to focus on
-where they can add the most value.
+where they can add the most value. If you're not a developer, this section of
+the website is probably not for you.
 
-There are two types of SDKs: clients (including bots) and appservices (used for
-bots and sdks). If you are a developer and are not sure what the difference
-between clients and appservices, please head to [the Developers documentation](/docs/developers/).
+There are two fundamental types of SDKs: clients and appservices. This
+distinction is quite academical, and we went for a more pragmatic approach. You
+will find below SDKs by type of product you may want to create (namely: clients,
+bots, bridges). Don't be surprised if a same SDK appears in several sections!
+
+If you are a developer and are not sure what the difference between clients and
+appservices, please head to [the developers documentation](/docs/developers/).
 
 ## Client SDKs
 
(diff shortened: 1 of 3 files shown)


content/ecosystem/sdks/sdks.toml
```diff
@@ -0,0 +1,20 @@
+[matrix-rust-sdk]
+language = "Rust"
+licence = "Apache-2.0"
+repository = "https://github.com/matrix-org/matrix-rust-sdk"
+purpose = ["client", "bot"]
```
jplatte, line 5:
```suggestion
purpose = ["client", "bot", "bridge"]
```

unless you want to list the `matrix-sdk-appservice` Rust crate separately? (it's in the same repository)

content/ecosystem/sdks/index.md
```diff
@@ -2,3 +2,58 @@
...
+
+The following bridges SDKs are widely used in the Matrix ecosystem and are
+recommended by the Foundation:
+
+{{ sdk_table(type="bridge", featured=true) }}
+
+If you're not happy with the recommended SDKs, you can try one of the following
+ones as well.
+
+{{ sdk_table(type="brudge", featured=false) }}
```
jplatte, line 59:
```suggestion
{{ sdk_table(type="bridge", featured=false) }}
```

content/ecosystem/sdks/sdks.toml
```diff
@@ -0,0 +1,20 @@
...
+repository = "https://github.com/maubot/maubot"
+purpose = ["bot", "bridge"]
+featured_in = ["bot", "bridge"]
+
+[nio]
+language = "Python"
+licence = "Apache-2.0"
+repository = "https://github.com/poljar/matrix-nio"
+purpose = ["bot"]
+featured_in = []
```
jplatte, line 20:
```suggestion
```

templates/shortcodes/sdk_table.html
```diff
@@ -0,0 +1,30 @@
...
+            <th>Licence</th>
+            <th>Repository</th>
+        </tr>
+    </thead>
+    <tbody>
+    {% for name, sdk in sdks %}
+    {% if not featured and sdk.featured_in is not containing(type) %}
+        {% set display = true %}
+    {% elif featured and sdk.featured_in is containing(type) %}
+        {% set display = true %}
```
jplatte, line 16:
```suggestion
    {% if featured and sdk.featured_in is containing(type)
        or not featured and not sdk_featured_in
        or not featured and sdk_featured_in is not containing(type) %}
        {% set display = true %}
```

jplatte (approved):

templates/shortcodes/sdk_table.html
jplatte: (I checked, while Tera doesn't have parentheses, `and` has higher precedence than `or`, like everywhere else)

thibaultamartin: Is it safe to assume that `or not featured and not sdk_featured_in` is for the case where there is no `featured_in` for a specific sdk in `sdks.toml`?

I guess `containing(type)` returns false if `featured_in` does not exist, and `not containing(type)` returns true… which seems to already cover the case?

Is `sdk_featured_in` a typo for `sdk.featured_in`?

jplatte: Yes, that one underscore should be a period, and I was assuming that `containing(type)` would throw an error if the LHS is undefined but if it just returns `false`, you can omit the middle condition.

thibaultamartin: It does throw an error… but it still does with the suggestion (after s/_/./)

It looks like it's trying to evaluation the if as a whole and doesn't stop at the first `or`. I'm tempted to make the `featured_in` field mandatory in the toml to avoid over-complexifying the template. Wdyt?

jplatte: So it doesn't short-circuit? 😬
I would go for nested ifs then, but that's just personal preference. We can keep the field mandatory.

PR #1586, opened 2022-12-07
Remove Google Hangout from Bridges

mpldr:
Since Google Hangout has been shut down, keeping it's bridges listed on the Website does not make sense.

Link: https://killedbygoogle.com/


mpldr:
Cleaned up the history a bit.

PR #1592, opened 2022-12-15
Push sdks data from current m.org

thibaultamartin:


jplatte (approved):
Shouldn't every SDK have some value in `purpose`?

Issue #1505, opened 2022-10-14
Rework the documentation


thibaultamartin:
## Goal of the page

The documentation aim to be as product-agnostic as possible. In some specific cases (such as when talking to the general public), the flagship clients are used to make things more concrete and easier to understand.

Product-specific documentations belong to the products themselves and should not be hosted here.

Legacy documentations need to be ported to Zola, and redirections need to be set-up so the old links still work.

- Everyday People should find everything they need to get started using Matrix with their family and friends
  - Create a Matrix account (opinionated: on matrix.org because the general public doesn't care about homeservers, using Element because it's the flagship client), invite friends, create a small group chat
  - Follow a matrix.to link to join a small group chat
- Community Managers should be able to
  - Create an account on matrix.org (with an explanation that they can get their own domain eventually)
  - Create rooms
  - Understand aliases and matrix.to links
  - Understand Power Levels
  - Understand moderation
  - Switch providers (either by picking one from the hosting page, or by becoming one themselves)
- Developers should be redirected
  - To product-specific tutorials
  - To the ecosystem pages
  - To the spec
- Entrepreneurs TBD

## Content of the page

The landing page of the documentation section should contain tiles on top:

- One with a large 🧑, and the title "I want to start chatting using Matrix"
- One with a large 👪, and the title "I want to use Matrix for my community or organisation"
- One with a large ⚒️, and the title "I want to develop a client, bridge or bot"
- One with a large 🕸️, and the title "Legacy Docs"

aaronraimist:
> The documentation aim to be as product-agnostic as possible. In some specific cases (such as when talking to the general public), the flagship clients are used to make things more concrete and easier to understand.

In some specific cases, such as the general case that most people fall into 🤔

I'm guessing you are already going to need to create two different versions for the getting started tutorial, one for Element mobile and another for Element Web/Desktop. It wouldn't be *that* much more work to create a couple more for let's say FluffyChat and Cinny.

> Create a Matrix account (opinionated: on matrix.org because the general public doesn't care about homeservers

They don't care because they don't know they should care. The whole point of documentation is to provide instructions and best practices for how to do something.

Most likely they are coming from another instant messaging app which has no concept of picking a server. The site should give a brief and gentle introduction to the topic. It doesn't have to be detailed and the site can still suggest matrix.org but it should also at least vaguely mention that you can pick a server  run by your friends, yourself, or an online community.

> Create an account on matrix.org (with an explanation that they can get their own domain eventually)

[...]

PR #1617, opened 2023-01-25
Moderation guide update

thibaultamartin:
Part of #1505


content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
 ```
 
-TODO screenshot of oping mjolnir
+The bot really does need those privileges to be able to actually kick and ban
+people from the rooms it protects, or change other people's power level. Keep in
+mind that whoever has access to Mjolnir credentials effectively has powerlevel
+100 in those rooms. In this configuration Mjolnir is de-facto the owner of your
+community rooms. Some more advanced configurations allow you to keep more
+privileges than mjolnir, but they are quite tedious to set-up. We will not cover
+in this documentation.
```
jimmackenzie, line 102: Some minor rewording tweaks to make it clearer that there is a common setup which works, but which has some considerations the user should think about. Then we point them to the material where they can learn more, and decide if they want a more advanced setup. I also simplified the language and sentence structure.

```suggestion
You should spend some time thinking about what privileges you want to give to your Mjolnir. A common setup is to give the Mjolnir user power level 100. This gives Mjolnir the permissions to kick and ban people from the rooms it protects. This has a couple of implications. First, anyone with access to the Mjolnir has power level 100 in the rooms it protects. Also, Mjolnir is the de-facto owner of those community rooms. For more advanced set-ups, read the [spec covering power levels](https://spec.matrix.org/v1.5/client-server-api/#mroompower_levels).
```

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
 
 ## Getting moderation reports
 
-- You need your own instance while Aristotle isn't merged in the spec and
-  deployed in the wild?
+By default, the moderation reports go to the reporter's homeserver
+administrator. This is suboptimal for a variety of reasons. If the person
+running mjolnir and the person running the homeserver are the same, mjolnir can
+intercept some calls to display reports in its command room instead. Such
+reports look like below.
```
jimmackenzie, line 110: I don't agree that the default is suboptimal. It is a choice, and there are benefits and drawbacks to each approach.

```suggestion
By default, the moderation reports go to the reporter's homeserver
administrator. If you are the homeserver admin, you can choose to send the reports to the Mjolnir management room. Those reports look like this:
```

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+to shout in the void, without realising it.
 
-## Working together
+Shadow-banning is not part of the Matrix specification itself, which is why it
+is implemented in the non-standard [Admin API of Synapse](https://matrix-org.github.io/synapse/latest/usage/administration/admin_api/index.html).
 
-- Subscribing to ban lists
-- Publishing yours
+Homeserver administrators who want to resort to shadow banning can read about it
+[in Synape's documentation](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#controlling-whether-a-user-is-shadow-banned).
```
jimmackenzie, line 372: I'd maybe omit this section entirely?

jimmackenzie (commented):
I've added some comments and some suggestions, this looks promising!

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+follows the usual `#room_name:server.tld` format.
+
+Then to make mjolnir follow this list, you need to issue the following command
+in its control room
+
+```
+!mjolnir watch <room alias/ID>
+```
+
+For example to subscribe to the `#matrix-org-coc-bl:matrix.org` ban list
```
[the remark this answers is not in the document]
MTRNord: Well the tos list only applies the matrix-org homeserver terms of server. Not the overall matrix code of conduct. So the tos usually is not the one to use. Tos is meant to mainly apply to the homeserver while COC is a more general document applying to a greater part of matrix-org. (like the repos too iirc).

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
 /op @yourMjolnirBot:example.com 100
 ```
 
-TODO screenshot of oping mjolnir
+The bot really does need those privileges to be able to actually kick and ban
```
MTRNord, line 96: It would be nice to explain (the arguably harder) way of having it only Moderator (pl 51 for example) and allowing acl for pl51. That way the bot isn't an admin which always is annoying when you need to remove it later on in case a hs gets unreachable. If it is 51 and has acl perms it still works but admins are also able to remove it when needed. Mostly a quality of life change imho.

MTRNord: Obviously this makes it so you can't op people anymore fully with it.

MTRNord: (turns out I missed the next few sentences :)

How about documenting it behind a spoiler thingy as an advanced section?

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+100 in those rooms. In this configuration Mjolnir is de-facto the owner of your
+community rooms. Some more advanced configurations allow you to keep more
+privileges than mjolnir, but they are quite tedious to set-up. We will not cover
+in this documentation.
 
 ## Getting moderation reports
 
-- You need your own instance while Aristotle isn't merged in the spec and
-  deployed in the wild?
+By default, the moderation reports go to the reporter's homeserver
```
MTRNord, line 106: I fear it won't get updated again but maybe add a link to the msc element Web implements that sends them to mods instead?

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+
+### Redacting a specific message
+
+If you have moderation reports enabled and click on the
+{{ mjolnirbutton(text="🗍 Redact") }} button below it, mjolnir will redact this
+message. Alternatively, if you don't have reports enabled or want to redact a
+specific message from mjolnir, you can instruct mjolnir to do so by sending a
+message in mjolnir's control room.
+
+To redact a specific message, you need to find its permalink. A permalink is a
```
MTRNord, line 135: Iirc the event id alone is enough. Not sure if that's easier or harder to get. It may depend on the client.

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+!mjolnir redact <event permalink> 
+```
+
+For example
+
+```
+!mjolnir redact https://matrix.to/#/!yOatELRSQXzfQMmxjH:matrix.org/$F76L2figPEC240TFaUkHoKPxxhJ3P54vP4hi14Sd8xw?via=matrix.org&via=t2bot.io
+```
+
+You may wonder why you should use mjolnir if you can simply redact the message
```
MTRNord, line 157: Maybe it should be mentioned that redactions from mjolnir primarily are useful for mass redactions? As in redacting every message of a user. Since mjolnir is set up to not be rate limited it usually doesn't fail like a client would.

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+As you probably guessed, you are not going to ban a user but a server this once.
+For example, to ban the `maliciousdomain.tld` domain:
+
+```
+!mjolnir ban spam server maliciousdomain.tld
+```
+
+Mjolnir is going to add the domain to the `spam` ban list, and whenever someone
+from that homeserver tries to join your community they are going to be banned.
+
```
MTRNord, line 287: Missing is banning/disabling rooms for server admins. Or I guess in general the Synapse plugin is missing.

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+!mjolnir list create spam my-community-spam-ban-list
+```
+
+You only need to create a ban list once, and can then add as many users and
+servers as you want to that ban list. You can also create as many ban lists as
+you want.
+
+Communities often create two ban lists: one for spam, and one for code of
+conduct infringements.
+
```
MTRNord, line 320: Maybe a mention how to set a list as a default would also be useful.

MTRNord, line 320: Maybe a mention how to set a list as a default would also be useful.

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+```
+
+For example to subscribe to the `#matrix-org-coc-bl:matrix.org` ban list
+maintained by the Matrix Foundation, you would issue the following command
+
+```
+!mjolnir watch #matrix-org-coc-bl:matrix.org
+```
+
+### Publishing ban lists
```
MTRNord, line 350: This probably can moved into the creation of the lists. It feels out of place here and isn't long enough to warrant a full section imho

MTRNord (changes requested):

content/docs/communities/moderation/_index.md
```diff
@@ -95,21 +93,280 @@ in the message bar at the bottom:
...
+specific message from mjolnir, you can instruct mjolnir to do so by sending a
+message in mjolnir's control room.
+
+To redact a specific message, you need to find its permalink. A permalink is a
+regular matrix.to link that you can obtain by hovering the message you want to
+delete, and clicking on the `…`, then on `Share`.
+
+![](./share_message.png)
+
+This will open a pop-up window from which you can copy the matrix.to permalink.
```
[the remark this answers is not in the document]
thibaultamartin: Fixed in [774bd87](https://github.com/matrix-org/matrix.org/pull/1617/commits/774bd8761340673270fd454fbe3178e6141f8c02). I don't think it's worth adding a sentence, since there's an example below with a matrix.to link :)

content/docs/communities/moderation/_index.md
thibaultamartin: Fixed in [4d55d99](https://github.com/matrix-org/matrix.org/pull/1617/commits/4d55d999d3bff150df01f55a68ddb3ce10afaaa9)

content/docs/communities/moderation/_index.md
thibaultamartin: Documenting it behind a spoiler is a good idea! I'd like to land a first version quickly though, so I think it makes sense to go without it at first.

I'm also wondering if this should be documented on the Matrix.org "running your own community" documentation or if it should end up in mjolnir's own docs?

content/docs/communities/moderation/_index.md
thibaultamartin: Ah, I thought this was made obvious in the intro paragraph. Do you think I missed something? Would you tweak it to make it more visible?

content/docs/communities/moderation/_index.md
thibaultamartin: I'm not sure I understand this suggestion :)

content/docs/communities/moderation/_index.md
thibaultamartin: Fixed in 3ecb5eca

thibaultamartin: Fixed in 3ecb5eca

content/docs/communities/moderation/_index.md
thibaultamartin: Fixed in [bec4c28](https://github.com/matrix-org/matrix.org/pull/1617/commits/bec4c28a9d72b1459bbe492e141a844e8873fd6d)[bec4c28](https://github.com/matrix-org/matrix.org/pull/1617/commits/bec4c28a9d72b1459bbe492e141a844e8873fd6d)

thibaultamartin:
Now that we only have content we want to add, I'm going to rebase and merge this as a first iteration as I'd like to start sharing this with a few external folks on the preview website. We will keep iterating to refine it and add useful content

content/docs/communities/moderation/_index.md
MTRNord: I feel like they could belong in either or both. A community does necessarily have their own homeserver. And equally a mjolnir mod might have a homeserver they want to moderate with mjolnir.

MTRNord: It somewhat only affects mjolnir and not communities in general I think. Like this setup is meant to benefit mjolnir. Maybe it is also helpful for other moderation bots but I feel like it is better in here.

PR #1611, opened 2023-01-16
[Zola] implement the new clients page design.

MTRNord:
# This PR depends on #1554 ! merge that one first!

Screenshot:
image

## Todos

- [x] Responsive design for hero header (svg hell?)

PR #1620, opened 2023-01-25
[zola] Add new client page design

MTRNord:
This is a new branch for #1611


sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
+        font-weight: 700;
+        font-size: 40px;
+        line-height: 120%;
+    }
+
+    p {
+        width: 100%;
+        font-style: normal;
+        font-weight: 400;
+        font-size: 18px;
```
thibaultamartin, line 28: I would use `rem` here as well

```suggestion
        font-size: 1.125rem;
```

sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
+    display: grid;
+    grid-auto-flow: column;
+    grid-template-columns: 1fr 1fr 1fr;
+
+
+    @media (max-width: 767px) {
+        width: unset;
+        grid-template-columns: 1fr;
+        grid-auto-flow: row;
+    }
```
thibaultamartin, line 52: I think it would make sense to use a flexbox + max-width to get the responsiveness "for free" instead of a forcing the layout on one or three columns

sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
-    background-color: rgba(43, 140, 201, .1);
-    border-radius: 1em;
-    padding: 1em;
+    justify-content: center;
+    align-items: center;
+    padding: 24px;
+    gap: 16px;
+
+    width: 281px;
+    height: 343px;
```
thibaultamartin, line 67: I think this would be a good case of using `rem`s instead of `px` and check how it behaves when using a different zoom level (zooming out to 60% and in up to 150% at least).

sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
-        display: flex;
-        flex-wrap: wrap;
-        justify-content: center;
-        margin-top: auto;
+    h3 {
+        margin: 0;
+
+        font-style: normal;
+        font-weight: 700;
+        font-size: 32px;
```
thibaultamartin, line 88:
```suggestion
        font-size: 2rem;
```

sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
+
 .platform-pill {
-    margin: .2em;
-    border-radius: 2em;
-    padding: 0 .6em;
-    color: #fff;
-    background-color: rgba(43, 140, 201, .8);
+    background: #E9E9E9;
+    border-radius: 16px;
+    padding: 6px 8px;
```
thibaultamartin, line 105: Probably worth using `rem`s here again to make the pills scale with the zoom level

sass/_clients.scss
```diff
@@ -1,77 +1,161 @@
...
+
+    font-style: normal;
+    font-weight: 400;
+    font-size: 16px;
```
thibaultamartin, line 109:
```suggestion
    font-size: 1rem;
```

templates/shortcodes/clients_showcase.html
```diff
@@ -7,9 +7,7 @@
         <img src="thumbnails/{{ client.thumbnail }}" alt="{{ name }} Preview">
         <h3>{{ name }}</h3>
         <p>{{ client.summary | safe }}</p>
-        <div class="platforms-links">
-            {{ clients::platform_links(packages=client.packages) }}
-        </div>
+        {{ clients::platform_links(packages=client.packages) }}
```
thibaultamartin, line 10: You may need a div around this, to be able to turn it into a row. At the moment it displays all of the platforms vertically, likely because of the `flex-direction: column;` of its direct `.featured-client-card` parent.

thibaultamartin (changes requested):
Thanks a bunch for the PR! There are a few tweaks needed, but that's already a great shot

sass/_base.scss
```diff
@@ -250,3 +250,57 @@ pre table {
...
+/* The black block on pages at the top. */
+.page-header {
+    width: 100%;
+    height: 377px;
+    background: #000000;
+    overflow: hidden;
+
+    .bg {
+        width: 100%;
+        position: absolute;
```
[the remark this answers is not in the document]
MTRNord: It is needed to make sure the text can be on top of the image div. That's why the wrapper also has an explicit relative.

sass/_base.scss
```diff
@@ -250,3 +250,57 @@ pre table {
...
+        height: 377px;
```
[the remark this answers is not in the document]
MTRNord: I need to double check but iirc due to the absolute positioning width 100% here wouldn't be the parent size. But I am not sure anymore.

MTRNord:
Will take a look later today or tomorrow :) agreeing with the rem stuff :)

templates/shortcodes/clients_showcase.html
thibaultamartin: I didn't understand what could cause the logos to go out of the cards, but it might very well be the platform links being vertically displayed instead of horizontally displayed

image

MTRNord: Huh. That's not what this looks like in my browser 😱 in mine it is like in the mockup. That looks like I am using something that your browser doesn't support?

thibaultamartin: I'm using `Firefox 109.0 (64-bit)` on `macOS Ventura 3.1 (22C65)`, nothing too fancy or exotic, and no WebExtension fiddling with the content (uBlock Origin would be the most likely to, but blocked 0% of requests on this page)

MTRNord: That's weird 🤔 on my Firefox it looked fine when writing it. 😱 That's weird.

thibaultamartin (approved):

PR #1629, opened 2023-01-28
Make the menu responsive

### LINKED ISSUE #1545: Style the Zola site header
The site header on the zola branch has a new feature: Sub-items for a few of the header links. We need to write some CSS to make it look good, both on desktop and on mobile.

thibaultamartin:
Closes #1545

### DIFF
--- a/sass/_header.scss
+++ b/sass/_header.scss
@@ -114,6 +114,30 @@
         label {
             cursor: pointer;
         }
+
+        details {
+            font-size: 1.5rem;
+            padding: 8px 24px;
+
+            summary{
+                display: flex;
+                justify-content: space-between;
+                margin: .5rem 0;
+
+                .arrow {
+                    background-image: url("/assets/down-arrow.svg");
+                    width: .8rem;
+                    height: .8rem;
+                    margin: auto 0;
+                    filter: invert(1);
+                }
+            }
+        }
+
+        details[open] summary .arrow {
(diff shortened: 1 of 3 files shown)


thibaultamartin:
Current status:

image

Left to do:
- [ ] Make the sections collapsible
- [ ] Make the menu itself scrollable

thibaultamartin:
Current look

## Desktop

image

## Mobile by default

image

## Mobile with a section expanded

image

sass/_header.scss
```diff
@@ -49,12 +49,12 @@
         @media screen and (max-width: 1023px) {
             position: fixed;
             z-index: 800;
-            top: 60px;
+            top: 5.937rem;
```
jplatte, line 52: This is a pretty weird value to use ^^
Maybe put it in a variable?

jplatte (approved):

sass/_header.scss
thibaultamartin: The point of using `rem`s was to allow the header to scale up when you zoom in. It's to avoid having the text overflowing from the navbar. I agree a variable would make sense, but I'm not entirely sure if there are best practices for using variables in SCSS. I'm wondering whether to start fiddling with variables right away or to wait for the great post-launch refactoring.

jplatte: Doesn't hurt starting now I would say. After this PR ;)

PR #1634, opened 2023-01-31
Zola Responsive Menu (iPhone compliant)

thibaultamartin:


templates/includes/header.html
```diff
@@ -10,27 +10,12 @@
...
         {% set href = "/" ~ link.section ~ "/" %}
         {% set section = get_section(path=link.section ~ "/_index.md") %}
-        <details {{ util::current_open(if_path_starts_with=href) }} class="section-wrap mobile">
-            <summary class="section {{ util::current_class(if_path_starts_with=href) }}">
-                {{ link.title }}
+        <div class="section-wrap">
+            <input id="{{ link.slug }}-submenu-checkbox" type="checkbox" class="submenu-checkbox" aria-hidden="true">
+            <div class="submenu-title-container">
+                <label for="{{ link.slug }}-submenu-checkbox" class="submenu-title">{{ link.title }}</label>
                 <div class="arrow"></div>
```
jplatte, line 17: Why is the arrow not inside the `<label>`?

sass/_header.scss
```diff
@@ -207,3 +187,50 @@
         }
     }
 }
+
+.submenu-checkbox {
+    display: none;
+
+    @media screen and (max-width: 1023px) {
+        &:not(:checked) + .submenu-title-container + .section-submenu-wrap {
```
jplatte, line 195: Given you moved the different submenu checkboxes into distinct parent containers (`section-wrap`), you can shorten this to

```suggestion
        &:not(:checked) ~ .section-submenu-wrap {
```

sass/_header.scss
```diff
@@ -207,3 +187,50 @@
...
+            display: none;
+        }
+
+        &:not(:checked) + .submenu-title-container {
+            .arrow {
+                background-image: url("/assets/down-arrow.svg");
+                width: 1rem;
+                height: 1rem;
+                margin: auto 0;
+                filter: invert(1);
```
jplatte, line 205: Are we using the same image as black colored elsewhere? I'm almot inclined to recomment `{% include %}`ing it instead of using `background-image` and setting `fill: #fff` via CSS on the contained `<path>`. Also why does that file contain more than a dozen empty SVG groups? o_O

templates/includes/header.html
thibaultamartin: Mostly to use `submenu-title-container` as a `flex` so I had more room to play with the arrow itself. Having a `div` inside a `label` also felt a little wrong.

Should I remove the container div and move the arrow inside the label?

sass/_header.scss
thibaultamartin: We are using the same image black coloured in the documentation yes. Is it more efficient to use `fill` than `background-image` and `invert`?

I'm going to give [scour](https://github.com/scour-project/scour) a go on these assets :)

thibaultamartin:
```
Scour processed file "down-arrow.svg" in 12 ms: 602/857 bytes new/orig -> 70.2%
…
Scour processed file "up-arrow.svg" in 6 ms: 631/896 bytes new/orig -> 70.4%
```

templates/includes/header.html
jplatte: I think you should yes, as right now tapping the arrow doesn't expand the menu.

templates/includes/header.html
```diff
@@ -10,27 +10,10 @@
         {% if link.section %}
         {% set href = "/" ~ link.section ~ "/" %}
         {% set section = get_section(path=link.section ~ "/_index.md") %}
-        <details {{ util::current_open(if_path_starts_with=href) }} class="section-wrap mobile">
```
jplatte, line 13: You removed the use of `current_open`, maybe you should turn it into `current_checked` instead?

sass/_header.scss
thibaultamartin: I'm not sure I parsed the last sentence correctly, I read it as "you're using the wrong tool for the job, it's not a problem" :D

templates/includes/header.html
thibaultamartin: This one actually always returned false, because `current_path` is not defined in `header.html`

sass/_header.scss
jplatte: Yes, that is what I was saying. You are using a much more powerful tool than necessary to do the coloring, so there is more potential for issues, but in practice it will likely be perfectly fine.

thibaultamartin:
@jplatte whenever you have time, I could use a hand to make the mobile menu scrollable vertically before merging this one :)

jplatte (approved):

PR #1641, opened 2023-02-02
Zola refactor

thibaultamartin:


thibaultamartin:
After discussing it internally we decided not to do this refactor since it could make navigation slower, particularly on mobile.

At the moment, the CSS for the whole website is bundled in single CSS file at build time. This means that the CSS for the whole website is loaded once and then cached for all pages.

Splitting CSS per-page would mean generating an additional request to retrieve the page-specific CSS, which means all pages would load significantly slower on poor connectivity (one request to load the page, parse the html, one additional request to load the css file for that page).

Given the tiny size of the full CSS file and the fact that on mobile bandwidth is much less an issue than latency, we're keeping a single CSS file for the whole website.

During the conversation we also mentioned a plan for the future if the CSS file became significantly heavier:
- Splitting the CSS in two files: one for the frontpage, and one for the rest of the website.
- Once on the frontpage, we have the possibility to [prefetch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ) the CSS for the rest of the website in the background (with the caveat that [it doesn't work on Safari, hence on iPhones](https://caniuse.com/link-rel-prefetch))

The plan above would allow to optimise first page speed, while potentially decreasing the load time for the rest of the website. It has been deemed overly complex for now but is worth keeping in mind.

Issue #1508, opened 2022-10-16
Nonfunctional menu button in narrow view


AndrewKvalheim:
On [/faq/](https://matrix.org/faq/), when the window is less than 992px wide, a menu button appears at the top right corner:

> <img alt="screenshot of 991px wide viewport" src="https://user-images.githubusercontent.com/1844746/196057782-03a4bd54-8c4f-4d46-b94e-cee989fb2f35.png" width="991">

It does nothing when clicked. The browser console reports what looks to be a malformed jQuery selector tracing back to Webflow:

```
Uncaught Error: Syntax error, unrecognized expression: #what-is-matrix%3F
```

Prakhar1701:
Hi @AndrewKvalheim, I am interested to work on this issue, please assign me this issue.

snehilchhabria:
I can work on this issue if I am guided with clear instructions on the required solution.

AshishPandagre:
I believe PR #1629 will fix this.

harshlancer:
Is this issue still open to work?

horizenight:
Still open for work ?

indrasn0wal:
Please assign me the issue

tony83033:
Hii @AndrewKvalheim sir could you pls assign me this issue

anurag464:
Please brief, what the menu button should show?

realvineeths:
U can check the readme for building after u have cloned it

How to build

    Make sure you have the prerequisites:
        Node.js and npm
        git clone https://github.com/matrix-org/matrix.org && cd matrix.org
    cd gatsby
    Run npm i to install dependencies
    npm run develop - It will take a while on your first build

Now you can visit the page you want to work on in the browser. For example: http://127.0.0.1:8000/docs/projects/try-matrix-now/

VaibhavAPatil:
I am **interested** in **this Issue**. Can I do **Pull Request** or you can **assign** this **issue to Me**.

Chaitu-Tatipamula:
hii i wanna get into this can you people help me out

ABHIXIT2:
Am ii am using it on browser and it is working operfectly fine..

B1TSH3LL:
i would like to work on this issue. please assign this to me.

richvdh:
> i would like to work on this issue. please assign this to me.

[...]

PR #1645, opened 2023-02-04
Fixed unrecognized expression: #what-is-matrix%3F

realvineeths:
Closes https://github.com/matrix-org/matrix.org/issues/1508

The issue is solved by eliminating the special characters, which used to show Uncaught Error: Syntax error, unrecognized expression.
Now its good to go.

### DIFF
--- a/gatsby/src/pages/faq.js
+++ b/gatsby/src/pages/faq.js
@@ -37,10 +37,10 @@ const Faq = ({ data }) => {
               Basics
             </h3>
             <div className="question">
-              <h4 id="what-is-matrix%3F">
+              <h4 id="what-is-matrix3F">
                 <a
                   className="permalink"
-                  href="#what-is-matrix%3F"
+                  href="#what-is-matrix3F"
                   aria-hidden="true"
                 >
                   &#128279;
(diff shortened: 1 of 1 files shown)


MTRNord:
This doesn't seem to solve that issue but instead makes these links invalid.

realvineeths:
> This doesn't seem to solve that issue but instead makes these links invalid.
But its working when i ran locally in my machine.

MTRNord:
> The link would be invalid if the href link does not match the id, but i have changed even the id. Please correct me if that would lead to an error.

Turns out indeed the links are valid. Does the icon still work though? The %3F afaik refers to the icon. so deleting % might break this

realvineeths:
> > The link would be invalid if the href link does not match the id, but i have changed even the id. Please correct me if that would lead to an error.
> 
> Turns out indeed the links are valid. Does the icon still work though? The %3F afaik refers to the icon. so deleting % might break this
There is no issue of such all the icons are still working.

thibaultamartin:
Closing this PR as obsolete since it doesn't target the rewrite of the website

PR #1647, opened 2023-02-07
Add missing CTAs

thibaultamartin:
We're not adding missing cats, but adding Call To Action buttons on the frontpage. Those new sections are pretty, but their point is to route visitors to the most relevant section following their interest :)


jplatte (approved):
> We're not adding missing cats, but adding Call To Action buttons

But

Why_not_both

Issue #454, opened 2019-06-11
https://matrix.org/docs/spec/appendices.html needs a proper page title


uhoreg:
https://matrix.org/docs/spec/appendices.html needs a proper page title instead of `spec_appendices.rst`

PR #1652, opened 2023-02-11
Fix: Updated title of appendices.html from spec_appendices.rst to Appendices

bhimanshukalra:
closes https://github.com/matrix-org/matrix.org/issues/454

Updated title of https://matrix.org/docs/spec/appendices.html from spec_appendices.rst to Appendices


turt2live (approved):
Thanks! Though please note that this spec document is not intended to be used anymore - please use spec.matrix.org instead

Issue #794, opened 2020-07-26
"iot through matrix" video not found (404)


pedro-nonfree:
https://matrix.org/blog/2015/04/08/video-iot-through-matrix

->

https://matrix.org/video/IoT-through-Matrix.mp4

benparsons:
@ara4n do you know if we can recover this video?

aaronraimist:
The Wayback Machine has it https://web.archive.org/web/20150919135540im_/http://matrix.org/IoT-through-Matrix.mp4

awtj8o81ryywg793:
YouTube also hosts a copy of the video: https://www.youtube.com/watch?v=YJHZSLNTtIU

PR #1653, opened 2023-02-11
Fix: Update video source to correct URL

bhimanshukalra:
closes https://github.com/matrix-org/matrix.org/issues/794

The current video source on https://matrix.org/blog/2015/04/08/video-iot-through-matrix is incorrect.
I've updated it as per the URL mentioned in the comments on the issue. Do let me know if some other URL needs to be used, or if some other approach needs to be used.

Thanks!


thibaultamartin:
This PR targets the outdated gatsby branch because we didn't have a chance to review it before merging in the new website. If you feel like it, please open a new PR to target the `main` branch, and sorry for not getting to this earlier!

PR #1655, opened 2023-02-12
Add filters to clients page

thibaultamartin:


thibaultamartin:
This is not done at all but I took a stab at creating filters on-brand with the rest of the m.org website

image

PR #1660, opened 2023-02-17
Mark Synapse blog posts for review by Synapse team

DMRobertson:
No idea if this will work, please check my working

Preview: https://pr1660--matrix-org-previews.netlify.app


MatMaul (approved):
GitHub UI was complaining about `Unknown owner on line 2: make sure the team @matrix-org/synapse-core exists and has write access to the repository `.
`core-team` already has write access so I added explicitly `synapse-core` too and it's fixed now.

Issue #1570, opened 2022-11-21
List public servers willing to accept new members


thibaultamartin:
Get something as user friendly as https://joinmastodon.org/servers and as complete as https://joinmatrix.org/servers/

Important aspects:
- Servers should opt-in to be listed there
- Data updates need to be automated
- "Use at your own risk, this is not endorsed by the m.org foundation" warnings need to be made, all while not being unnecessarily too scary
- Servers listed should agree to some basic principles?

The three filters joinmastodon.org has make sense to me:

- legal structure (is it an individual or an organisation I'm trusting with my account? Helps with the bus factor)
- sign-up speed (is it completely open? Are there manual reviews? Can help determine the quality of the instance and its likeliness to get banned)
- language (useful to get support from admins)

I believe it also makes sense to add a filter for whether the services are paid or free. This might be a criterion that fits in "legal structure" (e.g. "Individual admin", "Non-profit", "Paid service"). Another way to see it is to have a "Support level" filter with "Best effort (single admin)", "Best effort (team)" and "Paid service with support".

thibaultamartin:
joinmastodon.org

![image](https://user-images.githubusercontent.com/1688309/203257492-62525e2b-7eab-4ad4-a9c1-f474a85f766a.png)

pixelfed.org

![image](https://user-images.githubusercontent.com/1688309/203257661-7b0a4068-4614-4a2b-84d6-7edfbcba37f1.png)

MTRNord:
Another data point: this is the pad keymaker used together with the community to do something like this in an over engineered fashion. Maybe it is some use https://scalar.vector.im/etherpad/p/!bKsmsWVPSktSNkPEqf_nordgedanken.dev

thibaultamartin:
Based on feedback from https://mamot.fr/@thibaultamartin/109432454640400772 it seems that the most important things for people when choosing a federated _microblog_ instance are:
- Expected Longevity of the instance (based on whether they have "a team of multiple people working on it, had been around for a long time, and had enough financial reserves to run for at least six more months, with a steady income stream" and whether the instance is up to date or not).
- Server location/jurisdiction
- Rules/Code of Conduct

austinhuang0131:
I have moved my homeserver list to https://servers.joinmatrix.org which now has a filter feature.

MTRNord:
[...]

PR #1662, opened 2023-02-17
FAQs: Remove hijacked link

Aquathing:
Unfortunately, anchel.nl now redirects to the website of another app which appears to be a competitor of Matrix.

Perhaps, it would make sense to replace it with [JoinMatrix's list](https://joinmatrix.org/servers/), but I figured it would be better to first fix this, and later ask permission to the website's owner to mention it in Matrix's FAQs.

### DIFF
--- a/gatsby/src/pages/faq.js
+++ b/gatsby/src/pages/faq.js
@@ -982,11 +982,7 @@ const Faq = ({ data }) => {
               <p>
                 Clients can access any <a href="#definitions">homeserver</a>—you
                 don't have to use matrix.org, though historically it is the
-                largest public homeserver. anchel.nl lists{" "}
-                <a href="https://www.anchel.nl/matrix-publiclist/">
-                  free public homeservers
-                </a>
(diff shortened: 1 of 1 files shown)


Aquathing:
Funnily enough, https://www.anchel.nl/matrix-publiclist/, does redirect to the competitor's website while https://anchel.nl/matrix-publiclist/ (or anything else as path), redirects to a list of other websites.

Aquathing:
> Whoa, thank you very much for this 😱

No worries! Glad to help 🤗

austinhuang0131:
https://matrix.org/docs/guides/introduction#how-can-i-try-it-out already uses my (JoinMatrix.org) list, so for consistency, it should be linked here too.

thibaultamartin:
Heads up we have #1570 and we are interested in building such a list once the new matrix.org is out :)

joinmatrix.org is incredibly useful and should still be promoted before we add something similar to matrix.org.
Do you want to open a PR @austinhuang0131?

PR #1667, opened 2023-02-23
Zola tech docs

thibaultamartin:


templates/shortcodes/figure.html
```diff
@@ -0,0 +1,4 @@
+<div class="figure">
```
jplatte, line 1: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure

thibaultamartin: Solved in [f0c6068](https://github.com/matrix-org/matrix.org/pull/1667/commits/f0c6068170951e42345154324fe4a235f9a0de63)

jplatte (approved):

PR #1668, opened 2023-02-24
Add chatty client.

jonsger:
With release 0.7.0 chatty got now a rather mature Matrix support. So it was decided that it should be added to try-matrix-now: https://source.puri.sm/Librem5/chatty/-/issues/768

I need to verify some details, so long I leave this PR in Draft mode.


thibaultamartin:
Hej @jonsger, this PR targets the `gatsby` branch of our website, which is now an archive of the old website. Please reopen a PR against `main` if you want to add chatty to the new website :)

PR #1672, opened 2023-02-27
Global CSS cleanup

thibaultamartin:


sass/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '_normalize';
+//@import '_normalize';
```
jplatte, line 1: Probably committed by accident?

sass/_base.scss
```diff
@@ -269,25 +198,17 @@ pre table {
...
-            margin: 60px 32px 70px;
-        }
-
-        margin: 89px 80px;
-
         h1 {
             font-style: normal;
             font-weight: 700;
-            font-size: 3.125rem;
+            font-size: 2.8rem;
```
MTRNord, line 204: I wonder if these font sizes should be variables too or not. Mostly since we have some places (I think) where we have to reset this since for example the markdown stuff overwrote it, or we want markdown generated things to be one level smaller. So having a variable for that might be nice.

Though we might also just want to just use SCSS mixins for that instead. :thinking:

sass/_base.scss
```diff
@@ -314,7 +235,7 @@ pre table {
     background-repeat: no-repeat;
     padding: 1rem;
 
-    max-width: 45%;
+    max-width: 23rem;
```
MTRNord, line 238: Is this really going to cause the same effect, or is there some other reason behind this change?

thibaultamartin: The reason is that if the card is used somewhere that is super wide, the card will be very stretched. In practice 23rem "feels" nice as a maximum width

sass/style.scss
thibaultamartin: Fixed in 6b3e6442

jplatte (approved):

PR #1674, opened 2023-02-28
Zola clients no table

thibaultamartin:
Based on #1672 and replaces #1655


thibaultamartin:
Placeholder menus look like

image

thibaultamartin:
Filter for the platforms now works

https://user-images.githubusercontent.com/1688309/222142241-d91cf344-56b7-44b8-8193-364d1a30ea5d.mov

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
+document.addEventListener('DOMContentLoaded', (event) => {
+    // Display filters only if javascript is enabled
+    let filtersBar = document.getElementById("filters");
+    filtersBar.style.display = "flex";
```
jplatte, line 4: We can do this differently to avoid flickering on page load: Just add a `<noscript>` tag in the `head` that contains a stylesheet that hides the element.

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
...
+            this.allOf = allOf;
+            this.anyOf = anyOf;
+            this.makeMenuInteractive(filterId);
+            this.enableFilters(filterId);
+            this.isAndFilter = isAndFilter;
+        }
+
+        makeMenuInteractive(filterId) {
+            let filterButton = document.getElementById(filterId)
+            filterButton.addEventListener("mouseup", (event) => {
```
jplatte, line 17: I think this should be click, not mouseup: https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
...
+                let platformFilterMenu = document.getElementById(filterId+"-menu");
+                if(platformFilterMenu.style.display !== "block") {
```
jplatte, line 19: Your JS formatter is making some questionable choices IMHO 😅

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
...
+                    platformFilterMenu.style.display = "block";
+                } else {
+                    platformFilterMenu.style.display = "none";
+                }
+            });
+        }
+
+        enableFilters(filterId) {
+            let platformFilterMenu = document.getElementById(filterId+"-menu");
+            for(var platformOption of platformFilterMenu.children) {
```
jplatte, line 29:
```suggestion
            for(const platformOption of platformFilterMenu.children) {
```

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
...
+                if(!platformOption.classList.contains("reset-links")) {
+                    let optionId = platformOption.children[0].id;
+                    let checkbox = platformOption.children[0];
```
jplatte, line 32:
```suggestion
                    const optionId = platformOption.children[0].id;
                    const checkbox = platformOption.children[0];
```

static/js/clients.js
```diff
@@ -0,0 +1,143 @@
...
+            this.refreshCardsView();
+        }
+
+        checkAllBoxes(filterId) {
+            let platformFilterMenu = document.getElementById(filterId+"-menu");
+            for(let platformOption of platformFilterMenu.children) {
+                if(!platformOption.classList.contains("reset-links")) {
+                        platformOption.children[0].checked = true;
+                        this.boxWasChecked(platformOption.children[0].id);
+                }
```
jplatte, line 96: Are you using a formatter at all, actually? 😄

jplatte (commented):
Overall I would say the JS is pretty readable :)

static/js/clients.js
thibaultamartin: No, any recommendation :p ?

jplatte: The one you get by default in VSCode is decent, no idea if that's a thing you can set up independently though.

static/js/clients.js
thibaultamartin: Fixed in 05ac9513

static/js/clients.js
thibaultamartin: Fixed in 05ac9513

templates/ecosystem/clients.html
```diff
@@ -1,5 +1,9 @@
 {% extends "page.html" %}
 {% import "macros/clients.html" as clients %}
+{% block head_extra %}
+<noscript><link rel="stylesheet" href="/no-js.css" /></noscript>
```
jplatte, line 4: Seems like you forgot to commit that file.

thibaultamartin: I can see it in https://github.com/matrix-org/matrix.org/pull/1674/files#diff-8a85088eeb84225e3dc5b2229da18b417c810bc1ddb954f19e6acf201f5575a6, did I miss anything?

jplatte: Oh, I was searching for `no-js.css` but it's `no-js.scss`. Nevermind.

thibaultamartin: FWIW I used `.scss` because I think Zola generates a minified css from the scss, but I didn't test if zola minifies a "raw" css file.

I wonder if we will need something to minify the js as well eventually.

jplatte: I think unless we write a lot more JS, gzip should be sufficient. Of course if Zola supports it out of the box it doesn't hurt to use that, but I don't think it does.

thibaultamartin: From https://github.com/getzola/zola/issues/542#issuecomment-657169503 I'm not sure it does. But based on https://github.com/getzola/zola/issues/542#issuecomment-521409104, if it does it won't touch anything under `static`, so we need to move the js out of it. Maybe for a separate PR?

jplatte (approved):
Could use a little cleanup, but looks good overall.

static/js/clients.js
```diff
@@ -0,0 +1,141 @@
...
+                    }
+
+                    if (containsAllOf && containsAnyOf) {
+                        client.style.display = "flex";
+                    } else {
+                        client.style.display = "none";
+                    }
+                }
+            }
+        };
```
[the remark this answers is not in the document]
thibaultamartin: Fixed in 5de41466

sass/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '_normalize';
+//@import '_normalize';
```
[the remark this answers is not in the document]
thibaultamartin: Fixed in 535d7654, whoops

templates/macros/clients.html
```diff
@@ -16,4 +16,82 @@
...
+{% endif %}
+{% if client.features.sso == true %}
+feature-sso
+{% endif %}
+{% if client.features.multi_account == true %}
+feature-multi-account
+{% endif %}
+{% if client.features.multi_language == true %}
+feature-multi-language
+{% endif %}
```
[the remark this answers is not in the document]
thibaultamartin: Tested, it works :)
Fixed in b7dbb12

templates/macros/clients.html
```diff
@@ -16,4 +16,82 @@
...
+{% if client.maturity == "Alpha" %}
+maturity-alpha
+{% endif %}
+{% if client.maturity == "Beta" %}
+maturity-beta
+{% endif %}
+{% if client.maturity == "Stable" %}
+maturity-stable
+{% endif %}
```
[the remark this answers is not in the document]
thibaultamartin: Whoa super clever, fixed in d18ce98c

PR #1679, opened 2023-03-04
Add a popup to display client details

thibaultamartin:


thibaultamartin:
Currently looks like

image

thibaultamartin:
Client cards are clickable and make the details pop-up appear (without js!)

The details layout can still be a bit weird, in particular when a client supports all platforms or when it has no unsupported features (or both).

image

image

thibaultamartin:
Result so far 
https://user-images.githubusercontent.com/1688309/225060184-95afc87e-228c-4f5d-92a1-ca354cc2343a.mov

sass/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '_normalize';
+//@import '_normalize';
```
jplatte, line 1: Again? 😂

static/assets/stores/f-droid-upstream.svg
jplatte, line 1: Is this file included intentionally?

static/assets/stores/f-droid.svg
jplatte, line 1: It's a bit weird this has such a distinctly different font from the upstream one. But maybe not that relevant 🤷🏼

jplatte (approved):
Code looks clean enough, left minor comments.

static/assets/stores/f-droid-upstream.svg
thibaultamartin: Fixed in [dbb45c8](https://github.com/matrix-org/matrix.org/pull/1679/commits/dbb45c8615bcbf25cfc777966fb735d70fbfb912)

sass/style.scss
thibaultamartin: [dbb45c8](https://github.com/matrix-org/matrix.org/pull/1679/commits/dbb45c8615bcbf25cfc777966fb735d70fbfb912)

static/assets/stores/f-droid.svg
jplatte: I don't know if people will notice. Probably not too important.

PR #1685, opened 2023-03-14
Zola clients obvious filters

thibaultamartin:
Based on #1679


thibaultamartin:
Here's how it looks

https://user-images.githubusercontent.com/1688309/225104458-50c3511d-7c3c-4a2a-aa9e-a70ea6549bcc.mov

static/js/clients.js
```diff
@@ -1,41 +1,71 @@
...
+
             filterButton.addEventListener("click", (event) => {
-                let filterMenu = document.getElementById(filterId + "-menu");
                 if (filterMenu.style.display !== "block") {
                     filterMenu.style.display = "block";
+                    filterOverlay.style.display = "block";
+                    filterButton.style.zIndex = 150;
                 } else {
                     filterMenu.style.display = "none";
+                    filterOverlay.style.display = "none";
```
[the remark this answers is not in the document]
thibaultamartin: Solved in [edc69f5](https://github.com/matrix-org/matrix.org/pull/1685/commits/edc69f55f0f879176fd1a05263cbc110528f44ac), but happy to have your eye on it

sass/_clients.scss
```diff
@@ -193,6 +202,10 @@ body:has(.clients-card-deck > div > .client-checkbox:checked) {
                     user-select: none;
                 }
             }
+
+            &.display {
+                display: block;
+            }
```
jplatte, line 208: Instead of a default `display: none` and override, I personally prefer `&:not(.display)` style rules. In this case you wouldn't have to set the `display` property for the visible case then, because `block` is default (for most things).

thibaultamartin: Fixed in 3d8a1632

PR #1689, opened 2023-03-21
Zola servers

thibaultamartin:


thibaultamartin:
Currently looks like this

https://user-images.githubusercontent.com/1688309/226855978-23614886-5cd3-43c5-82f8-56c8c15b3a3f.mov

I'm not sure what causes the weird layout when servers are filtered out. Any clue @jplatte?

thibaultamartin:
Currently looks like
https://user-images.githubusercontent.com/1688309/226947227-65cbe867-7b4e-4a9b-83db-5046513931b3.mov

I'm not too happy with:
- needing to deselect everything if I want to look for a server in a specific language, though I couldn't find a better UX for it
- the UI that jumps when there's nothing to display
- lack of placeholder when nothing matches (it feels like something silently crashed)

Half-Shot:
Should obsolete be visible by default?

PR #1694, opened 2023-03-28
Zola bridges

thibaultamartin:
To merge after #1689


thibaultamartin:
Screenshot 2023-03-28 at 16-43-33 🛠️ Matrix org - Bridges

thibaultamartin:
Implementations page looks a tad empty with real life data, and some information is missing. We could fill the void with buttons for:

- The bridge documentation
- The Matrix room they can ask support in
- The code repository

It would also be nice to display the licence (licence can be any of the values in the "Identifier" column of https://spdx.org/licenses/) and maturity (Stable, Beta, Alpha, Obsolete) too.

image

PR #1699, opened 2023-04-06
[FIX] Nonfunctional menu button in narrow view #1508

AbhiJain2196:
1. Fixed special characters like matrix's to matrix\\'s, doesn't to doesn\\'t and "%3F" to "" in `h4` and `a` tag.
2. Fixed warning of `class` to `className` in Footer component.


richvdh:
@thibaultamartin my understanding is that we are not currently accepting PRs against master - is that correct?

thibaultamartin:
Hi all, just getting back from holidays. Indeed, we're currently working on a complete rewrite of the matrix.org website, and we don't accept changes against the `master` branch.

I will update the README accordingly, and set-up a tag system so it's more obvious where we accept help :)

PR #1701, opened 2023-04-14
Zola responsive sdks

thibaultamartin:


thibaultamartin:
More readable than the table we had previously.

It looks like the cards themselves could sport information about the type of project the SDK can be used for (bot, chat client, bridge) and a short description for each.

image

thibaultamartin:
Already looking better with descriptions.

image

PR #1703, opened 2023-04-15
Add a workflow for zola-check

MTRNord:
The workflow executes zola check and generates check results and file annotations from it.
It also mentions webarchive links if one is available.

This doesn't work on PRs yet, since the workflow doesn't diff results for PRs yet. Also, it currently takes about 3h for this repo currently.

More info on the workflow is here: https://github.com/MTRNord/zola-check-manager/
Example output is: https://github.com/MTRNord/zola-check-manager/runs/12772905612

### DIFF
--- a/.github/workflows/link-checker.yaml
+++ b/.github/workflows/link-checker.yaml
@@ -2,9 +2,6 @@ name: Check for broken links every week
 on:
   schedule:
     - cron: "0 12 * * 0"
-  pull_request:
-    branches:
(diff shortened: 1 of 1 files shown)


MTRNord:
Since I guess the checks permission doesnt work when I set it (Makes sense and is good that it doesn't on PRs tbh) here is what it _should_ look like if it would have properly worked: https://github.com/MTRNord/zola-check-manager/runs/12772905612

PR #1705, opened 2023-04-18
Make the docs responsive

thibaultamartin:


thibaultamartin:
This is the button while it was still a `div`, not relying on the checkbox trick. This is how it should look like eventually.

Screenshot 2023-04-18 at 14 48 08

thibaultamartin:
This is what I have using the checkbox trick. I'm not entirely sure why the button isn't positioned `2rem` from the bottom of the `docs-body` container.

Screenshot 2023-04-18 at 15 43 21

PR #1713, opened 2023-04-26
Individual clients view

thibaultamartin:
This makes both the clients grid and the clients details much more accessible, without having to rely on checkbox tricks.


thibaultamartin:
At the moment the details page doesn't look as good as it used to when in a pop up (which wasn't exactly terrific either) and is quite hard to read

image

It might be worth taking inspiration from the bridges details page (which needs some CSS tweaking too apparently) to make it easier to read

image

thibaultamartin:
Desktop view 

image

Mobile view

image

PR #1741, opened 2023-04-28
[Zola] Remove width and height attributes in youtube embeds to make sure these are responsive

MTRNord:
Tries to somewhat normalize iframes (not fully as you see) to make them responsive by removing width and height.

Also imports a few posts which got lost on the migrator due to a bug with the frontmatter in the gatsby original.


thibaultamartin:
There are path collisions, are you sure the posts were actually lost?

MTRNord:
> There are path collisions, are you sure the posts were actually lost?

< insert facepal emoji >. No :D I just didnt check hard enough apparently. Sorry. Will fix

PR #1762, opened 2023-05-02
Allow tabbing to ecosystem links in the header

jplatte:


thibaultamartin:
https://user-images.githubusercontent.com/1688309/235621235-1f588957-14ff-4288-83e5-2e0889ef7749.mov

Tab navigation shows the menu, but it doesn't seem to select the label. I don't know if this gets in the way of assistive technologies

thibaultamartin:
Apparently voiceover is not too happy with the hidden checkbox :|

https://user-images.githubusercontent.com/1688309/235623045-3d9ae776-a69d-4e33-90a5-d46f038292d2.mp4

jplatte:
Seems like a case of needing an `aria-label` or whatever it's called?

thibaultamartin:
Just gave it a go and it seems that's still confusing voiceover in the same way

jplatte:
What did you try exactly? I'm sure there's some (combination of) aria labels that fixes this. It's just about presenting a checkbox as a menu toggle, doesn't seem super crazy..

thibaultamartin:
Hm your branch untouched does seem to work with Safari… I wonder if that's a Firefox bug?

MTRNord:
> Hm your branch untouched does seem to work with Safari… I wonder if that's a Firefox bug?

I wonder if this should be reported via https://webcompat.com/ or bugzilla 🤔

PR #1784, opened 2023-05-09
Make sure the footer sticks to the bottom

jplatte:
If the content doesn't otherwise fill the viewport, make it expand so the footer is at the very bottom of the viewport.

| Before | After |
| ------ | ----- |
| Screenshot 2023-05-09 at 12-37-13 Matrix org - Podcasts | Screenshot 2023-05-09 at 12-38-14 🛠️ Matrix org - Podcasts |

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -23,6 +23,8 @@ html {
 body {
     margin: 0;
     min-height: 100svh;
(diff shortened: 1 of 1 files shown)


thibaultamartin:
Merci !

Issue #1778, opened 2023-05-05
Bullet points misaligned in providers cards


ifeeltiredboss:
1. Go to https://matrix.turbo.fish/ecosystem/hosting/
2. IMHO the bullet points should be aligned (top left)
<img width="882" alt="image" src="https://user-images.githubusercontent.com/111905360/236408367-9758d7bc-5d15-443e-89b9-edf0b72e8562.png">

PR #1790, opened 2023-05-10
Issue#1778

kartik126:
Fixed Bullet points misaligned in providers cards #1778. I have moved provider card description i.e

{{ provider.description | markdown | safe }}

inside details class div so that space-around property shouldn't affect the description element. And the elements p and ul inside details class div i gave text align left.

### DIFF
--- a/sass/_hosting.scss
+++ b/sass/_hosting.scss
@@ -33,8 +33,7 @@
                 max-width: 300px;
             }
 
-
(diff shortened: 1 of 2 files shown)


thibaultamartin:
Duplicate of #1787

thibaultamartin:
Egh, sorry that's not the one I meant to close as duplicate.

PR #1796, opened 2023-05-11
Make filters accessible via keyboard navigation

### LINKED ISSUE #1788: Filters are not accessible
https://github.com/matrix-org/matrix.org/assets/1688309/b7caadb7-f302-4f55-93f5-0375d5781764

- [ ] Turn the filters into `button`s
- [ ] Enter toggles the open/closed state of the menu
- [ ] Esc closes the menu

thibaultamartin:
Fixes #1788

### DIFF
--- a/sass/_projects.scss
+++ b/sass/_projects.scss
@@ -93,7 +93,7 @@
             gap: .8rem;
             z-index: 100;
 
-            padding: .2rem .8rem;
+            padding: .5rem .8rem;
 
             background-color: #fafafa;
             color: #000;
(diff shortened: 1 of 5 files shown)


thibaultamartin:
https://github.com/matrix-org/matrix.org/assets/1688309/9e5bc6e5-130f-4da1-8f5a-ad3b0dfb3650

Navigation is not super intuitive:

- It looks like when the focus goes outside of the filter menu, the filter menu should collapse and the overlay should disappear
- Pressing the escape key doesn't escape the menu

PR #1802, opened 2023-05-16
Don't show obsolete clients as default

kartik126:
Obsolete clients as default is not visible .I have removed checked attribute from obsolete filter option.


kartik126:
You're welcome !

PR #1809, opened 2023-05-24
Set @matrix-org/security as CODEOWNERS for security blog posts

dkasak:
After seeing it work with synapse-core, it seems like a nice addition, especially given the upcoming RACI matrix for the security process.

Alternatively we could also have it include `*vulnerabilit*` and/or `*advisor*`, but checking historical blog posts, most of the security-related ones have `security` in their title.

Preview: https://pr1809--matrix-org-previews.netlify.app


davidegirardi (approved):

MTRNord:
Might make sense to also do this for the zola branch so it is kept around for the new blog too. (though I think the new blog branch doesnt yet have a codeowner file)

MTRNord:
> @MTRNord I can do that, but hopefully the Zola team is keeping track of additions to the main line instead of just blindly trampling over stuff. ;P

We try! :D Its just a cat and mouse game :) But its totally a goal to sync blog posts and repo tooling afaik :)

PR #1814, opened 2023-05-30
Recreate the support page

thibaultamartin:


thibaultamartin:
I can't seem to get that pseudo-element overflowing to be visible ._.

image

PR #1824, opened 2023-06-13
Reduce index size

laurencegill:
Reduce the `search_index.en.js` file, it is 36.8 MB in size and takes >8s to build. This is too big to deploy on cloudflare pages.

This change reduces it to 2.7M (arguably still too big) and builds in <2s

We should look at the [search](https://www.getzola.org/documentation/getting-started/configuration/#:~:text=paths_keep_dates%20%3D%20false-,%5Bsearch%5D,-%23%20Whether%20to%20include) configuration in greater detail. Using a value of `100` and/or excluding the content would be ideal

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -5,7 +5,7 @@ base_url = "https://matrix.turbo.fish"
 compile_sass = true
 
 # Whether to build a search index to be used later on by a JavaScript library
-build_search_index = true
+build_search_index = false
 
 default_language = "en"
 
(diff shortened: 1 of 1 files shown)


laurencegill:
> How is GitHub rendering a proper diff if this branch isn't even based on main? Is the PR actually against `zola` and GitHub is just glitching for me? Anyways yeah, I'm pretty sure we don't need the search index right now.

`master` is still the default branch. `main` and `zola*` are being used for the zola deployments to cloudflare pages

I've disabled the search index entirely, thanks :smile:

jplatte:
Ahaa, I had missed that the default branch is master, but there's also main.

PR #1826, opened 2023-06-13
Zola base url

laurencegill:
Update the `base_url` as this redirects parts of the site to elsewhere which makes testing tricky.

For now, the `CF_PAGES_URL` is used to override this, and before deployment we'll set that only for non `main` deployments; those will use this setting (matrix.org)

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,5 +1,5 @@
 # The URL the site will be built for
-base_url = "matrix.org"
+base_url = "https://matrix.org"
 
 # Whether to automatically compile all Sass files in the sass directory
 compile_sass = true
--- a/config.toml
+++ b/config.toml
@@ -1,5 +1,5 @@
 # The URL the site will be built for
-base_url = "matrix-website.pages.dev"
+base_url = "matrix.org"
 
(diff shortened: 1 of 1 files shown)


thibaultamartin:
@laurencegill is there a preview corresponding to this value to test it? The docs make it look like it should be a domain, but in practice it seemed only to work with an URL, like the variable name indicates

laurencegill:
> @laurencegill is there a preview corresponding to this value to test it? The docs make it look like it should be a domain, but in practice it seemed only to work with an URL, like the variable name indicates

Good point, I thought I had it working, but it was late :grin:  I'll double check...

laurencegill:
@thibaultamartin https://zola-base-url.matrix-website.pages.dev/ is the preview

It wasn't using the updated build command where the `base_url` is set to the `CF_PAGES_URL` as I hadn't updated the cloudflare side of things when it initially built.

I'm going to implement a build script that doesn't do this for the main deployment do it'll always default to `matrix.org`, but I'll update it to something we know works in the mean time.

thibaultamartin (approved):

PR #1828, opened 2023-06-14
Update Integrations page description

HarHarLinks:
The integrations description sounded like only non-chat bridges are integrations which I felt wasn't accurate, so here is my proposal.


thibaultamartin:
This might be a bit long for the hero section of the page, do you think you can shorten it a bit without losing too much of its meaning?

PR #1830, opened 2023-06-15
Zola headers and redirects

laurencegill:
Converts the `header set` and `rewriterule` configuration from our apache config to cloudflares `_headers` and `_redirects` files :crossed_fingers:

### DIFF
--- a/static/_headers
+++ b/static/_headers
@@ -1,3 +1,18 @@
 # Remove the ACAO header which is added by default on Cloudflare Pages
+# Enables XSS filtering
+# Avoid MIME type sniffing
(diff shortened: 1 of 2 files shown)


thibaultamartin:
For the sake of disambiguation, those files are consumed by CF pages and are not relevant to Zola, right?

laurencegill:
> For the sake of disambiguation, those files are consumed by CF pages and are not relevant to Zola, right?

Yes these are presumably performed on whatever webserver cloudflare uses, before the request makes it to the zola site.

thibaultamartin (approved):

Issue #1845, opened 2023-06-16
Client pages can be improved


axelsimon:
A few things on the client-specific pages (such as https://matrix.org/ecosystem/clients/fluffychat/) can be improved:
- "last release" --> "latest release". Last is ambiguous
- E2ee, spaces, Voip 1to1, etc.:
  - As a newcomer, what do these mean, what are they? Am i expected to just know? It would be useful to introduce these as "Matrix feature support:"
  - Each of these would benefit from an explanation: what does E2EE mean? what is the difference between Voip 1to1 and Voip jitsi? etc. Ideally, we should describe the usefulness rather than give the technical feature: "private conversations" vs E2EE. Another option is to have each feature either link to a page explaining the feature in more detail, or a pop-up / bubble on hover giving further details as to what the feature means
  - Initialisms, if used, should be properly capitalised: E2EE, SSO, VoIP
  - Multi language: what does this mean? the app is translated into multiple languages? It supports multiple language inputs? If it's that it's translated into multiple languages, it may be better to say "translated" or "localised", ideally followed by the number of languages
- The area between "support room" "source code" and the "download on…" buttons looks a bit bare. I'd kind of expect a longer description of the client to go there.

Lastly, on the Fluffychat page, the "Use in your browser" button should lead to https://fluffychat.im/web/ and not just https://fluffychat.im/

anoadragon453:
I'd add to this list that I would prefer "VoIP 1to1"  to instead be "VoIP 1-to-1".

PR #1882, opened 2023-06-21
fixed issue #1845

harshlancer:
This is temporary fix however we can make it permanent with responsiveness of images


thibaultamartin:
Hej harshlancer, thanks for the contribution. Can you expand on why you think this PR fixes #1845?

PR #1892, opened 2023-06-26
Fix synapse broken links

thibaultamartin:
Fixes https://github.com/matrix-org/synapse/issues/15829
Fxies https://github.com/matrix-org/synapse/issues/15825


laurencegill (approved):
Good spot, I missed that one

PR #1900, opened 2023-06-30
Add Cosium/matrix-communication-client to the ecosystem page

reda-alaoui:
This change adds https://github.com/Cosium/matrix-communication-client to the ecosystem page.

### DIFF
--- a/content/ecosystem/sdks/sdks.toml
+++ b/content/ecosystem/sdks/sdks.toml
@@ -569,3 +569,16 @@ featured_in = []
 description = """
 Matrix client library for MicroPython.
 """
+
+[[sdks]]
+name = "matrix-communication-client"
+maintainer = "Réda Housni Alaoui"
+maturity = "Stable"
+language = "Java"
+licence = "MIT"
+repository = "https://github.com/Cosium/matrix-communication-client"
+purpose = ["client", "bot"]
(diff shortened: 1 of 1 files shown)


thibaultamartin:
The maturity for this project is stable, but the first commit seems very recent. This might be setting wrong expectations for people. Are you really sure you want that project to be listed as stable?

reda-alaoui:
Hello @thibaultamartin ,

The project is stable in the sense that I will make sure to not break any public api.
So I don't think it sets the wrong expectations for people. We use it in production for example.

But the project is not feature complete regarding the client-server specification.

Given these informations, do you still think it should be referenced as unstable?

reda-alaoui:
Hello 👋

thibaultamartin:
Ah I had missed your question! I'm okay to add it as stable, though I would recommend to list it as beta if you plan to get significantly closer to supporting the whole C-S API.

Let me know which one you're the most comfortable with, and I'll merge it

reda-alaoui:
> though I would recommend to list it as beta if you plan to get significantly closer to supporting the whole C-S API.

I think it will depend on community contribution and on my own needs.
Whatever the case, following semantic versioning should be enough to increase feature support while ensuring stability.

Therefore, please add it as stable :)

PR #1903, opened 2023-07-03
Display the full content of the article in the RSS feed

### LINKED ISSUE #1868: Rss feed not showing full article when more comment is added
It seems like the default template does use the short version instead of full article if available.

https://www.getzola.org/documentation/templates/feeds/

Suggests we can use a custom template to fix this.

thibaultamartin:
Fixes #1868 by overriding `atom.xml` and `rss.xml` default templates (default templates are in https://github.com/getzola/zola/tree/master/components/templates/src/builtins)

### DIFF
--- a/templates/rss.xml
+++ b/templates/rss.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
+    <channel>
+      <title>{{ config.title }}
+        {%- if term %} - {{ term.name }}
+        {%- elif section.title %} - {{ section.title }}
+        {%- endif -%}
+      </title>
+      <link>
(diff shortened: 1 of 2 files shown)


thibaultamartin:
Tulir suggests to limit the number of items in the feed to the ~20 most recent items.

I'm not sure it's a common practice. Off the top of my head RSS feeds are mostly used to get the latest news, not to browse the whole website, so it can make sense to limit the number of entries returned

templates/rss.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MTRNord, line 1: I _think_ we currently dont use rss.xml actually since we use the atom feed.

thibaultamartin:
I'm merging this fix already to remove one problem, and I opened #1918 to keep track of limiting the number of items

PR #1905, opened 2023-07-03
Fix Signal bridge documentation link

sarayourfriend:
Noticed this while browsing the documentation today on https://matrix.org/ecosystem/bridges/signal/. The documentation link should go to https://docs.mau.fi/bridges/python/setup.html?bridge=signal but goes to https://docs.mau.fi/bridges/python/setup.html?bridge=telegram (the telegram bridge docs) instead.


sarayourfriend:
Drafting because I've noticed a couple other documentation links that need updating as well and will include fixes for those here too.

PR #1914, opened 2023-07-04
tidy up language

neilisfragile:
To make the distinction between Matrix and the matrix.org homeserver clear


progval:
I don't understand this pull request. The matrix.org homeserver has no particular involvement with this bridge; saying "Matrix" was more correct.

PR #1922, opened 2023-07-07
[Clients / NeoChat] Fix flathub.app_id

Andanan:
The Flathub-Link on the Client-Page for NeoChat is currently broken because the app id is wrong.
This change fixes this, so that the Link on the Client-Page does not result in a 404.

Currently: [org.kde.NeoChat](https://flathub.org/apps/org.kde.NeoChat)
Fixed: [org.kde.neochat](https://flathub.org/apps/org.kde.neochat)


thibaultamartin:
Of nice catch! Thanks for the PR :)

PR #1930, opened 2023-07-07
Fix Author in Feed

MTRNord:
The authors array is in the taxonomies, not on the page itself.


MTRNord:
Actually closing as there is no technical reason that the current one fails. This needs deeper investigation

PR #1944, opened 2023-07-18
Add a New Post Checklist to PRs

MTRNord:
This adds a Pull request template to this repo for ensuring the posts don't have common issues that have been noticed over the last few TWIM editions. I am not trying to annoy editors here, but instead help them.

Zola uses CommonMark, while the old Blog used MDX. While both are markdown there is a small difference in some of the syntax especially around autolinking of links. This hopefully helps authors to double-check these things while they develop a muscle memory for the changed syntax.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,10 @@
+<!-- Thanks for submitting a PR! Please ensure the following requirements are met when writing a new post -->
+
+## New Post Checklist
+
+-   [ ] Links are linkified
+  -   [ ] Bare links are surrounded by `<>` (required since Zola follows commonmark strictly)
+  -   [ ] Links with names are of the pattern `[Name](URL)`
+-   [ ] The frontmatter is valid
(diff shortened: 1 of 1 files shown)


HarHarLinks:
there's probably a github action out there that can scan content for things that look like URLs and warn if they're not markdown...

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+<!-- Thanks for submitting a PR! Please ensure the following requirements are met when writing a new post -->
+
+## New Post Checklist
+
+- [ ] Links are linkified
+  - [ ] Bare links are surrounded by `<>` (required since Zola follows commonmark strictly)
```
jplatte, line 6: More like because `pulldown_cmark` doesn't support auto-detection of bare links: https://github.com/raphlinus/pulldown-cmark/issues/695

MTRNord: Oh! Well that's unfortunate. Reading the spec I thought this is actually intentional by the commonmark spec 😅

PR #1946, opened 2023-07-18
Fix: Unify author names which were unusual or duplicated

MTRNord:
There were some author names duplicated or merged into a single string. This commit fixes that.

image

Based on the overview at https://matrix.org/author/


uhoreg:
@lampholder would you prefer to be called "Tom Lant" or "Thomas Lant" when you're marked as an author in a blog post?

PR #1951, opened 2023-07-21
Fix some bare links

MTRNord:
Zola needs those <> around links to work


uhoreg (approved):
Argh!  I thought I caught all of those.

MTRNord:
Luckily redeployments are quick now :)

Issue #1853, opened 2023-06-16
Try Matrix page improvements


mpeter50:
On the [Try Matrix page](https://matrix.org/try-matrix/) the "Choose a server" section starts with recommending to use the server of Matrix.org, but then it continues with only a button to install one.
Although the page describing the server software options notes that the user does not need to run their own server, this is kind of a rollercoaster. Before clicking the button they might think that they do need to install one, and at that point you might lose the user's interest in continuing.

In the section itself although it is mentioned that Matrix.org hosts a public server, it is not clear (to the new user) which one is it, partly because of the capitalized M. At this point the server address might not be important though, but later when they are walked through registration and login.
Then, it is recommended to select an existing server, but they are not told about the options, or how they could find them. Also, the word "existing" in that sentence implies as if the server of Matrix.org would not exist yet.
It is also not obvious where can the user actually decide on which server to use. You could mention (possibly on the server list page) that this will happen on the registration form.

As a solution,
- you could add a new button besides "Install your own" that leads the reader to a new page that lists several public servers,
- and also somehow clarify near the "Install your own" button that going through that is entirely optional. Maybe even with a different button style or something like that.

There should also be an addition to the [Matrix for IM page](https://matrix.org/docs/chat_basics/matrix-for-im/)
- mentioning where can a user set their homeserver (also marking it on the screenshot),
- and highlighting that by default matrix.org is selected, so if thats fine to them, they dont have to deal with that,
- but again linking to the server list page somewhere below the screenshot, in case they may choose one now that they know where and how to do so

On the server page you may consider ordering the servers by some kind of score, that measures the quality of configuration of that server. For inspiration, you may check out [how Cryptpad has done it](https://cryptpad.org/instances/). They describe their process [here](https://docs.cryptpad.fr/en/admin_guide/admin_panel.html#list-my-instance-in-public-directories).

> In order to be listed here these services have passed a series of checks to ensure they use an up-to-date version of CryptPad and have the recommended security and privacy settings. You can check their realtime availability with our dedicated status page. Requests for support should be directed to their administrators.

PR #1952, opened 2023-07-22
link joinmatrix.org for "find an existing server"

austinhuang0131:
as a temporary measure to #1570

see also #1853


thibaultamartin:
One mangled and demangled branch later, I could push my changes to add a CTA pointing to joinmatrix.org

It looks like

image

and

Screenshot 2023-08-01 at 11 51 05

Are you happy with those additional changes @austinhuang0131?

austinhuang0131:
sure

Issue #1955, opened 2023-07-26
Blog category URLs changed


clokep:
See matrix-org/synapse#16005, it seems that the blog categories used to be available at things like https://matrix.org/blog/category/releases, but are now at https://matrix.org/category/releases (note the missing `/blog`).

thibaultamartin:
Relevant context in the advocacy internal channel: https://matrix.to/#/!wIPEfqhDsTwAzqFGvi:matrix.org/$AZoQ49bhS3ud0j0ToTdRnKWHQ1tY_xBdwd5erqOf7ZQ?via=matrix.org&via=element.io&via=t2l.io

clokep:
Thanks for fixing this so fast. 👍

PR #1958, opened 2023-07-27
Redirect for categories

laurencegill:
Fixes https://github.com/matrix-org/matrix.org/issues/1955


thibaultamartin:
Does it work if urls have a trailing slash too?

laurencegill:
> Does it work if urls have a trailing slash too?

It does not :smile: but I can add it?

thibaultamartin:
That would be useful yes, please :)

thibaultamartin (approved):
LGTM, thanks Laurence

PR #1997, opened 2023-08-31
Add telodendria to server list

array-in-a-matrix:


thibaultamartin:
@jordanbancino are you happy to get telodendria added to m.org as an alpha server implementation?

Is it ok to link to the GitHub mirror, or should we link to something else?

jordanbancino:
@thibaultamartin Sure, that's fine with me. It's probably good to get Telodendria some more exposure because I am a full-time student and I work part-time as well, so I don't really have time to write much code these days. If I could get some help, that'd be awesome.

Linking to the GitHub mirror is okay, the official repository doesn't really have a "link" as it is just a CVS module available over anonymous SSH. I haven't set up a CVS web viewer yet, but if I ever do that, we can always update the repository link.

One thing I do want to note is that the room should probably be `#telodendria-general:bancino.net` instead of `#telodendria-patches:bancino.net`. The patches room is specifically for receiving patches from developers and discussing them.

thibaultamartin (approved):

PR #1999, opened 2023-09-01
Limit the RSS feed to 15 entries

### LINKED ISSUE #1918: Limit the number of items returned in the atom feed
Some websites seem to limit it to 10-20 items. At the moment we return all the articles, and that's quite a lot.

I need to assess whether it's a common/best practice, and it's then simple to implement in the atom feed template.

thibaultamartin:
Fixes #1918

### DIFF
--- a/templates/atom.xml
+++ b/templates/atom.xml
@@ -22,7 +22,7 @@
     <id>{{ feed_url | safe }}</id>
     {% set pageslen = pages | length %}
     {% if pageslen > 15 %}
-    {% for i in range(end=15) %}
+    {% for i in range(end=1) %}
     {{ atom::entry(page=pages[i]) }}
     {% endfor %}
     {% else %}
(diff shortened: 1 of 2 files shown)


templates/atom.xml
```diff
@@ -19,25 +20,14 @@
     <generator uri="https://www.getzola.org/">Zola</generator>
     <updated>{{ last_updated | date(format="%+") }}</updated>
     <id>{{ feed_url | safe }}</id>
+    {% set pageslen = pages | length %}
+    {% if pageslen > 15 %}
```
jplatte, line 24: Is this necessary? Have you tested what `end=15` does if there are fewer (or exactly) 15 pages?

thibaultamartin: It seems needed, here's the output of `zola serve` without it

```
Error: Failed to serve the site
Error: Failed to render 'atom.xml'
Error: Reason: Variable `pages[i]` not found in context while rendering 'atom.xml': the evaluated version was `pages.1`. Maybe the index is out of bounds?
```

jplatte (approved):

PR #2003, opened 2023-09-04
Unearth wp-content to unbreak many images

thibaultamartin:


aaronraimist:
@thibaultamartin you probably shouldn't fix that typo since it breaks the PGP signature

PR #2009, opened 2023-09-11
Add anchor links

### LINKED ISSUE #1760: Anchors don't appear for titles in the blog posts
Zola supports anchor insertion: https://www.getzola.org/documentation/content/linking/#anchor-insertion

JacksonChen666:
See https://www.getzola.org/documentation/content/section/#front-matter after the configs, it writes:

"Keep in mind that any configuration options apply only to the direct pages, not to the subsections' pages."

So it must be applied to all _index.md for it to be everywhere.

I have not applied the same thing for folders without a `_index.md` file, and anchor links doesn't seem to be applied to `<h1>` elements (maybe within `<header>`). It's possible I may have missed something.

fixes #1760

### DIFF
--- a/content/blog/2014/09/_index.md
+++ b/content/blog/2014/09/_index.md
@@ -1,4 +1,5 @@
 +++
 transparent = true
 render = false
+insert_anchor_links = "left"
 +++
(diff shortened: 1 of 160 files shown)


thibaultamartin:
Thanks for the PR and sorry for getting back to you this late. I'm a bit wary of adding a change to so many files at once, something feels wrong here.

It looks like Zola doesn't support another way to doing it, but I wonder if that's something that should be brought and fixed upstream 🤔

PR #2022, opened 2023-09-21
Expand feed link to all pages.

kevincox:
Previously the link was only present on the index, but this meant that it was easy to miss that the feed existed when viewing other pages, notably the blog pages.


thibaultamartin:
I like the idea, though there might be a catch! If someone is one the page of a section the blog (e.g. https://matrix.org/category/security/) they will expect to get a RSS feed for that category (in our example: https://matrix.org/category/security/atom.xml)

kevincox:
Multiple feed lists can be provided. Maybe we can update this title to "All Posts" and then the category links can be called "Security Posts" or similar.

PR #2028, opened 2023-09-29
Make the E2EE guide non-legacy

thibaultamartin:


content/docs/matrix-concepts/end-to-end-encryption/_index.md
```diff
@@ -1,6 +1,16 @@
...
+weight = 900
+template = "docs/with_menu.html"
+aliases = ["/docs/guides/end-to-end-encryption-implementation-guide", "/docs/legacy/e2e-implementation/"]
+
+[extra]
+updated = "2023-02-08T08:00:00Z"
+meta_description = """
+This guide is intended for authors of Matrix clients who wish to add support for
+end-to-end encryption. It is highly recommended that readers be familiar with
+the Matrix protocol and the use of access tokens before proceeding.
```
uhoreg, line 12: I wonder why the original text called out access tokens specifically. :shrug:

uhoreg (approved):
While we're here, we could also update the recommendation to use libolm to recommend vodozemac instead.  If we want to, possible text to replace the "The libolm library" section would be:

> ### Olm/Megolm implementations
>
> End-to-end encryption in Matrix is based on the Olm and Megolm cryptographic ratchets. The recommended starting point for any client authors is with the [vodozemac](https://github.com/matrix-org/vodozemac/), which contains implementations of all the cryptographic methods required.  [libolm](https://gitlab.matrix.org/matrix-org/olm) is also available.

PR #2041, opened 2023-10-10
Update matrix-commander.mdx

8go:
matrix-commander also supports MacOS and Windows, but on matrix.org webpage only Linux is listed. Why? How can I add MacOS and Windows to the platform list?

### DIFF
--- a/gatsby/content/projects/clients/matrix-commander.mdx
+++ b/gatsby/content/projects/clients/matrix-commander.mdx
@@ -17,7 +17,7 @@ screenshot:
 featured: true
 platforms:
     - Linux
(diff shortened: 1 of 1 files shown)


thibaultamartin:
Hej @8go, you have a good point: right now it's not possible to list TUI on anything other than Linux. I'll make it possible to add it.

Note that you will need to target the `main` branch for your changes, not `gatsby` which is the branch containing the old website!

8go:
Thank you for explaining this to me.

I will close this PR (on the wrong branch).

And I have opened a new PR targeting the `main` branch. See PR  #2051

PR #2046, opened 2023-10-13
Display the clients homepage in client details

### LINKED ISSUE #1970: please link to the homepages of featured apps!
this would be more helpful than linking just to the github and support room.

thibaultamartin:
Fixes #1970 but for all clients, not just featured apps

We have the data anyway. We just need an actual icon for, since in this PR I only recycled the docs icon

image


MTRNord (changes requested):
Looks good, but having 2 times the same icon seems potentially confusing to me

thibaultamartin:
Should be much better now 

image

MTRNord (approved):
Lgtm

PR #2051, opened 2023-10-14
Update matrix-commander.md

8go:
updated some values for matrix-commander CLI app


thibaultamartin:
Thanks for helping us keeping m.org up to date!

PR #2055, opened 2023-10-20
Ask for consent before displaying YT Player

### LINKED ISSUE #2000: Youtube Player embeds are not GDPR compliant
According to https://complianz.io/youtube-and-the-gdpr-how-to-embed-youtube-on-your-site/ both embedding YouTube player and YouTube-nocookie players are not valid options. I'm asking our legal team to confirm

thibaultamartin:
Fixes #2000

Todo:

- [x] Make the CI not scream on video ids
- [x] Migrate all posts to the `youtube_player` shortcode


thibaultamartin:
https://github.com/matrix-org/matrix.org/assets/1688309/3e4193c1-5e21-4924-85bc-593dba15c4eb

MTRNord:
I didnt yet try it myself but in the video the change in height is quite obvious. Anything we can do about reducing that? It seems a little weird to have the player vs the consent dialog being different in height.

MTRNord:
> Agree, though I'm not sure what's the best solution here. If we are confident the aspect ratio of the yt player is always going to be the same, we can force an [`aspect-ratio`](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) on the placeholder?
> 
> Edit: aspect ratio is not consistent on mobile and desktop

I will have a look in a moment when I am back at my pc. I think we can use the same size hack as used for the player I wrote. But I will need to test that :) will report back :)

sass/_base.scss
```diff
@@ -294,3 +294,39 @@ a>code {
...
+        color: #000;
+        cursor: pointer;
+
+        &:hover {
+            border: 1px solid #fff;
+            color: #fff;
+            background-color: #333;
+        }
+    }
+}
```
MTRNord, line 332:
```suggestion

.youtube_placeholder {
    display: flex;
    flex-direction: column;
    justify-content: center;
    aspect-ratio: 16/9;
    width: 100%;
    background-color: #333;

    div {
        display: flex;
        flex-direction: column;
        color: #fff;
        padding: 1.5rem;
        gap: 1.5rem;

        img {
            margin-inline: auto;
        }

        p {
            max-width: 30rem;
            text-align: center;
            margin-inline: auto;
        }

        button {
            max-width: fit-content;
            margin-inline: auto;
            padding: .5rem 1rem;
            border-radius: 9999px;
            background-color: #fff;
            border: 1px solid transparent;
            color: #000;
            cursor: pointer;

            &:hover {
                border: 1px solid #fff;
                color: #fff;
                background-color: #333;
            }
        }
    }
}
```

image
image

I dont think we can as close on mobile as on desktop due to the required text space. But this at least is closer to what the iframe does.

MTRNord (approved):
LGTM :)

Issue #2010, opened 2023-09-15
Refresh clients data automatically


thibaultamartin:
As of now, clients data in https://matrix.org/ecosystem/clients/ is not refreshed automatically. We can implement a GH Action to fetch the client information "somewhere" and update the data.

My naive approach would be to have a GH Action running every 24h that:

- Fetches all the `*.md` in [`/content/ecosystem/clients`](https://github.com/matrix-org/matrix.org/tree/main/content/ecosystem/clients), and for each file
  - Parses the frontmatter
  - Reads a `definition_file` url (https? git? TBD)
  - Fetches the definition file in the url that contains the updated data
  - Updates the data in the frontmatter of the .md file
- Opens a new PR to update the clients data
- Closes old unreviewed PRs it had previously opened

What's left to define:
- Is it acceptable to ask for the definition_file to be accessible via https?
- What information can be updated by the app maintainer, and what information can only be updated by an editor/curator?
- What format should the definition_file follow?

thibaultamartin:
## What information can be updated by the app maintainer?

- [ ] Data model right now
- [ ] Which information is sensitive/akin to the Foundation's opinion
- [ ] What is left to the maintainers

thibaultamartin:
## What format should the definition_file follow?

- [ ] What are existing formats that work well for that matter
- [ ] Can we reuse existing formats seamlessly? For each, are they simple to put together, are there tools to generate them, are they easy to update, are they used elsewhere relevant for the projects
- [ ] What information we need can they contain by default? What information doesn't fit in? Can they be extended?

cvwright:
How much of this info is contained in the Fastlane metadata that F-Droid uses? https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/

freelock:
There is the "Dublin Core" schema for defining standard elements that might be relevant here - http://purl.org/dc/elements/1.1/ . Drupal publishes an update feed for all of it's modules -- the feed for the Matrix API Drupal module we maintain is here: https://updates.drupal.org/release-history/matrix_api/current . At some point (when I find some time) I'll be spinning up a Chatrix module, to port over the work being done for a WordPress plugin that works in Drupal...

freaktechnik:
[...]

PR #2069, opened 2023-10-30
Add Element X

### LINKED ISSUE #2060: Add Element X to clients

thibaultamartin:
Closes #2060

### DIFF
--- a/content/ecosystem/clients/element-x.md
+++ b/content/ecosystem/clients/element-x.md
@@ -25,4 +25,4 @@ google_play_store.app_id = "io.element.android.x"
 apple_app_store = { app_id = "id1631335820", org = "vector" }
 +++
 
-Element X is prioneering the Matrix 2.0 implementation, notably supporting fast sync with sliding sync.
+Element X is pioneering the Matrix 2.0 implementation, notably supporting fast sync with sliding sync.
--- a/content/ecosystem/clients/element-x.md
+++ b/content/ecosystem/clients/element-x.md
@@ -1,0 +1,28 @@
++++
+title = "Element X"
+template = "ecosystem/client.html"
+[extra]
+thumbnail = "element-x.svg"
+maintainer = "Element"
+maturity = "Beta"
+repo = "https://github.com/vector-im"
+matrix_room = "#element-x-ios:matrix.org"
+licence = "Apache-2.0"
+last_release = "2023-09-21"
+featured = true
+[extra.features]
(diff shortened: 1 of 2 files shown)


content/ecosystem/clients/element-x.md
```diff
@@ -0,0 +1,28 @@
++++
+title = "Element X"
+template = "ecosystem/client.html"
+[extra]
+thumbnail = "element-x.svg"
+maintainer = "Element"
+maturity = "Beta"
+repo = "https://github.com/vector-im"
+matrix_room = "#element-x-ios:matrix.org"
```
MTRNord, line 9: Feels like an oversight of the data model that we only support one room when there are 2 rooms 😬 (iirc there is also an android room for element x specifically)

content/ecosystem/clients/element-x.md
```diff
@@ -0,0 +1,28 @@
...
+threads = false
+sso = false
+multi_account = false
+multi_language = true
+[extra.packages]
+google_play_store.app_id = "io.element.android.x"
+apple_app_store = { app_id = "id1631335820", org = "vector" }
++++
+
+Element X is prioneering the Matrix 2.0 implementation, notably supporting fast sync with sliding sync.
```
MTRNord, line 28:
```suggestion
Element X is pironeering the Matrix 2.0 implementation, notably supporting fast sync with sliding sync.
```

MTRNord (approved):
One typo. Otherwise lgtm

content/ecosystem/clients/element-x.md
thibaultamartin: Hm yes it's worth fixing it. We should definitely support the room variants. I wonder if we should change the data model more in-depth too.

I quite like having one tile per "project", even if it spans across several platforms

content/ecosystem/clients/element-x.md
thibaultamartin: pioneering maybe :D ?

MTRNord: Fixed my suggestion :)

content/ecosystem/clients/element-x.md
MTRNord: I guess this kinda is also related to https://github.com/matrix-org/matrix.org/issues/2010

thibaultamartin:
Let's add this project and rework the data model after

PR #2070, opened 2023-10-31
Upgrade typos

jplatte:


.github/workflows/spell-check.yaml
```diff
@@ -14,4 +14,4 @@ jobs:
       uses: actions/checkout@v2
```
thibaultamartin, line 14: GitHub says

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Can we use it as an opportunity to upgrade to v4 as well?

.typos.toml
```diff
@@ -1,64 +1,65 @@
...
+    "H5atH-tbeRo", # can't get more general regex for otwsu.toml working
+    "KSVTtGsolBXtfsLTeAciXnBc\\.png",
+    "MESSAGEs",
+    "\\`Ther\\`",
+    "\\[ba685b7\\]",
+    "\\[96e9afe6\\]",
 ]
 
 [default.extend-words]
 requestor = "requestor" # valid spelling variant
```
MTRNord, line 55: Hm why are some things moved to the regex while others are not?

jplatte: Nothing was moved to regex one-to-one, the things I moved are things that are not a single "word", ensuring that we still detect typos such as "ther" if it's not inside backticks and so on.

thibaultamartin (approved):
LGTM, thanks for taking care of it!

.typos.toml
MTRNord: Ohhhh. ok yeah makes sense

MTRNord (approved):

PR #2072, opened 2023-11-01
Prevent script bits getting markdown-processed

jplatte:


reivilibre (approved):
Tested to work.

(real fix is the removal of the newlines, which switch the Markdown parser back into expecting Markdown, which then treats the indented JS code as code blocks)

PR #2074, opened 2023-11-02
Use a web component to display the YT player placeholder

### LINKED ISSUE #2073: Improve the YouTube consent placeholder
The YouTube consent placeholder is currently a shortcode, which could be nice but the markdown parser gets in the way.

A better way to do it would be to define a `youtube-player` web component that asks for consent. The component would be defined in a `/js/components.js` file in which all the web components used on m.org are defined (so far only the youtube-player). That file would need to be added globally via the `skel.html` template. This allows us to use `<youtube-player>` everywhere without having to worry about embedding the `components.js` file

It should also be possible to un-consent, likely by going to the privacy policy page and adding a button to set the `ytConsent` key to `false` in the localStorage

thibaultamartin:
Closes #2073

### DIFF
--- a/static/js/components.js
+++ b/static/js/components.js
@@ -1,0 +1,128 @@
+class YoutubePlayer extends HTMLElement {
+    constructor() {
+        super();
+    }
+
+    connectedCallback() {
+        const shadow = this.attachShadow({ mode: "open" });
+        const style = document.createElement("style");
+        style.textContent = `
+            .youtube_placeholder {
+                display: flex;
+                flex-direction: column;
+                justify-content: center;
+                aspect-ratio: 16/9;
+                width: 100%;
+                background-color: #333;
+            }
+
+            .youtube_placeholder > div {
+                display: flex;
+                flex-direction: column;
+                color: #fff;
+                padding: 1.5rem;
+                gap: 1.5rem;
+            }
+
+            img {
+                margin-inline: auto;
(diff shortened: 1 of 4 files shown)


thibaultamartin:
The player seems to be having a hard time resizing correctly

https://github.com/matrix-org/matrix.org/assets/1688309/6d0be8ce-64ca-454f-949f-f1a08b397727

@MTRNord you were the Chief CSS Officer of the YouTube Player department, anything strikes you as problematic here?

static/js/components.js
```diff
@@ -0,0 +1,110 @@
+class YoutubePlayer extends HTMLElement {
+    constructor() {
+        super();
+    }
+
+    connectedCallback() {
+        const shadow = this.attachShadow({ mode: "open" });
+        const style = document.createElement("style");
+        style.textContent = `
+            .youtube_placeholder {
```
MTRNord, line 10:
```suggestion
        style.textContent = `
            iframe {
                aspect-ratio: 16/9;
                width: 100%;
            }
            .youtube_placeholder {
```

Fix for the scaling issue of the iframe. Seems like outer CSS will NOT affect webcomponents.

thibaultamartin: I had completely missed the external `iframe` CSS, thanks!

MTRNord (approved):

PR #2086, opened 2023-11-15
Fix feed authors

### LINKED ISSUE #2082: Rss Feed has Unknown in Author name field
See: https://matrix.org/atom.xml

```xml
<entry xml:lang="en">
    <title>The future of Synapse and Dendrite</title>
    <published>2023-11-06T00:00:00+00:00</published>
    <updated>2023-11-06T00:00:00+00:00</updated>
    <author>
        <name>Unknown</name>
    </author>
    <link rel="alternate" href="https://matrix.org/blog/2023/11/06/future-of-synapse-dendrite/" type="text/html"/>
    <id>https://matrix.org/blog/2023/11/06/future-of-synapse-dendrite/</id>
…
```

_Originally posted by @andrewshadura in https://github.com/matrix-org/matrix.org/issues/1897#issuecomment-1802249711_

MTRNord:
This should fix #2082 

Doing this right before changing trains so I didnt test it yet :D

### DIFF
--- a/templates/macros/atom.html
+++ b/templates/macros/atom.html
@@ -7,6 +7,8 @@
       <name>
         {%- if page.authors -%}
           {{ page.authors[0] }}
+        {%- elif page.taxonomies.author -}
+          {{ page.taxonomies.author[0] }}
         {%- elif config.author -%}
           {{ config.author }}
         {%- else -%}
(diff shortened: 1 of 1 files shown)


MTRNord:
On the question of why not use multiple tags: https://stackoverflow.com/questions/5855993/multiple-authors-in-rss-or-atoms

Seems like the support for this is generally poor.

thibaultamartin (approved):

PR #2088, opened 2023-11-16
Adjusting maturity pills on client fixed #2058

### LINKED ISSUE #2058: Pills can be pushed outside of card by long titles
image

GsnMithra:
I have centered out the title text, and added back the maturity pills to the center of the card.

Fixes #2058


GsnMithra:
Hey, I've started exploring the subgrid approach, and there seems to be some confusion. While I work on resolving the issue, let's have just two cards per row.

two cards per row

PR #2098, opened 2023-11-25
More advanced fork of mattermost appservice bridge is added

nataraj-hates-MS-for-stealing-github:
I guess matrix-as-mm should also be mentioned in the list of mattermost appservice bridges


thibaultamartin:
Thanks for helping us keep the list of bridges up to date :)

PR #2110, opened 2023-12-16
Correct the Neochat info

jacks0n9:
Neochat actually does not have support for e2ee yet as seen in this screenshot:
Screenshot 2023-12-15 at 9 09 29 PM

### DIFF
--- a/content/ecosystem/clients/neochat.md
+++ b/content/ecosystem/clients/neochat.md
@@ -12,7 +12,7 @@ website = "https://apps.kde.org/neochat"
 matrix_room = "#neochat:kde.org"
 featured = false
 [extra.features]
-e2ee = true
+e2ee = false
 spaces = true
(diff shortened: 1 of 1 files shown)


MTRNord:
Hi :) I have asked over in the support chat about this ( https://matrix.to/#/!oXVtZGWmHbaJvGTerl:kde.org/$k20tNeuKO3JcKxB3pm65JZnCt-GAQZgN-JpouuziWdg?via=kde.org&via=matrix.org&via=envs.net )

Turns out you seem to be hitting a known missing part of the code which is being worked on. E2EE in neochat only currently works for new messages. Possibly you might be interested in following the progress of https://github.com/quotient-im/libQuotient/pull/687 which will from what I was told fix your issue. From the view of the website this still means they support E2EE even though its not yet fully featured.

PR #2114, opened 2023-12-23
Update servers.toml

ItsQuadrus:
This PR updates the Synapse repository link on the Servers page (<https://matrix.org/ecosystem/servers>) to reflect the project's move from matrix-org to element-hq.

### DIFF
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -4,7 +4,7 @@ description = "Matrix.org homeserver"
 author = "Matrix.org team"
 maturity = "Stable"
 language = "Python"
-licence = "Apache-2.0"
+licence = "AGPL-3.0-or-later"
 repository = "https://github.com/element-hq/synapse"
 room = "#synapse:matrix.org"
 
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -5,7 +5,7 @@ author = "Matrix.org team"
 maturity = "Stable"
 language = "Python"
 licence = "Apache-2.0"
-repository = "https://github.com/matrix-org/synapse/"
+repository = "https://github.com/element-hq/synapse"
 room = "#synapse:matrix.org"
 
 [[servers]]


content/ecosystem/servers/servers.toml
```diff
@@ -5,7 +5,7 @@ author = "Matrix.org team"
 maturity = "Stable"
 language = "Python"
 licence = "Apache-2.0"
```
MTRNord, line 7: CC @thibaultamartin @joshsimmons 

I am wondering here if it makes sense to instead split this into 2 entries. One for the Apache and one for the AGPL version. 

Either way if its not going to be split there needs to be the license adjusted. Which brings me to: Is synapse agpl3-only or agpl3-or-later? :thinking:

thibaultamartin: Since the Foundation is not going to maintain its own branch, I'm not sure we want to lead people into the direction of an archived repository?

@joshsimmons any strong opinion about having one or two entries for Synapse (the defunct foundation-owned one and Element's)?

joshsimmons: @MTRNord @thibaultamartin  Eep! Somehow I missed the notification, apologies to have been blocking in silence. Thanks for the PR and the thoughtful notes and questions!

This won't be news to y'all, but this is a great example of where we need some consistent guidelines to refer back to – because I can argue this myself several different ways 😅 

My main guiding principle here is to make Matrix accessible while hewing to our values: if easily possible within the current framework, let's split the entries and mark the unmaintained one as obsolete so it doesn't add friction. If not possible, let's just list the new one.

As to the license, yeah, can confirm having double checked just now: agpl3-or-later

Thank you! ❤️

thibaultamartin, line 7: Please update the license as well :)

```suggestion
licence = "AGPL-3.0-or-later"
```

MTRNord: I think (unless someone shouts) I will do the split as part of the gatsby project catchup/cleanup ( #2141 ) I wanted to do this week anyway. That would unblock this PR and get the newer info out-there :)

MTRNord (approved):

PR #2132, opened 2024-01-14
Add Draupnir to the integrations list

MTRNord:
This adds draupnir to the list :)

### DIFF
--- a/content/ecosystem/integrations/integrations.toml
+++ b/content/ecosystem/integrations/integrations.toml
@@ -186,7 +186,7 @@ repository = "https://github.com/jenkinsci/matrix-communication-plugin"
 [[integrations]]
 name = "Time-to-Matrix"
 description = """
-A time-like command that will send end of an arbitrary command output 
+A time-like command that will send end of an arbitrary command output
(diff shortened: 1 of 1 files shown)


content/ecosystem/integrations/integrations.toml
```diff
@@ -204,4 +204,15 @@ maintainer = "etke.cc"
...
+room = "#source:etke.cc"
+
+[[integrations]]
+name = "Draupnir"
+description = """
+A moderation tool for Matrix.
+"""
+maintainer = "Gnuxie"
+language = "TypeScript"
+licence = "COOPERATIVE SOFTWARE LICENSE"
```
MTRNord, line 216: Not sure how else to do it. This has no spdx identifier. So I took the title from the LICENSE file

MTRNord: It has just been relicensed as AFL-3.0

thibaultamartin (approved):
Reviewing this made me realise that the Integrations page needs to be revamped/reorganised/split, but this is out of the scope of this PR.

Looks good to me, thanks for adding it!

Issue #1584, opened 2022-12-07
Define formal processes for determining maturity and featured status of projects


thibaultamartin:
So far the maturity of projects and their featured status has been a "finger in the air estimate", which is flaky, unfair, and impossible to automate.

We need a formal process to assess the maturity and featured status of all projects, including
- [ ] Clients
- [ ] Servers
- [ ] Bots
- [ ] Bridges
- [ ] SDKs

vaguerant:
Hi, I started the above-linked conversation. Expanding slightly on ninchuka's comment as it relates to this issue, I think one formal criterion for "maturity" should be that the software can perform its intended behavior without the need for an alternative software to get it ready.

In Element X's case, the inability to create users or join rooms mean that the app is non-functional without the user employing another client to set everything up before switching to EX. The client only works when used alongside a different (read: mature) client.

Obviously, this idea can be generalized to other clients and parts of the Matrix ecosystem. It's fine and expected for them to rely on other parts of the ecosystem (a client needs a server, etc.) but they should never rely on a different implementation of the same part in order to work.

BrenBarn:
I think it would be good to have a list of features that are a prerequisite for listing in the "featured" category.  What would be even better would be a three-category listing with like "mature and full featured" vs. "mid-tier" vs. "unfinished/WIP" clients.

We have to remember that some people are going to come to this website knowing *absolutely nothing* about Matrix and may be looking to find a client to take their first step with Matrix.  Every client that is featured should be usable *from scratch* to get a decent Matrix experience.

Given that, I would say that creating an account and joining rooms are basic features.  A client without those features is in WIP status, not "ready to be advertised to the general public" status and certainly not "featured" status.

Here is a stab at features I'd consider necessary:

For "ready to use" apps:
* room joining
* complete e2ee support, including. . .
  * cross signing
  * key backup/restore
* basic account management, including. . .
  * create/delete account
  * change password
  * add/change email (for password reset etc.)
* and of course the basic messaging features

For "featured" apps, in addition:
* comprehensive notification settings
* search (in both encrypted and unencrypted rooms)
* spaces
* voice/video calls

This is just off the top of my head, I'm sure there are more features I consider so basic I'm not thinking of them right now. . . :-)

Then any app with all features in both lists could be "featured".  Apps with all the "ready to use" features could be listed as "other options" or something, and apps that don't have all the "ready to use" features would be in a separate category like "early adopter/not recommended for general use" or something.

[...]

PR #2143, opened 2024-01-24
Stop advertising lack of widget support

afranke:
Considering this feature is not even in the spec and only 3 out of 21 listed clients implement it, I don’t think it’s fair to have it in the list. We could do with one less unticked box on the page.

I only touched “my” client so far, but I can amend the PR to remove it from all the non-supporting ones if deemed appropriate. What do you reckon?

Of course, considering the page dynamically picks up whatever feature is listed in the Markdown file, clients that do implement widgets can keep the feature listed!

### DIFF
--- a/content/ecosystem/clients/cinny.md
+++ b/content/ecosystem/clients/cinny.md
@@ -16,7 +16,6 @@ e2ee = true
 spaces = true
 voip_1to1 = false
 voip_jitsi = false
-widgets = false
 threads = false
 sso = true
 multi_account = false
--- a/content/ecosystem/clients/commet.md
+++ b/content/ecosystem/clients/commet.md
@@ -16,7 +16,6 @@ e2ee = true
 spaces = true
 voip_1to1 = false
 voip_jitsi = false
-widgets = false
 threads = false
 sso = false
 multi_account = true
--- a/content/ecosystem/clients/element-x.md
+++ b/content/ecosystem/clients/element-x.md
@@ -15,7 +15,6 @@ e2ee = true
 spaces = false
 voip_1to1 = false
 voip_jitsi = false
-widgets = true
 threads = false
 sso = false
 multi_account = false
--- a/content/ecosystem/clients/element.md
+++ b/content/ecosystem/clients/element.md
@@ -15,7 +15,6 @@ e2ee = true
 spaces = true
 voip_1to1 = true
 voip_jitsi = true
-widgets = true
 threads = true
 sso = true
 multi_account = false
--- a/content/ecosystem/clients/fluffychat.md
+++ b/content/ecosystem/clients/fluffychat.md
@@ -17,7 +17,6 @@ e2ee = true
 spaces = true
 voip_1to1 = true
 voip_jitsi = true
-widgets = false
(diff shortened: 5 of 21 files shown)


thibaultamartin:
Thanks for the suggestion, it makes sense to me. The initial reasoning for adding it was to provide helpful guidance for people looking for a client regardless of whether the feature is in the spec or not… which should not be the Foundation's stance as it paves the road for a "de-facto spec" when we should promote the canonical spec instead.

We don't have criteria for which non-standard feature should be included. Adding vendor-specific features adds 'pressure' to other developers to add those features, despite being vendor prefixed and out of spec.

The Foundation should encourage the original developers to get their features in the spec if they want to advertise it here.

Linking this comment to #1584 for further reference.

> Of course, considering the page dynamically picks up whatever feature is listed in the Markdown file, clients that do implement widgets can keep the feature listed!

Ideally we'd check that the clients all advertise their (lack of) support for the same feature set programmatically (but had to be pragmatic with our resources to get the website out). I'd rather remove the criterion from all clients than to leave it only for some. Can you do this as part of this PR?

afranke:
> I'd rather remove the criterion from all clients than to leave it only for some. Can you do this as part of this PR?

Sure, done.

> We don't have criteria for which non-standard feature should be included. Adding vendor-specific features adds 'pressure' to other developers to add those features, despite being vendor prefixed and out of spec.

I think there’s a balance to be found. “Multi-account” is not part of the spec, but it also doesn’t belong in the spec and won’t ever be there, as it’s not a “client-server” thing and there is no related API. It’s 100% local.

MTRNord:
> I think there’s a balance to be found. “Multi-account” is not part of the spec, but it also doesn’t belong in the spec and won’t ever be there, as it’s not a “client-server” thing and there is no related API. It’s 100% local.

Similar edgecase might be `voip_jitsi` since isnt that widget based mostly? but people do show often links to it too.

afranke:
> Similar edgecase might be `voip_jitsi` since isnt that widget based mostly? but people do show often links to it too.

True, Jitsi is very similar. Again, only 3 clients out of 21 support it, although, funnily enough, not the same 3 (Element X does Jitsi but not widgets, Fluffychat does widgets but not Jitsi). But I didn’t remove it so far because I didn’t feel it was as unfair as widgets. I see the feature as more… high profile?

 I think the “voip_jitsi" wording should be replaced with "Conference calls", which is more implementation neutral and fits with what’s going with Element Calls. I can do the renaming as part of this PR too.

thibaultamartin (approved):

thibaultamartin:
> I think the “voip_jitsi" wording should be replaced with "Conference calls", which is more implementation neutral and fits with what’s going with Element Calls. I can do the renaming as part of this PR too.

I'm a bit wary of being too implementation neutral, especially with MSC3401 coming. If a client that only supports Jitsi-based calls tries to place a call and the other party only supports MSC3401 based calls, users will end up being confused.

Open to feedback on this!

thibaultamartin:
Sounds like a sensible approach to me. Let's go with that and revisit later if needed.

PR #2145, opened 2024-01-24
Improve wording

afranke:
“Last release” implies there won’t be another one. Hopefully this is not the case.


afranke:
This addresses the first point of #1845

thibaultamartin (approved):
Thanks for tackling this!

PR #2156, opened 2024-01-31
Drop supporters section from the index

thibaultamartin:
Drop the supporters section entirely from the index since it's a Platinum member perk and we don't have any as of now


joshsimmons (approved):
LGTM, thank you!

PR #2158, opened 2024-01-31
Make the support CTA more appealing

thibaultamartin:


thibaultamartin:
https://github.com/matrix-org/matrix.org/assets/1688309/f9776295-a35f-4f3d-b75a-e0780f172cb1

PR #2165, opened 2024-02-03
Add spacing and alignment on mobile to the Supporters page

MTRNord:
Before:

grafik

After:

grafik


thibaultamartin (approved):
I had completely missed that during FOSDEM. Thanks a bunch!

PR #2172, opened 2024-02-09
add IndieHosters and XWiki as members

joshsimmons:
\o/


thibaultamartin (approved):
We might want to crop/rework the padding of the logos, but not a blocker to merge this :)

Issue #2178, opened 2024-02-13
Improve the onboarding experience via matrix.org


thibaultamartin:
https://blog.koehntopp.info/2024/02/13/the-matrix-trashfire.html gives the perspective of a person trying Matrix for the first time, and we can do much better!

The main gripes on the matrix.org website are:

- People on macOS miss the install link if they don't open the client details, and can go straight to the app store
- The landing page is too empty and doesn't say much
- The menu goes in responsive mode too early
- The Try Matrix section is too confusing for the general public, and leads to dangerous instances

The rest seems to be related to Element and Element X themselves and is beyond what the matrix.org website can do.

MTRNord:
Additionally, I noticed that currently on the "try matrix now" page the flow is: you get a client and are in the flow, then realise "oh I need a server. What is that?", dive back into the page if you not already lost it and then mess it up only to later realise there was a guide on it.

Instead, it seems to be a better idea to have the download link at the bottom after the explanation. Yes sure it is then further away but it possibly eliminates some confusion in the apps itself.

spaetz:
Given that  MR https://github.com/matrix-org/matrix.org/pull/2179 has already been merged, here my comments that I posted post-merger again:


`**Install Element**`
May I suggest a different wording here? "Install Element" sounds like a high hurdle already and puts me off. And at least for Element-web, that is not needed (even if you decide to install something later). "Use Element" might be a compromise that sounds like the hurdle is much lower (just open it in a webbrowser, hey!).

In the case of mobile, users would likely want to install an app though...

`**Browse clients**`
Only playing a little stupid, but "[Browse clients]" might also be suboptimal. A browse client is e.g. hydrogen or element-web :-). Something like "Discover more clients"?


`**Start chatting**`
More wording nitpicking, sorry. The "Register and start chatting section" has a "Start chatting" button. This should IMHO lead to https://app.element.io/#/register rather than just to some longish manual.
Also the "Read the docs" button uses such a very generic term that I did not know what to expect (the spec?) when clicking it. Perhaps call it "Read client manual" or something?

In the case of mobiles, one could link to the fluffychat web app instead of the element-web register page, which AFAIK has a decent mobile layout? Less hurdles and less intermediary steps and manuals to actually start chatting...

BMaxV:
You're being good sports about this!

[...]

PR #2179, opened 2024-02-13
Improve the Try Matrix page

thibaultamartin:
Contributes to #2178

### DIFF
--- a/templates/try-matrix.html
+++ b/templates/try-matrix.html
@@ -15,7 +15,10 @@
                 <p>We recommend Element as a safe default, but there is a variety of
                     clients available.
                 </p>
-                <a href="/ecosystem/clients" class="call-to-action">Browse clients</a>
+                <div class="cta-wrapper">
+                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
+                    <a href="/ecosystem/clients" class="call-to-action secondary">Browse clients</a>
+                </div>
             </div>
             <img src="/assets/try-matrix/client.svg" alt="Artist view of an abstract matrix client." />
         </div>
@@ -24,14 +27,16 @@
         <div class="step choose-a-server">
             <div class="instructions">
                 <h3>Choose a server</h3>
-                <p>Matrix.org is provided for free by the Foundation, but we
-                    encourage you to <a href="https://joinmatrix.org/servers">find an existing server</a>
-                    or get hosting for your organisation.
+                <p>
+                    Matrix is a decentralised protocol: like email, you need to have an account provider to start using
+                    matrix.
                 </p>
-                <div class="cta-wrapper">
-                    <a href="https://joinmatrix.org/servers" class="call-to-action">Find a public server</a>
-                    <a href="/ecosystem/servers" class="call-to-action">Install your own</a>
-                </div>
+                <p>
+                    But picking a provider can be overwhelming for your first steps. The Matrix.org Foundation is
+                    hosting
+                    the matrix.org Matrix server for every newcomer to create an account. You can safely use it to
+                    create an account when getting started.
+                </p>
             </div>
             <img src="/assets/try-matrix/server.svg" alt="Artist view of an abstract server." />
         </div>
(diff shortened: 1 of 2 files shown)


templates/try-matrix.html
```diff
@@ -24,14 +27,16 @@ <h3>Choose a client</h3>
...
-                    encourage you to <a href="https://joinmatrix.org/servers">find an existing server</a>
-                    or get hosting for your organisation.
+                <p>
+                    Matrix is a decentralised protocol: like email, you need to have an account provider to start using
+                    matrix.
+                </p>
+                <p>
+                    But picking a provider can be overwhelming for your first steps. The Matrix.org Foundation is
+                    hosting
+                    the matrix.org Matrix server for every newcomer to create an account. You can safely use it to
```
MTRNord, line 37: I think it would be nice to say how to use it. I think thats one of the key issues of the blogpost we saw. We dont talk enough about the "how".

thibaultamartin: Hm, do you suggest adding instructions regarding how to create an account using Element?

MTRNord: If we link to element directly now, I would argue we should do that yes.

templates/try-matrix.html
```diff
@@ -15,7 +15,10 @@ <h3>Choose a client</h3>
                 <p>We recommend Element as a safe default, but there is a variety of
                     clients available.
                 </p>
-                <a href="/ecosystem/clients" class="call-to-action">Browse clients</a>
+                <div class="cta-wrapper">
+                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
```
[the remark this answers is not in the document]
MTRNord: Hm yeah makes sense.

templates/try-matrix.html
thibaultamartin: Does it work for you if we link to https://matrix.org/docs/chat_basics/matrix-for-im/#creating-a-matrix-account ?

MTRNord: yes

MTRNord (approved):

templates/try-matrix.html
```diff
@@ -12,10 +12,13 @@ <h1>{{ page.title }}</h1>
...
             <div class="instructions">
                 <h3>Choose a client</h3>
-                <p>We recommend Element as a safe default, but there is a variety of
-                    clients available.
+                <p>
+                    We recommend Element as a safe default, but there is a variety of clients available.
                 </p>
-                <a href="/ecosystem/clients" class="call-to-action">Browse clients</a>
+                <div class="cta-wrapper">
+                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
```
spaetz, line 19: May I suggest a different wording here? "Install Element" sounds like a high hurdle already.  And at least for Element-web, that is not needed (even if you decide to install something later). "Use Element" might be a compromise that sounds like the hurdle is much lower (just open it in a webbrowser, hey!).

In the case of mobile, users would likely want to install an app though...

spaetz: Only playing a little stupid, but "[Browse clients]" might also be suboptimal. **A** `browse client ` is e.g. hydrogen or element-web :-). Something like "Discover more clients"?

templates/try-matrix.html
```diff
@@ -44,7 +49,11 @@ <h3>Register and start chatting</h3>
                     browse the <a href="/docs/chat_basics">chat basics</a>
                     documentation.
                 </p>
-                <a href="/docs/chat_basics" class="call-to-action">Read the docs</a>
+                <div class="cta-wrapper">
+                    <a href="/docs/chat_basics/matrix-for-im/#creating-a-matrix-account" class="call-to-action">Create
+                        an account</a>
+                    <a href="/docs/chat_basics" class="call-to-action secondary">Read the docs</a>
```
spaetz, line 55: More wording nitpicking, sorry. The Register and start chatting section has a "Start chatting" button. This should IMHO lead to https://app.element.io/#/register rather than just to some longish manual. Also the "Read the docs" is a very generic term that *I* did not know what to expect (the spec?). Perhaps call it "Read client manual" or something. I'll stop here.

PR #2192, opened 2024-02-16
fix: update maturity level of gomatrix SDK to "Obsolete"

MTRNord:
Fixes https://fosstodon.org/@lil5/111940191501771658


joshsimmons (approved):
Thank you!

PR #2182, opened 2024-02-13
Add Distributions Page

MTRNord:
Fixes #2068

image

# TODOs:

- [ ] Display support level on the cards itself or split it into 2 sections of cards.

PR #2194, opened 2024-02-17
Syncloud provider

cyberb:


thibaultamartin:
Hello and thanks for opening the PR! Syncloud looks like a very interesting service: providing people with the opportunity to self-host without the hassle of making their whole network public and fiddling with dyndns is really helpful!

We have not made it obvious enough that hosting providers listed on matrix.org need to do the hosting themselves. We think Syncloud would be a better fit for the upcoming distributions page (#2182)

We encourage you to to update this PR to add Syncloud to the distributions page once #2182 is merged!

cyberb:
Thanks for the guidance! As suggested, I've moved Syncloud to the new distributions page. Closing this in favour of #3546.

PR #2206, opened 2024-02-22
Add acter.global as hosting provider

gnunicorn:
As it says on the tin: Acter offers hosting of matrix home servers (for a specific target audience), too.


thibaultamartin:
Hej @gnunicorn and thanks for opening the PR! Love what you're building with acter, and we need to advertise it on the Matrix.org website

We don't have formal guidelines _yet_ for what is a hosting provider, but there seems to be a consensus emerging around it needing to be the main product, or at least having a section on the website explaining how to discard the non-matrix bits of the product if you're not interested in them.

We have opened #2218 to come up with more formal requirements and will keep you updated as the section emerges! Looking forward to advertising Acter, keep up the good work 💪

gnunicorn:
I am not insisting Acter shows up on that page. I just answered the call I saw on some social platform to add hosting providers that offer Matrix to add them, and thought, that well, technically we do offer Matrix hosting and then some additional stuff (as do others), that can be used with any regular Matrix client. I thought the call was because there was a need to boost the number of providers shown to have an appealing set to refer to. If our entry isn't wanted, I won't be insisting this should be added.

joshsimmons:
@gnunicorn @thibaultamartin With apologies to both of you, I put out that call on social and did not do enough pre-work to set PRs like this up for success! I'm delighted by the work that Acter is doing and it is definitely something we want to showcase. I'll follow up internally to get our ducks in a row and come back to this.

hasan42810-code (commented):
All show chad

PR #2207, opened 2024-02-22
Add MSC1929 support contacts

turt2live:
The "admin" contact is largely meant to be used for reporting users and such, so we supply our abuse contact there.


joshsimmons (approved):
Woohoo! Thank you.

PR #2252, opened 2024-03-21
Move visuals out of the way as the screen narrows on index

thibaultamartin:


thibaultamartin:
When the screen narrows, the visuals get in the way. This was supposed to be addressed by a media query to change the position of the elements, but feels jumpy and it doesn't really work.

image

This PR fixes this by moving the decoration out of the way as the screen narrows. In the recording it steel feels a bit jumpy because of `flex-wrap` happening on other elements of the page.

https://github.com/matrix-org/matrix.org/assets/1688309/ddecfeba-66f7-448b-b171-2f48a97a791f

PR #2267, opened 2024-03-29
Prevent the submenu from disappearing

thibaultamartin:


MTRNord (changes requested):
as mentioned in the DM it is not quite up enough to work it seems. at least on my 4k monitor :) Feels like its off by ~1-2px or so

MTRNord:
Replaced by https://github.com/matrix-org/matrix.org/pull/2297

PR #2277, opened 2024-04-03
Clarify title of MSC in Spec release blogpost

MTRNord:


turt2live:
(the changelog in the blog post doesn't matter as much, and we're not likely to do a patch release to fix the changelog - I'd suggest relying on the click-through link for this one)

MTRNord:
> (the changelog in the blog post doesn't matter as much, and we're not likely to do a patch release to fix the changelog - I'd suggest relying on the click-through link for this one)

In all honestly that's imho horrible but I leave it at that (with my personal private hat and not a Website maintainer hat).

turt2live:
That's a significantly harsher reaction than I was expecting. Would you be able to walk me through why it's a horrible approach?

PR #2280, opened 2024-04-03
Update links to the Ruma website

jplatte:
We're moving from ruma.io to ruma.dev.

Intentionally left the text alone, seems weird to retcon it ^^


MTRNord (approved):
LGTM

PR #2288, opened 2024-04-11
Fix typo in TWIM 20240329

Kladki:
It may be a good idea to have spell check in the CI to prevent future typos.


MTRNord (approved):
Thanks. We do actually have a CI for typos, but sadly some things still fall through.

Issue #2278, opened 2024-04-03
Hover on Ecosustem menu item is broken


shadow-identity:
Repro: 
1. Go to matrix.org
2. Hover over Ecosystem menu item to see the dropdown list
3. Try to select any item in the dropdown.

https://github.com/matrix-org/matrix.org/assets/1223112/2a3492a3-91c2-453c-ae14-4b2735689591

Firefox Dev Edition 125.0b7 (64-bit)
macOS 14.4.1 (23E224)

MTRNord:
Yeah thats known and being worked on at #2267

Issue #2276, opened 2024-04-03
Maturity button is pushed off the card if the title  is long on SDKs and Servers


tawandamoyo:
If the name of the SDK or Server provider is too long the maturity button ie "beta", "stable" etc is pushed out of the card

![cardBug2](https://github.com/matrix-org/matrix.org/assets/17948595/490d4e31-cb16-45fc-a72f-3e234e60cbae)
![cardBug](https://github.com/matrix-org/matrix.org/assets/17948595/9fd0d2a2-5d99-46d6-91f0-530ce0cdf4ad)

tawandamoyo:
already noted at #2058

Issue #2261, opened 2024-03-28
Regression: The margin-top on the try-matrix button should only show up on mobile but not always.


MTRNord:

thibaultamartin:
Fixed in [6b41360](https://github.com/matrix-org/matrix.org/commit/6b4136090c55f685ef0253599e4aa41302752ac9)

Issue #2180, opened 2024-02-13
The navbar collapses too early


thibaultamartin:
The navbar collapses into a responsive menu while the screen is still wide. This is rather unsettling.

https://github.com/matrix-org/matrix.org/assets/1688309/995a861e-3d24-4a6d-821b-8f658c2b4cc1

This happens because the navbar has plenty of items in it.

Some suggestions to improve:

- Documentation -> Docs
- less padding
- drop Shop, maybe put that in the footer, is not a priority

PR #2301, opened 2024-04-19
Testing

MTRNord:
Some more recent issues (#2278 , #2276 , #2261 and #2180 and probably others) have shown a pattern of things that may be possible to catch with testing. So I decided to get a stab at building a simple playwright based testsuite for this. 

The testsuite will also bring tests for accessibility based on https://playwright.dev/docs/accessibility-testing which hopefully helps us on that front as well.


MTRNord:
One thing to clear up is if we follow wcag2a or wcag2aa :thinking: (it can be configured)

My opinion is to go with aa since we want a partial redesign anyway. But if not its easy to change the tests to only run against a which should allow us to run all tests afaik without skipping.

MTRNord:
One recurring issue is the contrast of the urls we link to by the way.

MTRNord:
Since we also just had the case that the frontpage failed due to the link in the new banner (which is also fixed by #2319 ) here is an example where we can find the html reports of these errors: https://github.com/matrix-org/matrix.org/actions/runs/8859887109?pr=2301 under artifacts there is a `html-report--attempt-1`  artifact which will contain a human readable format. Sadly I didnt find a nice way to have it like a comment on PRs. In the future maybe thats a fun sideproject for someone to do but for now there isnt anything like that :)

playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        /* Test against branded browsers. */
+        // {
+        //   name: 'Microsoft Edge',
+        //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+        // },
+        // {
+        //   name: 'Google Chrome',
+        //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+        // },
```
thibaultamartin, line 69: Why are mobile tests commented?

tests/filters.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    // Check if platform filter works
+    await page.locator("#filter-platform").scrollIntoViewIfNeeded()
+    await page.locator("#filters-overlay").isHidden()
+    await page.locator("#filter-platform-menu").isHidden()
+    await page.locator("#filter-platform").click()
+    await page.locator("#filters-overlay").isVisible()
+    await page.locator("#filter-platform-menu").isVisible()
+
+    // Check if platform filter can be used
+    await page.locator("div.reset-links:nth-child(9) > button:nth-child(1)").click()
```
thibaultamartin, line 16: I wonder if `nth-child` is our best option here. Minor changes in the UI could break tests. I didn't make my mind around whether it's a good thing or not.

tests/filters.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    await page.locator("#filters-overlay").click()
+    const clients = await page.locator("#all-clients").all()
+    for (const client of clients) {
+        await client.isHidden()
+    }
+
+    await page.locator("#filter-platform").click()
+    await page.locator("div.reset-links:nth-child(9) > button:nth-child(2)").click()
```
thibaultamartin, line 24: Same as above with `nth-child`

tests/menu.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    await page.getByLabel('Ecosystem').hover();
+    const elements = [
+        page.locator(".section-submenu > a:nth-child(1)"),
+        page.locator(".section-submenu > a:nth-child(2)"),
+        page.locator(".section-submenu > a:nth-child(3)"),
+        page.locator(".section-submenu > a:nth-child(4)"),
+        page.locator(".section-submenu > a:nth-child(5)"),
+        page.locator(".section-submenu > a:nth-child(6)"),
+    ]
```
thibaultamartin, line 15: I'd be surprised if there wasn't a "cleaner" way to get all direct children links

playwright.config.ts
MTRNord: Mostly the default template. I didnt question it tbh. We can try them :)

tests/filters.spec.ts
MTRNord: We can probably make it smarter. these are just from the devtools copying the css-selector.

playwright.config.ts
MTRNord: Oh and the filter and menu tests fail on mobile currently (as expected). So thats something I need to solve

tests/menu.spec.ts
MTRNord: There was and I fixed it :)

playwright.config.ts
MTRNord: And fixed :)

PR #2307, opened 2024-04-21
Contributor name change

MRAAGH:
To respect this name change, I'd like to update the mention in my past announcement


joshsimmons (approved):
Very thoughtful, thank you. And congrats to Ellie 💖

PR #2309, opened 2024-04-25
Publish the post-mortem of the outage of Monday 22

thibaultamartin:


MTRNord (approved):

.typos.toml
```diff
@@ -26,6 +26,9 @@ extend-ignore-re = [
     # youtube embed in articles
     "\\{\\{ youtube_player\\(video_id=\"[^\"]+\"\\) \\}\\}",
 
+    # Matrix
+    "morg", # shorthand for m.org or matrix.org
```
clokep, line 30: I think using an abbreviation for this is an anti-pattern to be avoided. It adds jargon with no additional value.

PR #2316, opened 2024-04-27
Add Synapse update to TWIM 2024-04-26

anoadragon453:
Oops, I forgot to TWIM this week. I am a bad person.

This announces the latest Synapse security release, so should probably go in.


joshsimmons (approved):
good catch, thank you!

PR #2319, opened 2024-04-27
Fix the contrast of links and ensure we underline those links

MTRNord:
This fixes some errors from my testing branch. I thought it might be a good idea to get these on a dedicated branch.

Explicitly it fixes https://dequeuniversity.com/rules/axe/4.9/color-contrast and `Ensure links are distinguished from surrounding text in a way that does not rely on color` which is part of WCAG 2.1 (A), WCAG 2.0 (A), WCAG 2.2 (A), Trusted Tester and EN 301 549

# Before and after images
image
image

This gives us _at least_ a 5.40:1 contrast ratio on all backgrounds. Which is enough for the AA requirements.

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -62,6 +62,10 @@ a {
     text-decoration: underline;
 }
 
+nav a {
+    text-decoration: none;
+}
+
 ul,
 ol {
     margin-top: 0px;
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -11,7 +11,7 @@
         --color-text-light: #595959;
     }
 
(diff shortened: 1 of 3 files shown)


thibaultamartin:
This seems to have a negative side effect on e.g. CTAs

image

MTRNord:
Hm I thought it would be ok like that but I can remove the underline easily :)

MTRNord:
@thibaultamartin fixed your concern :)

MTRNord (commented):
# TODOs

- [x] Ecosystem cards now are fully underlined which is a bug

PR #2324, opened 2024-04-28
Improve the Nominee Page

MTRNord:
This does a few things:

- Sort the member types to match the above text for easier reading flow
- Add the maximum possible seats to the title to avoid having to scroll up as often
- Make the members look like headers (semantically they aren't, but visually it makes it easier to follow together with the next change)
- Add headers instead of a list to each member to make clear which part refers to what

## Before
image

## After
image

### DIFF
--- a/sass/_elections.scss
+++ b/sass/_elections.scss
@@ -7,10 +7,27 @@
         color: var(--color-text);
     }
 
-    .nominees button {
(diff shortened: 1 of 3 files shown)


joshsimmons (approved):
Great refinements, thank you!

PR #2333, opened 2024-05-05
Add conduwuit and repo for jSynapse to servers metadata

clokep:
Conduwuit is a hard-fork of conduit, so is worthwhile listing as a separate homeserver. Adds the repository for jSynapse, which was missing.

### DIFF
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -71,6 +71,7 @@ author = "Swarmcom"
 language = "Java"
 maturity = "Obsolete"
 licence = "Apache-2.0"
+repository = "https://github.com/swarmcom/jSynapse"
 
 [[servers]]
 name = "Maelstrom"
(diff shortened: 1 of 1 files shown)


clokep:
CC @girlbossceo -- not sure if I portrayed conduwuit correctly?

girlbossceo:
Looks good! Is there room for a name change in the future? The name conduwuit occasionally causes confusion as to whether we're a soft fork or hard fork or not, and I would like to eventually get it rebranded (still with the uwu-ness, but detaching from Conduit), just not right now.

clokep:
> Looks good! Is there room for a name change in the future?

I don't see why not! We will just rename it (and maybe leave a reference for formerly known as?)

girlbossceo (approved):
Sounds good, as the conduwuit maintainer this PR has my approval.

thibaultamartin (approved):

PR #2338, opened 2024-05-09
Add elections page to a dropdown in the navigationbar

### LINKED ISSUE #2317: Make elections page accessible through the navbar
maybe under about.

maybe also mention it on the about page.

MTRNord:
Fixes #2317

PLEASE make sure that this actually didnt break existing links before merging! I think I got them all but I may have missed one.


content/foundation/_index.md
```diff
@@ -0,0 +1,8 @@
++++
+title = "The Matrix.org Foundation"
+sort_by = "weight"
+render = false
++++
+
+* [About](about) &ndash; The Matrix.org Foundation About page
```
MTRNord, line 7: I think this is technically unreachable since its render=false and the about page itself takes this url.

MTRNord:
image

Adds a submenu called "Foundation" (due to the way the hover logic works I had to nest it like this) and it also fix the weird padding situation the submenu was in.

HarHarLinks:
I'm not sure "About Matrix" is the best title for the about page

thibaultamartin:
I find "About Matrix" to be okay ish, since it talks both about the protocol and the foundation. In the future we might split the content in two distinct pages, but as it is now it works for me.

Do you have ideas that work better?

HarHarLinks:
I don't think it really is about the protocol, and it is in the Foundation dropdown already, I'd keep to the previous simple "about" or go for something that hints more in the direction of project governance

thibaultamartin (approved):

PR #2347, opened 2024-05-17
Update twim-guide.md

HarHarLinks:
Please check that I caught all s/author/editor occurrences

### DIFF
--- a/content/twim-guide.md
+++ b/content/twim-guide.md
@@ -41,7 +41,7 @@ to help the editors to lower their workload when publishing the post.
 ### Markdown
 
 The bot expects markdown as part of the plaintext event body. Please use a client which
-does include the markdown in the plaintext. While we currently accept other formats as well
+sends the markdown formatted text in the "plaintext" body. While we currently accept other formats as well
 it will add manual work for the post by a human.
 
 ### Do not treat it as a Changelog
--- a/content/twim-guide.md
+++ b/content/twim-guide.md
@@ -5,9 +5,10 @@ title = "Writing a good \"This Week in Matrix\" entry"
 This Week in Matrix (TWIM) is a weeekly digest on the blog in which
 anyone can post small concise updates about their Matrix projects, events
 and happenings.
-Those updates are collected in a digest, published every Friday on
-the [matrix.org blog](/blog/twim). It aims to allow the community to show off their
-progress in the Matrix ecosystem.
+Those updates are collected through the Matrix room
+[#thisweekinmatrix:matrix.org](https://matrix.to/#/%23thisweekinmatrix%3Amatrix.org)
+and published every Friday on the [matrix.org blog](/blog/twim).
+It aims to allow the community to show off their work in the Matrix ecosystem.
 
 ## Prerequisites
 
--- a/content/twim-guide.md
+++ b/content/twim-guide.md
@@ -15,7 +15,7 @@ There are some guidelines you should be aware of before publishing your
 update to TWIM:
 
 Like all things on matrix.org your post has to follow and comply with the
-[Matrix.org Code of Conduct](/legal/code-of-conduct). Additionally the authors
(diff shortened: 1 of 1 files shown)


content/twim-guide.md
```diff
@@ -27,28 +27,28 @@ Matrix room.
...
+does include the markdown in the plaintext. While we currently accept other formats as well
 it will add manual work for the post by a human.
 
 ### Do not treat it as a Changelog
 
 Copying a Changelog is easy but not what TWIM is about. Ideally we prefer posts
-which are a few paragraphs long outlining features which were added or changed
-over a bulletpoint list. This makes it more readable to humans. You may still of course
+which are a few paragraphs long outlining some highlights
+over a long detailed list. This makes it more readable to humans. You may of course still
```
MTRNord, line 51: Hm isnt this the opposite of the original? A bulletpoint list for me is a list of not detailed things. You essentially just have headers in it.

content/twim-guide.md
```diff
@@ -27,28 +27,28 @@ Matrix room.
...
 Making a good entry can have some challenges. Below are some recommendations
-to help the authors to lower their workload when publishing the post.
+to help the editors to lower their workload when publishing the post.
 
 ### Markdown
 
-The bot expected markdown as part of the plaintext event body. Please use a client which
-does send the plaintext as markdown. While we currently accept other formats as well
+The bot expects markdown as part of the plaintext event body. Please use a client which
+does include the markdown in the plaintext. While we currently accept other formats as well
```
MTRNord, line 44: The wording "send" was actually used as a stricter form of this. Meaning we expect it to include _only_ markdown in the ideal case. "include" could mean that it contains multiple formats which would still need human intervention.

MTRNord (changes requested):

content/twim-guide.md
HarHarLinks: - I think having bullet points as a semantic element is OK, e.g. see element X updates. I would actually prefer it to a paragraph where I only know after reading how many topics were covered.
- I think posts, whether they are written as text or 1 sentence per bullet point, should focus on highlights and not every single detail (such as commit messages). That level of detail should be included by linking to release notes/changelog if desired.

content/twim-guide.md
HarHarLinks: I don't understand "sending the plaintext as markdown". Rather I think what you mean is to send markdown (i.e. text with symbols) in the plaintext `body` of message events. That is in contrast to
- HTML `formatted_body`, because zola needs the md
- plain plaintext in the plaintext `body`, i.e. markdown symbols removed

People will/should/have been all along sending messages (in most cases) by typing in markdown, which their client will convert to HTML and include that in the actual event. I don't think there is a reasonable way around that. What you want here is the behavior of element web, which keeps the entered markdown untouched in the body as a fallback.

I actually think that the spec
> The plain text version of the HTML should be provided in the `body`.

specifies that the `body` would not be the original markdown (which is an element and other clients implementation detail to format messages in the first place) but the HTML with all tags stripped.

MTRNord: Well we had the case before that we got orgmode instead since thats what the client uses. With then a codeblock for markdown. And I would like to discourage this :)

HarHarLinks: Exactly, it should "include markdown in the plaintext" rather than whatever else. We might also word this entirely different, e.g. "use a client which sends the markdown formatted text in the "plaintext" `body`". It's a family technical requirement that might be confusing for non-technical submitters though.

MTRNord: We agreed upon your "use a client which sends the markdown formatted text in the "plaintext" body" suggestion for this in the office hour. Could you update it accordingly? :)

HarHarLinks:
```suggestion
sends the markdown formatted text in the `plaintext` body. While we currently accept other formats as well
```

PR #2355, opened 2024-05-22
Use `system-ui` as default font

veyndan:
I'm on Ubuntu, and the font that's currently shown for me on matrix.org is Nimbus Sans. The font in the existing font list that is causing this is the font "Helvetica", which I do not have installed.

From my cursory understanding of font matching, when running `fc-match helvetica` in the terminal, I get `NimbusSans-Regular.otf: "Nimbus Sans" "Regular"`, which explains why I'm being shown Nimbus Sans.

As it appears from the existing font list, the desired result was to use the font of the system the user is on. By using `system-ui`, we can achieve this. In my case, I would be shown the Ubuntu Font, as that's my default on my system.

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -35,7 +35,8 @@ body {
     background-color: #fff;
     color: var(--color-text);
 
-    font-family: system-ui,
+    font-family: "Helvetica Neue",
+        system-ui,
         sans-serif,
         "Apple Color Emoji",
         /* Emojis*/
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -35,13 +35,7 @@ body {
     background-color: #fff;
     color: var(--color-text);
 
-    font-family: "Helvetica Neue",
-        "Helvetica",
-        Arial,
-        -apple-system,
-        BlinkMacSystemFont,
-        "Segoe UI",
-        Roboto,
+    font-family: system-ui,
         sans-serif,
         "Apple Color Emoji",
         /* Emojis*/


HarHarLinks:
Nimbus Sans is a clone of Helvetica (https://en.wikipedia.org/wiki/Helvetica#Nimbus_Sans) so this might be working as intended. Is Helvetica free? (Can we default to free fonts for our free standard's website?)

veyndan:
Ah, TIL about the purpose of Nimbus Sans.

> Is Helvetica free?

Nope. It's proprietary.

The issue with defaulting to free fonts is that we're either assuming that the user has one of the listed free fonts already installed on their computer, or we increase the footprint of the website by having the user download a font.

Is the suggestion to use a free font because of a perceived issue or deficiency in the usage of `system-ui`?

thibaultamartin:
Thanks for the PR! We are missing `system-ui` for sure, but I would go with https://systemfontstack.com. Also please don't remove Helvetica Neue from the list and keep it in the first position since it's part of Matrix' branding :)

veyndan:
> We are missing `system-ui` for sure, but I would go with https://systemfontstack.com/.

`system-ui` isn't mentioned on https://systemfontstack.com at all, indicating that the site's recommendations perhaps predate `system-ui` (or at least predate it having strong cross-browser support). I was using https://github.com/system-fonts/modern-font-stacks?tab=readme-ov-file#system-ui as a reference, which seems to have much more up-to-date advice.

> Also please don't remove Helvetica Neue from the list and keep it in the first position since it's part of Matrix' branding :)

Added back!

veyndan:
Not sure why CI is failing. Running `zola build` locally passes :shrug:

MTRNord:
> Edit: I just noticed that main is failing with the same error, so the CI failure is unrelated to this PR.

oops thanks for noticing. my fault :D

veyndan:
Rebased onto `main` and CI is now passing :)

thibaultamartin:
Sorry we're just reviewing it right now, and it looks like [using system-ui is a rather bad idea](https://infinnie.github.io/blog/2017/systemui.html). Let's use https://systemfontstack.com/ but with Helvetica Neue on top

veyndan:
I don't mean to be a bother about this, but the article you linked is from 2017, and the advice has significantly changed since then. The article points to Bootstrap getting rid of `system-ui` in April 2017:

> Apr 10 update: Bootstrap has also removed `system-ui` from the font-family stack as of [this pull request](https://github.com/twbs/bootstrap/pull/22377).

However, Bootstrap later added back `system-ui` to its font stack with seemingly no issues as `system-ui` has remained in their stack for the past 4 years (https://github.com/twbs/bootstrap/pull/30561). The mentioned issue on Chinese Windows appears to be a bug of the distant past (https://github.com/twbs/bootstrap/pull/22377#issuecomment-2143801992).

I've not heard any recent (i.e., last 2 years) complaints about `system-ui`, as all prior issues appear to have been fixed quite a while ago.

mirabilos:
It’s still a cross-platform concern. On Linux with no desktop environment, there is not even a way to *set* that.

Please do **not** use `system-ui`.

mirabilos:
The problem is that it *does* resolve to a font, but to one that is shit **and** that users cannot configure, unlike the standard `serif`, `sans-serif` and `monospace`.

veyndan:
That sounds like an issue with your very specific set up, not an issue with `system-ui`. I admittedly use Linux with a desktop environment, but looking around online, it doesn't seem like setting a default font for Linux without a desktop environment should be a problem.

You're making a lot of claims without backing them up. If you could link to something that backs up your claims, I'd be happy to give it a read.

mirabilos:
On Sun, 20 Oct 2024, Veyndan Stuart wrote:

>but looking around online, it doesn't seem like setting a default font
>for Linux without a desktop environment should be a problem.

Oh, tell me where ☺ I only got here via other threads where people have
this problem.

>You're making a lot of claims without backing them up. If you could
>link to something that backs up your claims, I'd be happy to give it a
>read.

Careful, that’s a very Wikipedia attitude.

To read, there is the other issue that was linked (via which I came
here). Other than that, does not my complaint that it results in an
illegible font suffice?

HarHarLinks:
For reference, here is what the spec site uses: https://github.com/matrix-org/matrix-spec/blob/fca171427f11233cb1b82982ac541dd206ebbff8/assets/scss/_variables_project.scss#L42-L53
It may make sense to align with it.

PR #2365, opened 2024-05-28
Update hosting provider text for etke.cc

spantaleev:


thibaultamartin (approved):
Thanks for helping us stay up to date!

PR #2366, opened 2024-05-29
Add: (clients/ement.md) Ement.el

alphapapa:
After about 3.5 years of development and releases, and being published in several Linux distros, I think it's time to add it to the official clients list.  :)  I tried to follow the format in the other clients' files; please let me know if any changes are needed.  Thanks.


content/ecosystem/clients/ement.md
```diff
@@ -0,0 +1,28 @@
...
+[extra]
+thumbnail = "ement.svg"
+maintainer = "Adam Porter"
+maturity = "Stable"
+repo = "https://github.com/alphapapa/ement.el"
+matrix_room = "#ement.el:matrix.org"
+licence = "GPL-3.0-or-later"
+language = "Emacs Lisp"
+latest_release = "2024-04-16"
+featured = true
```
thibaultamartin, line 13:
```suggestion
featured = false
```

We don't have formal processes to define what is featured or not (it's tracked in https://github.com/matrix-org/matrix.org/issues/1584) but the rule of thumb is "can it be installed and used by the general public".

While emacs is certainly very popular in the tech-savvy populations, it might be slightly too powerful to handle for the general public :)

thibaultamartin (changes requested):

content/ecosystem/clients/ement.md
alphapapa: Emacs may have a reputation for being software only used by developers and such, but you might be surprised by how many non-developers use it.  Emacs is widely used by people from a variety of backgrounds for a variety of purposes (such as writing, research, bookkeeping, publishing, note-taking, communication, blogging, PIM--the list is nearly endless).  Many of its users have never written a line of code.  And every year EmacsConf includes presentations from non-software developers.

So I would respectfully disagree: if "the general public" includes "people who use PCs" (i.e. not just smartphones/tablets), then Emacs certainly qualifies as usable by them.  (It's not like, e.g. CLI-only apps; although Emacs can run in a terminal, it has a full GUI as well.)

But if you do insist on this change, please go ahead and make it (assuming the UI allows you to as easily as clicking that button).

Thanks.

alphapapa: Ok, I'll go ahead and click it myself.  :)  Thanks.

MTRNord (approved):

alphapapa:
Thanks, @MTRNord.

PR #2386, opened 2024-06-08
Fix ement.el breaking due to invalid html (a inside of a)

### LINKED ISSUE #2380: ecosystem/clients page not rendered correctly
Since merging https://github.com/matrix-org/matrix.org/pull/2366, it seems that something in the `clients.html` template is not rendering correctly.  Here's a screenshot of the rendered HTML and one of the DOM, showing that the `{% for client in clients %}` block at the bottom of the template is doing something weird with the newly added client's card.

bug

dom

Not being familiar with Zola and Tera directly (though they are similar to Hugo), I can't discern the problem quickly enough, so I'll leave this bug report.  Thanks.

MTRNord:
Fixes #2380

### DIFF
--- a/.github/workflows/zola.yml
+++ b/.github/workflows/zola.yml
@@ -17,7 +17,7 @@ jobs:
 
       - name: Download Zola
         run: |
-          version='v0.17.2'
+          version='v0.18.0'
           archive="zola-${version}-x86_64-unknown-linux-gnu.tar.gz"
           curl -L "https://github.com/getzola/zola/releases/download/${version}/${archive}" -o "$archive"
           tar xf "$archive"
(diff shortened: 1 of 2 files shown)


MTRNord:
This PR requires zola 0.18.0 :/ Hence cloudflare fails

thibaultamartin:
@guillaumevillemont can you could update the version of Zola CF Pages?

guillaumevillemont:
> can you could update the version of Zola CF Pages?

I've done it for previews deployments and it's happy. Should I do it now for the production too ?  I'm not sure about backward compatible Zola is (like, if you'd merge other PR before this one for example)

MTRNord:
> > can you could update the version of Zola CF Pages?
> 
> I've done it for previews deployments and it's happy. Should I do it now for the production too ?  I'm not sure about how backward compatible Zola is (like, if you'd merge other PR before this one for example)

If it works in the preview it works in production as well. According to the changelog there have been no breaking changes and the ci seems to be happy as well :) so imho it's ready for prod too.

guillaumevillemont:
@MTRNord alright done, it's using v0.18.0 everywhere now !

MTRNord:
> @MTRNord alright done, it's using v0.18.0 everywhere now !

Thanks @guillaumevillemont ! 🦆

thibaultamartin (approved):

PR #2410, opened 2024-06-20
Update Nheko information

deepbluev7:
Using the old flathub id would be very confusing. And while it is sad, mujx hasn't been a maintainer for a while. They certainly started the project and did a humongous of work, but I think maintainers should probably reflect current contacts at this point?

The room change is mostly a visual thing. It points to the same room, but we usually use the nheko.im one and it is the canonical alias, so it probably is better to document that this way as well.


HarHarLinks:
resolve s#2409

MTRNord (approved):

PR #2411, opened 2024-06-20
Fix Ement.el client card overlapping others

veyndan:
This isn't a great solution, as I just ripped out the link to Emacs (which I think is useful), though I don't understand Zola well enough to fix the root cause. If anyone has a solution that keeps the link, please add to or replace the PR with your solution!

|Before|After|
|---|---|
|image|image|


HarHarLinks:
dupe of #2386?

veyndan:
Oops, it is! I'll close this one out then.

PR #2414, opened 2024-06-20
Add a missing display flex to ensure the submenu hover works on 1x 1440p

### LINKED ISSUE #2413: navbar dropdowns broke
image

MTRNord:
Fixes #2413


HarHarLinks (approved):
i'm not sure why you chose to left align it relative to the navbar item (b259de2), i think it would look better centered below it at a glance.
image
but the real issue is fixed.

PR #2425, opened 2024-07-09
Update Flathub badge

bragefuglseth:
See https://fosstodon.org/@bragefuglseth/112753049810610017


thibaultamartin (approved):
Thanks for helping us following Flathub's guidelines!

PR #2430, opened 2024-07-19
fix: top page horizontal scroll bar issue

konan-5:


MTRNord:
Hi :) thanks for your contribution. We are currently still at v0.18.0 of Zola meaning we still require the value to be `generate_feeds = true`. Could you also make sure to not commit the zola submodule please? :) It seems to be not complete anyway in its current form as it doesn't point anywhere.

MTRNord (changes requested):
See my other comment please :)

konan-5:
hello @MTRNord 
thank you for your reply
I rebased commit

konan-5:
Pushed again, please check :)

konan-5:
hello @MTRNord 
did you have a good weekend?
could you drop any message?

MTRNord (approved):

PR #2436, opened 2024-07-24
neochat: add technology

noraj:


thibaultamartin:
Hej thanks for contributing to matrix.org. We don't think this change is particularly needed nor brings any value to the website. If this is a strong requirement from the Neochat maintainers, can you ask them to endorse such a PR?

noraj:
1.) It was for homogeneity, examples from others clients specifying the technology used on https://matrix.org/ecosystem/clients/:

- Nheko: Desktop client for Matrix **using Qt and C++20.**
- Fractal: A Matrix messaging app **for GNOME, written in Rust**. Its interface is optimized for collaboration in large groups, such as free software projects, and will fit all screens, big or small.
- gomuks: A terminal Matrix client **written in Go.**
- gotktrix: Matrix client **in Go and GTK4.**
- iamb: A terminal-based Matrix client with Vim keybindings **written in Rust.**
- Watch The Matrix: A Matrix client for watchOS **written in Swift**
- Quaternion: **A Qt5-based** IM client for Matrix.

2.) I think this change is needed and brings any value: As a KDE user I'm looking for a Qt client because it will better integrates with my DE and bring less dependencies, also the must is if there is already a KDE based one for the best integration. So adding `based on Qt and KDE frameworks` is not for show but actually meaningful information that is a useful selection criteria.

Issue #2443, opened 2024-08-01
Consider adding a "Upcoming Client" Section next/under the featured section


MTRNord:
Many people seem to mistake featured for this. We might be able to do a compromise with this by offering like a "the next hot client according to the community" section.

It should be however clear that these may not be stable clients.

HarHarLinks:
I don't really get it?

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/2443#issue-2442683669):

> Many people seem to mistake featured for this. We might be able to do a compromise with this by offering like a "the next hot client according to the community" section.

Since it seems the reason for this addition is due to some people misunderstanding what "featured" means, should we add further clarification in regard to that?

What would the criteria be for this new category, for both assigning it to a client and removing it?

HarHarLinks:
I don't think it's a priority to add such a section, I don't know what we would currently put there. We can discuss how to make "featured" clearer - i.e. I believe for bridges we use the term "recommended".

MTRNord:
From the office hour on 15.1.26: We consider "Popular" might be a better wording. However, we likely need to make a policy first to do so. #1584

PR #2448, opened 2024-08-06
Update chatty information

agx:


MTRNord (approved):
Thank you!

agx:
Thanks for approving this so quickly! What are the requirements to make a client a "featured client" ? I was cautious to add that.

MTRNord:
> Thanks for approving this so quickly! What are the requirements to make a client a "featured client" ? I was cautious to add that.

Hi :) Currently there are no formal requirements defined, so this ends up being a decision between me and Thib as essentially the core editors of the website. There are plans to formalise this eventually, but our TODO list sadly doesn't have much of a priority for this currently. :/ So we take a rule of thumb in terms of "is maintained", "is usable by our grandparents" and "is visually pleasing". Many of these are subjective.

We also learned that clients which clearly feel unfinished or "Beta" are usually not appreciated by the community to be featured.

However, I had a thought as well that it might be interesting to list "upcoming" clients as semi-featured clients as well. https://github.com/matrix-org/matrix.org/issues/2443 there wasn't a deeper thought or talk about this however at this time.

agx:
Thanks for the details: maybe to clarify: I wasn't thinking that Chatty is ready by now. I was more checking for the criteria list (as I figured matrix upstream would have good idea about user expectations) so that we could work towards it.

MTRNord:
> Thanks for the details: maybe to clarify: I wasn't thinking that Chatty is ready by now. I was more checking for the criteria list (as I figured matrix upstream would have good idea about user expectations) so that we could work towards it.

hm yeah a checklist currently doesnt exist.

agx:
If any criteria come to mind maybe just dump them here? Not as a binary in or out list but just to get an idea which can maybe later on be formalized?

PR #2451, opened 2024-08-09
#2203 issue : Update skel.html with meta tags

AllMightLegend:


MTRNord (changes requested):
It seems like not only is the Skeleton duplicated but also the meta Tags being added already exist in the file. Therefor I am closing the pr :)

PR #2454, opened 2024-08-12
Be more specific about what correct linking is in the TWIM guide.

Gnuxie:
Signed-off-by: Gnuxie <Gnuxie@protonmail.com>


MTRNord (approved):
Thanks a lot!

PR #2461, opened 2024-08-16
Re-add code block that mysteriously disappeared

turt2live:


turt2live:
I think docs2markdown just failed to handle the element, fwiw

Issue #2197, opened 2024-02-20
Generate the content of the root ecosystem page


thibaultamartin:
https://github.com/matrix-org/matrix.org/blob/main/content/ecosystem/_index.md is plain text, we can generate the content (in a shortcode?) from the actual data instead

PR #2466, opened 2024-08-17
Issue: Generate the content of the root ecosystem page #2197

NischalPaliwal:
This is what I tried doing to solve after researching from the web ->
Step 1: Created a data file
I created a new file called matrix-projects.yaml in the data directory. This file will contain the data for the Matrix projects and tools. I added the data manually, but it can also be fetched from an external source. The file contains categories, and within each category, there are projects with their names, descriptions, and URLs.

Step 2: Created a shortcode
I created a new file called matrix-projects.html in the layouts/shortcodes directory. This file contains the template for rendering the Matrix projects and tools. The shortcode uses the data from the matrix-projects.yaml file and loops through each category, displaying the category name, and then loops through each project, displaying the project name, description, and URL.

Step 3: Used the shortcode
I replaced the manual list in the Markdown file with the shortcode {{< matrix-projects >}}. This will render the Matrix projects and tools dynamically from the data file.

Created a data directory ->
Since the data directory is not present by default in the Matrix.org repository, I created a new directory named data in the root of the repository, alongside the content, layouts, and static directories. This is where I stored the matrix-projects.yaml file.

-> I am totally a beginner to this project having a large codebase so if I have produced a wrong solution with wrong approaches to the issue I apologize for that !!!

### DIFF
--- a/content/ecosystem/_index.md
+++ b/content/ecosystem/_index.md
@@ -1,11 +1,1 @@
-+++
-title = "The Matrix Ecosystem"
-sort_by = "weight"
-+++
-
-* [Clients](clients) &ndash; The Matrix ecosystem is vibrant. Whatever platform you use, someone has probably already developed a client for it.
-* [Hosting](hosting) &ndash; Everyone can host their Matrix server, but not everybody wants to. These are the providers the Matrix.org Foundation knows about.
-* [Servers](servers) &ndash; Advanced users may want to run a homeserver by themselves for more independence and sovereignty. Here are the servers advanced users can run.
-* [Bridges](bridges) &ndash; Bridges allow you to connect Matrix to a third-party platform, and interact seamlessly.
-* [Integrations](integrations) &ndash; Useful integrations for more productive or fun conversations in the Matrix ecosystem.
-* [SDKs](sdks) &ndash; Develop great apps, bots, and bridges. Focus on what you do best. Let the SDKs do all the heavy lifting.
+{{< matrix-projects >}}
--- a/layouts/shortcodes/Matrix-projects.html
+++ b/layouts/shortcodes/Matrix-projects.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>The Matrix Ecosystem</title>
+</head>
+<body>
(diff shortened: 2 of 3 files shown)


content/ecosystem/_index.md
```diff
@@ -1,11 +1 @@
-+++
-title = "The Matrix Ecosystem"
-sort_by = "weight"
-+++
```
HarHarLinks, line 4: This does not compile without any preamble

content/ecosystem/_index.md
```diff
@@ -1,11 +1 @@
...
-
-* [Clients](clients) &ndash; The Matrix ecosystem is vibrant. Whatever platform you use, someone has probably already developed a client for it.
-* [Hosting](hosting) &ndash; Everyone can host their Matrix server, but not everybody wants to. These are the providers the Matrix.org Foundation knows about.
-* [Servers](servers) &ndash; Advanced users may want to run a homeserver by themselves for more independence and sovereignty. Here are the servers advanced users can run.
-* [Bridges](bridges) &ndash; Bridges allow you to connect Matrix to a third-party platform, and interact seamlessly.
-* [Integrations](integrations) &ndash; Useful integrations for more productive or fun conversations in the Matrix ecosystem.
-* [SDKs](sdks) &ndash; Develop great apps, bots, and bridges. Focus on what you do best. Let the SDKs do all the heavy lifting.
+{{< matrix-projects >}}
```
HarHarLinks, line 1: this is not working

HarHarLinks (changes requested):
Hello, thank you for your interest in contributing to the matrix.org website.

It seems you idea was to refactor this part such that the list of content is in a structured file and generating the page via templates. We keep such files containing structured data under /content, and we use toml format, like content/ecosystem/servers/servers.toml. However that is not the issue here.

The idea from #2197 was to generate this page *without* having to maintain its content twice. That is, the info is already available by the existence of content/ecosystem/clients/_index.md, content/ecosystem/servers/_index.md, etc. so instead of hardcoding a copy, we would like to reference this data instead.

HarHarLinks:
superseded by #2820

PR #2489, opened 2024-09-12
Rename responsible disclosure to CVD

Erethon:
Given how the industry has moved on from calling it "responsible" disclosure and even the wikipedia page has been renamed (https://en.wikipedia.org/wiki/Responsible_disclosure redirects to https://en.wikipedia.org/wiki/Coordinated_vulnerability_disclosure), I think it makes sense to switch this on the Matrix website too.


dkasak (approved):
Thank you for the PR! This was already our intention, we just didn't get to it yet.

PR #2497, opened 2024-09-16
Show all categories on blog.html

### LINKED ISSUE #2495: Not all categories shown on /blog page
The /blog page shows only one category of an entry, while the entry page shows multiple

HarHarLinks:
resolves #2495


MTRNord (approved):
seems to work as advertised

PR #2498, opened 2024-09-22
Add config options for the conference information banner and disable it since conf is over

MTRNord:


thibaultamartin (approved):
Surely going to be handy next year, thanks for the clean-up

PR #2500, opened 2024-09-22
Add Parsee to the XMPP bridge list

seija-amanojaku:


content/ecosystem/bridges/xmpp/bridges.toml
```diff
@@ -18,3 +18,24 @@ channel = true
...
+
+[[bridges]]
+name = "parsee"
+maintainer = "LDA"
+summary = """
+The jealous XMPP-Matrix bridge
+"""
+maturity = "Alpha"
+language = "C99"
+license = "AGPL-3.0"
```
tulir, line 30: I think this needs to have `-only` or `-or-later` as a suffix

tulir (approved):

PR #2510, opened 2024-09-26
Set width and height on the images in the footer to reduce the layout shift

### LINKED ISSUE #2463: Layout shift on the footer
The footer seems to cause significant layout shift due to the icons. Without checking I assume we did not set height and width when we should.

MTRNord:
Fixes #2463 

This changes nothing visual apart from causing the width and height being allocated before the css loaded which causes the footer to shift less vertically while images are loading.


thibaultamartin (approved):
Very good, thanks!

PR #2511, opened 2024-09-26
Add apple-touch-icons

### LINKED ISSUE #2482: `apple-touch-icon` is missing from markup
Currently, there’s no `link` tag for a good-resolution `apple-touch-icon` so the low-resolution favicon is picked up when bookmarking the site on iOS, which gives a poor look next to other proper bookmark icons, or no icon is shown at all.

The `apple-touch-icon`, when supplied, will also show up here:

IMG_2032

MTRNord:
Fixes #2482 

This is not tested as I dont own a device for testing this. This followed apples guidelines hence the weird sizes :shrug:


thibaultamartin:
We tested on an actual device, it needs some padding

thibaultamartin (changes requested):
Need the extra padding after testing on-device

PR #2515, opened 2024-09-27
Convert the twim image gallery to a shortcode and a small doc on the figures

MTRNord:
This essentially formalizes what I did in the twim post and also makes it work with mobile better.


MTRNord:
Turns out this solution only works for max 2 images. I tried various alternatives but none of them seem to work like they should. Need to spend another day of css magic on this I think.

PR #2516, opened 2024-09-28
adapt text to image info (room_join.svg)

flowolf:
before:
image

this PR fixes a text discrepancy to the image.


MTRNord (approved):
Thank you!

PR #2527, opened 2024-10-14
nheko: Mark Multi Account Feature as true

Integral-Tech:
According to [Nheko README](https://github.com/Nheko-Reborn/nheko), it supports `Creating separate profiles (command line only, use -p name)`.

### DIFF
--- a/content/ecosystem/clients/nheko.md
+++ b/content/ecosystem/clients/nheko.md
@@ -18,7 +18,7 @@ voip_1to1 = true
 voip_jitsi = false
 threads = true
 sso = true
-multi_account = false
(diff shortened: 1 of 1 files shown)


HarHarLinks:
Thank you for pointing this out. Afaik Element Desktop supports the same. Although technically correct to an extent, I don't think command line flags are what is meant by this feature when comparing to apps that have an account switcher in the UI where both accounts are logged in and running at the same time such as fractal or Tammy. To keep the list items readably short, I don't think we can afford to clarify this in place.

Integral-Tech:
Okay, I see

PR #2540, opened 2024-10-21
Add CI to lint markdown

MTRNord:
This uses the same thing as https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint so it should be compatible.

the line-length currently is not enforced as this would need manual fixing of almost all 1k files we have and we just have to do this slowly outside of this PR otherwise we will need weeks or months to do this.


README.md
```diff
@@ -18,3 +18,14 @@ zola serve
 
 Zola will build the website and start a web server, usually at
 <http://127.0.0.1:1111>
+
+## Running the markdown linter
+
+We are running the <https://github.com/DavidAnson/markdownlint-cli2>2> markdown linter on the website.
```
HarHarLinks, line 24: >2>2?

PR #2542, opened 2024-10-27
Add POSIX Shell Matrix SDK

HarHarLinks:
Signed-off-by: Kim Brose <2803622+HarHarLinks@users.noreply.github.com>

### DIFF
--- a/content/ecosystem/sdks/sdks.toml
+++ b/content/ecosystem/sdks/sdks.toml
@@ -660,3 +660,16 @@ featured_in = []
 description = """
 Matrix client SDK in C#
 """
+
+[[sdks]]
+name = "POSIX Shell Matrix SDK"
+maintainer = "wj25czxj47bu6q"
+maturity = "Beta"
+language = "POSIX Shell"
+licence = "MIT-Modern-Variant"
+repository = "https://gitlab.com/matrix-mellifluousness/sdk"
+purpose = ["client", "bot"]
+featured_in = []
+description = """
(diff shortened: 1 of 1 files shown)


content/ecosystem/sdks/sdks.toml
```diff
@@ -660,3 +660,16 @@ featured_in = []
 description = """
 Matrix client SDK in C#
 """
+
+[[sdks]]
+name = "POSIX Shell Matrix SDK"
+maintainer = "wj25czxj47bu6q"
+maturity = "Beta"
+language = "POSIX Shell"
+licence = "MIT-Modern-Variant"
```
MTRNord, line 669: This seems to not be the case according to https://gitlab.com/matrix-mellifluousness/sdk/-/blob/ec678e2a9a7011d8031a5654a7743cb7cb6245e5/LICENSE.txt which claims "all rights reserved". Which also to me reads like a non OSI license even though it might be based upon MIT.

MTRNord (changes requested):

content/ecosystem/sdks/sdks.toml
HarHarLinks: I got it from
image

thibaultamartin: The "All rights reserved" clashes with the rest of the license. I'd suggest to ask them to update their license so we can add it to m.org

thibaultamartin (changes requested):

thibaultamartin (approved):
Nevermind I just read the MIT modern variant.

PR #2555, opened 2024-11-09
Rename Docs to "Learn Matrix" and add it to the Try Matrix page

MTRNord:
This is a result of a discussion with @HarHarLinks .

The icon illustration is the Gnome UI Icon "document-text" which is CC Zero licensed and was modified to fit the same corner style like the other illustrations. However I would prefer a proper illustration here instead.

grafik


HarHarLinks:
I was comparing the m.org website's categories to where other projects might have "Learn more" and "Get started" sections. To me, "Docs" sounds too dry and not interesting to actually read, but more technical - when actually we want it to be end user facing docs that go a little beyond "getting started" (i.e. jumping in head first to try it out).

We should also clean the section's contents itself up. I find it does not fit the way the rest of the website is designed and the mix of "Matrix for chat" and "Spec guide for implementing authenticated media in the federation API" don't at all fit together.

thibaultamartin:
Point taken and I agree that Docs is a bit try. I'm wondering if "Learning Matrix" is really more engaging though.

Could something along the lines of "Get started" or "Use Matrix" make more sense?

templates/try-matrix.html
```diff
@@ -9,6 +9,26 @@ <h1>{{ page.title }}</h1>
 
 <div class="try_matrix">
     <div class="step-background">
+        <div class="step choose-a-client">
```
MTRNord, line 12: The class is probably wrong :D

thibaultamartin, line 12:
```suggestion
      <div class="step">
```

templates/try-matrix.html
```diff
@@ -9,6 +9,26 @@ <h1>{{ page.title }}</h1>
...
+                    If you're new to Matrix, we recommend starting with the <a
+                        href="/docs/chat_basics/matrix-for-im/">Matrix for Instant Messaging Guide</a>.
+                </p>
+                <div class="cta-wrapper">
+                    <a href="/docs" class="call-to-action">Learn Matrix</a>
+                    <a href="/docs/communities/getting-started/" class="call-to-action secondary">Build a Community on
+                        Matrix</a>
+                </div>
+            </div>
+            <img src="/assets/try-matrix/handbook.svg" alt="Artist view of an abstract matrix client." />
```
thibaultamartin, line 28: The alt text probably needs an update :)

templates/try-matrix.html
```diff
@@ -23,7 +43,7 @@ <h3>Choose a client</h3>
             <img src="/assets/try-matrix/client.svg" alt="Artist view of an abstract matrix client." />
         </div>
     </div>
-    <div class="step-background coloured">
+    <div class="step-background">
```
thibaultamartin, line 46: Not a blocker for the PR, but it'd be nice to use `:nth-child()` instead of manually adding/removing the class

https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#keyword_values

thibaultamartin (changes requested):
Left a few nits here and there, nothing too significant.

As an aside, the handbook svg doesn't look fantastic but in all fairness all the pictures on the existing page are not exactly great either.

We will have to request some Jordan time to get nicer illustrations, but that's not a blocker for this PR.

PR #2560, opened 2024-11-16
Add SchildiChat Next

SpiritCroc:
Element X is currently listed as a separate client to Element, so I guess it makes sense to also have a dedicated entry for SchildiChat Next.


MTRNord (approved):
Thank you!

Issue #2598, opened 2024-12-28
The client chooj's picture has incorrect URL


farooqkz:
The URL is `https://matrix.org/ecosystem/clients/https://chooj.app/screenshot.png` instead of `https://chooj.app/screenshot.png`

PR #2600, opened 2024-12-29
issue #2598

Vedant7789:
replaced the previous link with the new link


thibaultamartin:
This PR doesn't solve the problem. The problem has been solved in #2601

PR #2615, opened 2025-01-17
Add older hall of fame entry

davidegirardi:
Adds an older entry after getting an OK from the reporter.


thibaultamartin:
@davidegirardi just to remove any ambiguity, you already got the "OK" from the contributor?

davidegirardi:
Yes.

thibaultamartin (approved):

PR #2632, opened 2025-02-12
Add missing RSS feed, update zola

kilimnik:
I noticed that the templates include an `rss.xml` definition, though you are not using it. The default zola behaviour is to use the `atom.xml`, so I updated the `config.toml` to use both.

I also changed `generate_feed` to `generate_feeds` as this is the new option which should be used with the new version.

As this is a new zola feature, I also decided to update zola to the newest version.


MTRNord:
@thibaultamartin what version does cloudflare currently use? :) We obviously cant merge this if our cloudflare deployment doesnt support it yet.

kilimnik:
> @thibaultamartin what version does cloudflare currently use? :) We obviously cant merge this if our cloudflare deployment doesnt support it yet.

You are uploading the built html page to cloudflare. The zola build process is running inside the GitHub Actions

MTRNord:
> > @thibaultamartin what version does cloudflare currently use? :) We obviously cant merge this if our cloudflare deployment doesnt support it yet.
> 
> You are uploading the built html page to cloudflare. The zola build process is running inside the GitHub Actions

Its a little more difficult than that in reality since we also use the preview feature of cloudflare which is built on the side of cloudflare instead :)

MTRNord:
In the meanwhile 0.20.0 has been released. Could you update this PR to it? :) https://github.com/getzola/zola/releases/tag/v0.20.0

MTRNord:
(SRE was pinged we are looking into the cloudflare side of things still)

thibaultamartin:
This is https://github.com/element-hq/sre-internal/issues/3390 in SRE's internal issue tracker

HarHarLinks:
While zola was updated, the rss.xml is not there yet.

HarHarLinks (approved):

PR #2545, opened 2024-10-29
Replace all static image file formats with avif ones

MTRNord:
More file size optimizations:

grafik

It seems to have good coverage according to https://caniuse.com/avif I think.

PR #2636, opened 2025-02-13
Initial archive page

MTRNord:
Currently it shows the full content.

Lighthouse is obviously unhappy.
Async loading of images reduces this page from ~500MB of content to ~50MB. It technically is global though. However it is most likely fine.

#2545 Would (according to lighthouse) reduce it up to another ~30MB in transmitted size. So this would help.

There is no final decission yet if we want full posts or just links

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -25,6 +25,8 @@ taxonomies = [
 highlight_code = true
 highlight_theme = "visual-studio-dark"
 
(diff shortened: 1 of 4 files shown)


MTRNord:
This broke with 0.19+ in zola. We will not merge it at this time therefor

PR #2646, opened 2025-02-25
adding to and updating member list

bluesomewhere:
Recognizing Blender and iKono as new members, and updating Fractal Networks per rebranding to Hintjen.


HarHarLinks (approved):
Should we remove old unneeded logos?

MTRNord (approved):
Looks good. As @HarHarLinks says we likely want to remove the old image. The repo is already bigger than the linux kernel. Lets clean up where we can so the README command has a chance of making it small for newcomers :)

PR #2647, opened 2025-02-28
Fix well known CORS to allow looking up support info from browser clients

TheArcaneBrony:
Adds cross-origin headers to well-known records to support looking up support info and future records.

This change also includes the `server` record, as it's useful in the context of LibMatrix/RMU to correctly address a server when looking up implementation type (which enables some implementation-specific optimisations and features).

I can downgrade this to only cover client and support records if wanted.

### DIFF
--- a/static/_headers
+++ b/static/_headers
@@ -14,7 +14,7 @@
 
 
 # Allow ACAO for well-known client
-/.well-known/matrix/client
+/.well-known/matrix/*
   Access-Control-Allow-Origin: *
 
 # Add content type for jira archive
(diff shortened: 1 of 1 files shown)


static/_headers
```diff
@@ -14,7 +14,7 @@
 
 
 # Allow ACAO for well-known client
```
davidegirardi, line 16: We can go for `*` but update this comment too then.

davidegirardi (changes requested):

static/_headers
TheArcaneBrony: Sorry for the delay, I've updated the comment now.

davidegirardi (approved):

MTRNord (approved):

PR #2650, opened 2025-03-04
Mention Benevity in our donation page

thibaultamartin:


thibaultamartin:
Self check-marking since my fellow MTRNord is not available 🤝

PR #2668, opened 2025-03-28
fix funding.json morg-spec repositoryUrl.wellKnown location

HarHarLinks:

oopsie woopsie


MTRNord (approved):
Nvm it actually is good

PR #2677, opened 2025-04-08
add barkvr to ecosystem client list

zodiepupper:
hello, i've been working on my own matrix client for a couple years now and was thinking it might be a good time to look into adding to the ecosystem client list on the website. 

Signed-off-by: Zodie Woolsey <zodie@pupper.dev>


content/ecosystem/clients/barkvr.md
```diff
@@ -0,0 +1,29 @@
...
+voip_jitsi = false 
+threads = false 
+sso = false 
+multi_account = true 
+multi_language = false 
+[extra.packages] 
+windows_installer = "https://github.com/zodywoolsey/barkvr/releases/latest" 
+macos_installer = "https://github.com/zodywoolsey/barkvr/releases/latest"
+other_linux_link = "https://github.com/zodywoolsey/barkvr/releases/latest"
+other_android_link = "https://github.com/zodywoolsey/barkvr/releases/latest"
```
HarHarLinks, line 26: I'm not sure our template understands `other_android_link`, maybe compiling does something magical that I can't tell by reading the code at https://github.com/matrix-org/matrix.org/blob/main/templates/macros/clients.html. I can see how it would be useful though, I would welcome the contribution if you were interested in that.

zodywoolsey: yeah, i tested and they don't work but i left them in there because i was kinda encouraging my thought that every link field should have an platform-unbiased link. especially if a dev doesn't want to use a major distribution platform.

i looked through the code for the template already. i had a hard time finding where those fields could be added. i was initially looking for a way to figure out what all field names were supported. 

i will spend some more time on looking for how i can add more fields.

zodywoolsey: actually, i think i figured it out. i'm testing the new fields right now

zodywoolsey: what new fields should i add? just an other_platform_link for each platform?

zodywoolsey:
okay, i broke everything, please hold

content/ecosystem/clients/barkvr.md
zodywoolsey: okay, here is the PR to add other fields for iOS and android
https://github.com/matrix-org/matrix.org/pull/2690

zodiepupper:
is there anything I can do to get my app on the clients page please? I'm planning to release the actual alpha soon and would love it to be visible on the site if possible.

zodiepupper:
hello, i'm wondering if someone can please review this or tell me what to change. i would rather have the client listed over waiting for a semantic change i thought could be useful.

to be honest it's a bit painful to see so many other clients get added while i'm just waiting with no response.

weeman1337 (changes requested):
This PR now depends on another PR.

Can you just update this one to match the [content spec](https://github.com/matrix-org/matrix.org/blob/2fef1dd9325a3eb031b288efb7e896a83cf45af7/CONTENT.md?plain=1#L183)? i.e. remove the unknown fields.

Also please add a licence for the icon according to [the REUSE spec](https://reuse.software/).

zodiepupper:
okay, sorry for all the force pushes. i believe this should be sufficient, please let me know if not. thank you so much for responding!!!

content/ecosystem/clients/barkvr.md
```diff
@@ -0,0 +1,28 @@
...
+featured = false
+[extra.features]
+e2ee = false
+spaces = true
+voip_1to1 = true 
+voip_jitsi = false 
+threads = false 
+sso = false 
+multi_account = true 
+multi_language = false
```
awtj8o81ryywg793, line 21: Please update this section to bring it in line with recent changes to the required data needed for clients.

See: https://github.com/matrix-org/matrix.org/blob/main/CONTENT.md#clients

Currently, the website fails to build with the following errors:

```
Error: Failed to serve the site
Error: Failed to render page '/app/content/ecosystem/clients/barkvr.md'
Error: Reason: Failed to render 'ecosystem/client.html'
Error: Reason: Tried to iterate on a container (`features`) that has a unsupported type
```

content/ecosystem/clients/barkvr.md
```diff
@@ -0,0 +1,28 @@
...
+sso = false 
+multi_account = true 
+multi_language = false 
+[extra.packages] 
+windows_installer = "https://github.com/zodywoolsey/barkvr/releases/latest" 
+macos_installer = "https://github.com/zodywoolsey/barkvr/releases/latest"
+other_linux_link = "https://github.com/zodywoolsey/barkvr/releases/latest"
++++
+
+open source godot matrix client and xr collaborative creation tool!
```
awtj8o81ryywg793, line 28:
```suggestion
Open-source Godot Matrix client and XR collaborative creation tool!
```

awtj8o81ryywg793 (changes requested):
I’m not too sure where the best place for this project would be on matrix.org, maybe listed as a client, or in the yet-to-be-built “Other” section (see: https://github.com/matrix-org/matrix.org/issues/3539) that includes projects like Neoboard (Stand-alone), Cactus Comments, etc. We’ve yet to define what we mean by "client", and it may take a while to do so (see: https://github.com/matrix-org/matrix.org/issues/3592). However, here is a little feedback.

PR #2682, opened 2025-04-10
Adding a link to the merch shop to the donation page.

yncyrydybyl:


MTRNord (approved):

HarHarLinks (changes requested):
Thank you for your contribution. I'm not entirely comfortable with just dumping this button in the header, but there is certainly value in discussing this topic within the working group, so thanks for pointing it out.

I would say at a minimum the merch button should appear somewhere further down on the page as well e.g. around the other money donation options, if we accept calling merch sales support.

Generally I think it would be appropriate to mention the shop on this page. While it is specifically about monetary donations, wearing the merch is explicitly support also in the non-monetary sense: it is one of the reasons while we go to the effort of bringing it to conferences.

I don't think it is currently well-defined or -known how the merch sales benefits the Foundation, at least not to the website working group. I think it would be ideal, if we add the link to this page, to also add a paragraph explaining in what ways merch sales support the Foundation, as well as some other details about it, at least as long as these are not explained on the shop page itself. This could include for example: from where will my package be shipped? Do I need to expect having to deal with customs?

The shop site itself has a "Refund and Exchange" page at https://shop.matrix.org/refund-exchange which somewhat unexpectedly touches on some of these questions. Perhaps the website WG could improve that as well.

thibaultamartin:
> Thank you for your contribution. I'm not entirely comfortable with just dumping this button in the header, but there is certainly value in discussing this topic within the working group, so thanks for pointing it out.

The main issue I have with putting it so prominently is that people can feel that buying merch is helping the foundation significantly. In practice the Foundation doesn't have a significant markup on the merch, and doesn't earn it anywhere near as what people give it on Donorbox

HarHarLinks:
Superseded by #2725

PR #2690, opened 2025-04-15
added fields for clients to use for alternate links for android and ios along with new SVGs for both

zodywoolsey:
added fields for clients to use for alternate links for android and ios along with new SVGs for both

i was adding my matrix client to client list and found no way to add platform-independent links for where to download the client. 

like for android the only options were google play and fdroid

this just extends the `other_linux_link` idea to include `other_android_link` and `other_ios_link`


templates/macros/clients.html
```diff
@@ -1,8 +1,8 @@
 {% macro classes(client) %}
-{% if client.packages.apple_app_store.app_id and client.packages.apple_app_store.org %}
+{% if client.packages.other_ios_link or client.packages.apple_app_store.app_id and client.packages.apple_app_store.org %}
```
HarHarLinks, line 2: If you allow me to be so pedantic, I would love to order this consistently with how the other platforms work ;)

```suggestion
{% if client.packages.apple_app_store.app_id and client.packages.apple_app_store.org or client.packages.other_ios_link %}
```

HarHarLinks (changes requested):
Thank you for this contribution! The code looks good and works. The buttons look good to me, too, thanks for going the extra mile for that!

Sadly something iffy about the buttons or use of logos and text marks therein is about their licensing. Sorry, I only realised this after requesting this feature or I would have mentioned it. Could you tell us the sources for the pieces used, and if you happen to know them, the licenses? Then we can figure out how we can proceed there :)

zodywoolsey:
> Thank you for this contribution! The code looks good and works. The buttons look good to me, too, thanks for going the extra mile for that!
> 
> Sadly something iffy about the buttons or use of logos and text marks therein is about their licensing. Sorry, I only realised this after requesting this feature or I would have mentioned it. Could you tell us the sources for the pieces used, and if you happen to know them, the licenses? Then we can figure out how we can proceed there :)

the android assets are from here:
https://developer.android.com/distribute/marketing-tools/brand-guidelines
the source of the iOS icon asset was from the app store badge that was already present in the repo at:
https://github.com/matrix-org/matrix.org/blob/main/static/assets/stores/app_store.svg

lemme know if i should recreate the button graphics.

zodywoolsey:
ok, i pulled your suggested change and squashed the commits to keep it at only 1 commit since it's such a small feature change. lemme know if there is anything else you would like me to do here. 

^-^

zodywoolsey:
realized i didn't add these fields to the docs so i added them and i also added the `other_linux_link` option in the docs so it's more obvious that it exists.

zodywoolsey:
is there anything else i should do to get this merged? i think am satisfied with my changes at this point.

zodywoolsey:
i think i'm satisfied with these changes, what should i do to get it reviewed or ready for review?

HarHarLinks:
Sorry for letting you wait so long, the website maintainers are stuck on figuring out how to proceed: #2694

zodywoolsey:
i don't think thee conflict, right? the other PR is for showing the linux download button, but mine is just for mobile. i can revert any of my changes that conflict, if that makes it easier

zodiepupper:
hello, will this ever be reviewed? i would really like to get this blocker out of the way so i can list my client on the ecosystem clients page.

PR #2694, opened 2025-04-16
Render a button for other linux sources for ecosystem/clients

HarHarLinks:
Seems to have been an oversight?


static/assets/stores/download_for_linux.svg
HarHarLinks, line 1: idk, should this be a a triangle or something?

HarHarLinks: @thibaultamartin do we have a guideline/source how to create these?

thibaultamartin:
This wasn't an oversight but a deliberate decision to only either list flathub, since it's the go-to platform to distribute Linux apps on all distros, or to link to the website of the project where developers can expand on the additional options they have to support (native packages, appimages, etc).

We can't really have a single button "for Linux" since it can be a rpm, a deb, a link to flathub, an appimage, etc.

Are you happy to close this PR, or do you want to discuss it further?

HarHarLinks:
> I like the idea of a "Releases" link that leads to a web page where the maintainer can list options themselves 👍

How do you imagine this to look like? I think it should still be part of the download buttons, because that is consistent with the UX of "where do I get this thing".

static/assets/stores/download_for_linux.svg
TheOneWithTheBraid: Shouldn't it maybe be a Tux silhouette ?

thibaultamartin:
How about having a "Releases" button with a logo along the lines of https://www.svgrepo.com/svg/535480/list-ordered ?

MTRNord:
> How about having a "Releases" button with a logo along the lines of https://www.svgrepo.com/svg/535480/list-ordered ?

Maybe instead we should use a generic Download icon instead (based on discussions in the Office hour 22.01.26)

PR #2715, opened 2025-05-01
recognize infra.run and filament as members

bluesomewhere:


MTRNord (approved):
Hm filament has no light theme logo right? Feels like it would fit better. Otherwise seems fine :) Happy to have them!

bluesomewhere:
Thank you! I'll see if we can wrangle a different logo from them.

PR #2716, opened 2025-05-01
update terms of reference (slight adjustment to election timeline)

bluesomewhere:
Changes are limited to the election timeline – previously:

* nomination period was 1 week
* campaigning period was 3 weeks
* and dates were anchored to Anywhere on Earth "timezone"

Now:

* nomination period is 2 weeks
* campaigning period is 2 weeks
* and dates are anchored to UTC timezone

This emerged from feedback after the last election, was discussed in the Governing Board's Governance Committee, was raised to the whole Governing Board for visibility, and then was approved by the Guardians.

The rationale: folks don't need 3 weeks to campaign, but they may need more than 1 week to get their nomination in, especially since it's not uncommon for someone to be OOO for a week at a time. And the timezone change is because AOE was confusing for some and is not properly implemented in most election tooling.


MTRNord (changes requested):
A sentence or 2 noting that it changed and a TLDR diff would be amazing here I think?

At least here on the PR so we have it somewhere. Dont think it fits on the page itself.

bluesomewhere:
Ohh yes thank you. Updated my opening comment with a description of the changes, the process, and the rationale.

MTRNord (approved):
Happy with the updated PR description now :) Seems enough of a paper trail for me :)

PR #2720, opened 2025-05-02
fix nameclash in navbar template page variable

HarHarLinks:
the iterator variable `page` nameclashed with the page variable of the current page (https://www.getzola.org/documentation/templates/pages-sections/#page-variables: "Whichever template you decide to render, you will get a page variable in your template [...]"), which means from a page with the navigation disabled you would override the other individual pages' setting.

This led to the following bug on the elections page:
image


MTRNord (approved):
seems sensible I guess

PR #2725, opened 2025-05-05
Revamp support membership

thibaultamartin:


templates/support.html
```diff
@@ -5,211 +5,92 @@
...
-            <div class="cards">
-                {% for supporter in supporters.associate %}
-                <a href="{{ supporter.website }}" class="supporters-card">
-                    <img src="/support/{{ supporter.logo }}" alt="{{ supporter.name }}'s logo">
-                    <span>{{ supporter.name }}</span>
-                </a>
-                {% endfor %}
-            </div>
-        </div>
-    </div>
```
HarHarLinks, line 92: ftr this means the references from recent blog posts get "broken"

templates/support.html
```diff
@@ -5,211 +5,92 @@
...
                 </p>
             </div>
+            <a href="/membership" class="call-to-action">Become a member</a>
         </div>
     </div>
-    <div id="all-your-comms">
+
+    <div id="working-group">
         <div class="content">
+            <h2>Join a working group</h2>
```
HarHarLinks, line 60: maybe worth mentioning wrt volunteering, I think the SCT has also repeatedly said you can support it by writing spec PRs and stuff, but they also struggle to find these contributors.

HarHarLinks (commented):
I think I would like the /support headlines to be a bit more specific about who you are donating to where applicable. The navbar says "Support **us**".

templates/membership.html
HarHarLinks, line 1: there is an issue with the Membership tiers title alignment compared to the member constituencies

MTRNord (commented):
LGTM but the comments from kim make still sense to tackle I think (hence not yet a green tick but a neutral tick)

HarHarLinks:
> LGTM but the comments from kim make still sense to tackle I think (hence not yet a green tick but a neutral tick)

I closed some, the remainders I was at a loss how to implement. Suggestions?

MTRNord (approved):
After speaking to kim I think it makes sense to do the open comments at $future_date_here.

PR #2726, opened 2025-05-05
Navbar experiments?

MTRNord:
I got tired of the current navbar design and wanted more colour, so I came up with this:

[Bildschirmaufnahme_20250506_000628.webm](https://github.com/user-attachments/assets/f4999d86-e568-4c4a-81e0-ad62312316d9)

Mobile works just like before.

### DIFF
--- a/sass/_header.scss
+++ b/sass/_header.scss
@@ -75,26 +75,29 @@
             align-items: center;
         }
 
-        .nav-links:not(.primary) {
-            position: relative;
 
-            &::after {
-                content: "";
-                position: absolute;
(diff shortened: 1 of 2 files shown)


thibaultamartin:
I like the idea of adding more color to the matrix.org website (and brand in general) to keep a clean aesthetic but get past the dull B&W.

I would rather ensure we have consistent design/branding guidelines before we implement it though. I'm concerned about making small, inconsistent jarring changes here and there, instead of bringing a harmonious set of changes :)

PR #2732, opened 2025-05-10
Make floating images ultrawide screen freindly

Louis-Ladd:
On ultrawide monitors, the dynamic view width calculations for the images ended up causing them to overlay buttons which would cause half the button to stop working.

image

This better places the elements in a better looking place and doesn't affect the buttons functionally anymore.
Before:
image

After:
image

Signed-off-by: Louis Harshman <lewisharshman1@gmail.com>

### DIFF
--- a/sass/_index.scss
+++ b/sass/_index.scss
@@ -326,6 +326,10 @@
             top: 100px;
             left: calc(60dvw - 800px);
 
+            @media (min-width: 2000px) {
+                left: 20dvw;
+            }
+
             width: 240px;
             height: 287px;
         }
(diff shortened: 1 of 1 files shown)


HarHarLinks:
Hi, thanks for this contribution!

git blame says https://github.com/matrix-org/matrix.org/pull/2252/files seems to have introduced some bugs including this one. It's unclear to me where the hardcoded 800px come from.

I think we have a content max-width of 1000 or 1200px on this site, and I believe the intention is for these graphics to roughly stick to that, even on ultrawide. I played with it a bit on that basis, what do you think: https://harharlinks-homepage-wide-sc.matrix-website.pages.dev/

Louis-Ladd:
Looks good to me, It responds well and looks good on mobile, widescreen, and ultrawide.

HarHarLinks:
superseded by #2734

PR #2728, opened 2025-05-08
Fix Sidebar Overlap with Footer

sanjayy-gowdaa:
Resolves #2521 

Fixed the issue where the sidebar was being overlapped by the footer when scrolling to the bottom of the page.
Changes made :
file: _docs.scss
flex-grow: 1; to take up vertical space 
position: sticky; for the sidebar menu, Keeps the element in flow, sticks only when needed, and scrolls away if the container ends.
(.docs-body .docs-content  .docs-menu updated accordingly)
matrix_footer_pic

https://github.com/user-attachments/assets/432ce410-f234-4b5b-86db-439470757978

Issue #2521, opened 2024-10-05
scrolling vs footer


HarHarLinks:
https://github.com/user-attachments/assets/b60e86ef-b0f9-48ab-95f7-e26004ff2b90

MTRNord:
Which browser and resolution etc? :)

HarHarLinks:
Firefox, Arch Linux, 4K @2x scaling, using 2 fingers on a trackpad.

I think the real issue is that the footer goes over (z-index) the nav, instead of under (vertically).

Maybe this makes it clearer, Vivaldi	6.9.3425.3, Arch Linux, 1440p, using a mouse:

https://github.com/user-attachments/assets/b8b94bb3-0274-4497-a4c7-c920a46cd365

Ritesh051:
I have fixed this issue and I have made a PR also.

HarHarLinks:
fixed by #2878

PR #2736, opened 2025-05-11
Fix Sidebar Overlap with Footer

sanjayy-gowdaa:
Resolves (#2521 )
Continued from #2728

Fixed the issue where the sidebar was being overlapped by the footer when scrolling to the bottom of the page.

Thank you for the detailed feedback and for pointing out the formatting and workflow issues!

My previous patch included unrelated formatting changes, which made it harder to review. I'll follow your suggestion and resubmit a new PR with only the minimal, relevant changes to address the issue. I'll also ensure that .editorconfig is respected by using the recommended VS Code plugin and double-check formatting before committing.

matrix_footer_pic2


HarHarLinks:
Superseded by #2878

PR #2746, opened 2025-05-20
format md lists in 2025 nominee statements

HarHarLinks:

:tophat: Website & Content WG cleanup crew :broom:


HarHarLinks:
@bluesomewhere 2 candidates are missing platforms/bios. is that correct? can we put a note in saying that we are waiting for these candidates to hand theirs in later instead of having the website look broken?

bluesomewhere:
> @bluesomewhere 2 candidates are missing platforms/bios. is that correct? can we put a note in saying that we are waiting for these candidates to hand theirs in later instead of having the website look broken?

Yeah that's correct, they chose not to submit platforms or bios. I'm not currently expecting them to, but to your point it does make the page look broken... Perhaps we drop in a "None submitted." or "N/A"? 🤔

MTRNord (approved):

MTRNord (approved):

PR #2752, opened 2025-05-22
Lightweight archive page

thibaultamartin:
Lightweight alternative to #2636 to let people Ctrl-F through the post titles


HarHarLinks (changes requested):
issues for MVP:
- discoverabilty? how to reach this page?
- can the path be /blog/archive without making it a blog post?

longer term iterations could be:
- filters
- actual search incl contents?

PR #2755, opened 2025-05-22
drop Futurewei, re-alphabetize Silver Members

bluesomewhere:
update to membership list


HarHarLinks (approved):

bluesomewhere:
> oh wow we should definitely automate the sorting....

definitely 😂

Issue #2756, opened 2025-05-22
sort Foundation members automatically


HarHarLinks:
via #2755

we should automate the alphabetic display of the Foundation members

sanjayy-gowdaa:
Hello,

In response to the request to alphabetically display the Foundation members, I’ve implemented a script that programmatically sorts the entries in supporters.toml by name within each tier .

I would need some clarification on the preferred integration approach:

Should the sorting be performed manually as a one-time operation before committing changes?

Should it be automated as part of the build process?

Please advise on which option aligns best with the project's workflow.

HarHarLinks:
Thank you for your interest in this issue. I think the most reasonable solution would simply sort the entries when displayed, i.e. using the Zola Tera template, and not the actual data source. I expect this should be close to a one liner.

PR #2761, opened 2025-05-24
Add script to sort supporter entries by name

shaikmohammadsameer22:
This pull request introduces a Python utility script located in `scripts/sort_supporters.py` that sorts the entries in `content/support/supporters.toml` alphabetically by name within each tier (e.g., gold, silver).

Changes:
- Added `sort_supporters.py` script.
- Applied the sorting to `supporters.toml`.

Usage:
Run from the repo root:
    python scripts/sort_supporters.py content/support/supporters.toml

### DIFF
--- a/content/support/supporters.toml
+++ b/content/support/supporters.toml
@@ -172,3 +172,4 @@ logo = "gnome.svg"
 name = "KDE e.V."
 website = "https://kde.org"
 logo = "kde.svg"
+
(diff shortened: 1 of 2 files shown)


HarHarLinks:
Thank you for the contribution. This appears to target #2756.

I don't think a script and without automatic execution in CI for sorting the entries is the appropriate solution here, see also the clarifying comment on #2756. Since #2763 already implements this approach, I am closing this PR in favor of #2763.

PR #2763, opened 2025-05-24
Sort foundation members alphabetically in display

sanjayy-gowdaa:
Solves Issue (#2756 )
This PR updates the Tera template to sort Foundation members alphabetically by name within each tier (Platinum, Gold, Silver, etc.) during rendering. The supporters.toml file remains unchanged.

Thank you for the clarification! I've implemented the sorting directly in the Tera template as suggested. Let me know if any changes are needed.

### DIFF
--- a/templates/macros/supporters.html
+++ b/templates/macros/supporters.html
@@ -1,0 +1,25 @@
+{% macro case_insensitive_sort_supporters(supporters) %}
+    {# Step 1: Extract all supporter names in lower case #}
+    {% set lower_names = [] %}
+    {% for supporter in supporters %}
+        {% set_global lower_names = lower_names | concat(with=[supporter.name | lower]) %}
+    {% endfor %}
+
+    {# Step 2: Sort lower-case names #}
+    {% set sorted_lower_names = lower_names | sort %}
+
+    {# Step 3: Re-match sorted names with original supporter objects #}
+    {% for lower_name in sorted_lower_names %}
+        {% for supporter in supporters %}
+            {% if supporter.name | lower == lower_name %}
+                {% set_global current_supporter = supporter %}
+                {% break %}
+            {% endif %}
+        {% endfor %}
+
+        <a href="{{ current_supporter.website }}" target="_blank" class="supporters-card">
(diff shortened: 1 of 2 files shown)


sanjayy-gowdaa:
Thank you for the suggestion!

I attempted to implement case-insensitive sorting in the Tera template to match the current manual ordering (ignoring capitalization). Here are the approaches I tried:

**Attempt 1:**
{% for supporter in supporters.silver | sort(attribute="name" | lower) %}

Just applies lower to the string "name", not the actual values.
So it gets ignored and ends up doing a normal case-sensitive sort.

**Attempt 2**: Inline dictionary creation with concat

{# Prepare a list for sorting Silver supporters #}
                {% set silver_to_sort = [] %}
                {% if supporters.silver %}
                    {% for supporter_item in supporters.silver %}
                        {% set silver_to_sort = silver_to_sort | concat(with=[{"sort_name": supporter_item.name | lower, "data": supporter_item }]) %}
                    {% endfor %}
                {% endif %}
                {% set sorted_silver = silver_to_sort | sort(attribute="sort_name") %}

                {% for item in sorted_silver %}
                    {% set supporter = item.data %} {# Get original supporter data #}
                    <a href="{{ supporter.website }}" target="_blank" class="supporters-card">
                        {{ supporter.name }}'s logo
                        <span>{{ supporter.name }}</span>
                    </a>
                {% endfor %}

Error: expected a value that can be negated

**Attempt 3**: Create a dictionary using set
{% set sort_item = {
    "sort_name": supporter_item.name | lower,
    "data": supporter_item
} %}
Error:expected a value that can be negated or an array of values

Tera doesn’t support setting dictionaries directly inside templates.

What would be the best way to proceed?

sanjayy-gowdaa:
Resolves(#2756 )

Thanks for the detailed breakdown! This turned out to be a great learning opportunity  I gained a much better understanding of Tera’s limitations and some creative ways to work around them, especially regarding sorting and macro structuring.

I’ve taken your suggestion and implemented the workaround as a macro in /templates/macros/supporters.html, inspired by the structure in nominee.html. I also refactored the main support.html file to use the macro.

HarHarLinks (approved):
Very nice, thanks again!

PR #2767, opened 2025-05-26
Advertise MAS through the old MSC2965 discovery mechanism

sandhose:
Turns out, some clients (like Element iOS and Android legacy) still use the old way of discovering MAS for linking some account management features, and we forgot to update the well-known when we deployed MAS on matrix.org.

### DIFF
--- a/static/.well-known/matrix/client
+++ b/static/.well-known/matrix/client
@@ -5,6 +5,10 @@
     "m.identity_server": {
         "base_url": "https://vector.im"
     },
+    "org.matrix.msc2965.authentication": {
+        "issuer": "https://account.matrix.org/",
+        "account": "https://account.matrix.org/account/"
+    },
     "org.matrix.msc4143.rtc_foci": [
         {
             "type": "livekit",


HarHarLinks (commented):
Seems sound and matching the spec and real paths. I'm not enterely clear on: this doesn't appear to be a stable endpoint but only a possible alternative to the MSC proper that was presented as part of the MSC?
I.e. is this for supporting only element specifically? That feels a bit odd. Maybe I missed something skimming the MSC?

That said I would expect a proper review to be done by SRE.

HarHarLinks:
@sandhose I see, makes sense. I suppose this is OK as a temporary measure. For context, [previous rulings](https://github.com/matrix-org/matrix.org/pull/2494#issuecomment-2355235778) on similar situations were that such things may be OK to do given a clear sunset of the bandaid solution. Aiui there is a policy to not deploy unstable features to m.org where possible to prevent a shadow-spec. I wonder if we should apply the same approach here.

thibaultamartin:
After talking to the Website WG, Quentin and SREs, here are the main points considered.

- The change is not really content but happens to be hosted on the matrix.org repository. As such, the Website WG is not the best placed to approve or deny this request.
- The change suggested doesn't add any maintenance burden or complexity.
- It fixes account deactivation issues on versions of Element, dubbed Element Classic, that are still actively distributed in the app stores and used by people.
- Element Classic does not support the current stable endpoints offered by MAS, and deploying this json file advertises them to Element Classic. This is a vendor-specific solution to help Element during a transition phase, while they either update Element Classic to reach the stable endpoints or sunset Element Classic.
- Given the impact on users of matrix.org who use Element Classic, the Foundation finds it is in its interest to accommodate Element while it works on getting Element Classic to match the stable version of the spec or sunsets Element Classic.
- There is a concern that this creates a deviation from the stable specification. The Website WG would like to revisit this change in 6 months.

thibaultamartin (approved):

PR #2768, opened 2025-05-26
Remove element wellknown for registration_helper_url

HarHarLinks:
:tophat: Website & Content WG

Per the removal deadline set by https://github.com/matrix-org/matrix.org/pull/2494#issuecomment-2355235778, I assume this helper can and should have been removed as matrix.org migrated to MAS.

@ara4n could you please merge or provide an update if that's not yet possible? :)


ara4n:
thanks - yup, should have remembered this when we deployed MAS.

Issue #2688, opened 2025-04-14
Some <h>eadings are smaller than regular text


HarHarLinks:
![Image](https://github.com/user-attachments/assets/d9e13e11-c36f-418f-b2d6-6273d95b654e)

anushkasomani:
Can I work on this issue?
ETA : 2-3h

HarHarLinks:
Feel free to just open a PR and linking it here. Please be comprehensive and consistent about styling all heading levels. I can't tell you which size it should be, except probably not smaller than regular text.

ChehakTrehan:
Hii @HarHarLinks I fixed this issue. Please review my PR
thanks
Chehak

ChehakTrehan:
Hii @HarHarLinks Please review my PR
Thanks 
Chehak

PR #2769, opened 2025-05-29
2688 : Fixed the heading 

ChehakTrehan:
As Some <h> headings were smaller than regular text. So i fixed it.

Thanks
Chehak


HarHarLinks:
Hello, thanks for your contribution!

While I think the Atom feed links are probably not supposed to be headers at all, nor should we style text using the `<small>` tags, this PR is not comprehensive across the whole site. It is also not what #2688 is about; it is about CSS of `<h5>` and when adjusted also consistency with the other levels `h1` through `h6`.

If you would like to tackle either issue, please open a new PR. :)

ChehakTrehan:
@HarHarLinks I apologize as I misunderstood the issue. As I am new to open source, I faced some challenges with the process. I will create and submit a new Pull Request shortly

PR #2788, opened 2025-06-19
fix a11y of the figure shortcode

HarHarLinks:
🎩 Website & Content WG

I have been advised that it is counter productive to double up alt and caption as screen readers will already read the caption anyway. This PR gives us the option to add alt in cases where we don't want to include a caption.

### DIFF
--- a/templates/shortcodes/figure.html
+++ b/templates/shortcodes/figure.html
@@ -1,4 +1,4 @@
 <figure style="height:100%;">
-    <img src="{{ img }}" alt="{{ caption }}" />
+    <img src="{{ img }}" {% if alt %}alt="{{ alt }}{% endif %}" />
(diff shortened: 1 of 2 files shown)


templates/shortcodes/figure.html
```diff
@@ -1,4 +1,4 @@
 <figure style="height:100%;">
-    <img src="{{ img }}" alt="{{ caption }}" />
+    <img src="{{ img }}" {% if alt %}alt="{{ alt }}{% endif %}" />
```
thibaultamartin, line 2: Should we test for the absence of caption here?

HarHarLinks: Alt should *never* be caption because then it would be read double by screen readers

thibaultamartin (approved):

PR #2808, opened 2025-07-03
make link between Element and the rust+js sdks clear

neilisfragile:
See https://github.com/matrix-org/matrix-rust-sdk/pull/5341 and https://github.com/matrix-org/matrix-js-sdk/pull/4901


content/ecosystem/sdks/sdks.toml
```diff
@@ -13,7 +13,7 @@ Matrix Application Service framework in Node.js.
 
 [[sdks]]
 name = "Matrix.org JS SDK"
```
HarHarLinks, line 15: (huh, first time I notice it being called this)

PR #2814, opened 2025-07-11
Separate Element "legacy" mobile client to highlight its becoming a classic.

mxandreas:
Submitting this as the member of the Element team.

- Check for common mistakes:
  - [x] Wrap plain URLs in `<>` to linkify them ([learn more](https://github.com/matrix-org/matrix.org/blob/main/CONTENT.md#publishing-to-the-blog)).
  - [x] Use the right level of headings: The page title will use a level 1 headings, so your headings should use level 2 and below.
  - [x] Use internal links: when linking to another page on <https://matrix.org>, use absolute `[label](/target)` or relative `[label](subpage)` links.
- For blog posts:
  - [x] Verify the date and post ordering on the `/blog` page, especially for multiple posts on the same day. Prefer UTC format, e.g. `2025-12-01T14:00:00Z` for Dec 1st, 2025, 2pm UTC.
- [ ] Your individual commits or pull request is [signed off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md).


content/ecosystem/clients/element-classic.md
```diff
@@ -0,0 +1,28 @@
...
+template = "ecosystem/client.html"
+[extra]
+thumbnail = "element-classic.svg"
+maintainer = "Element"
+maturity = "Stable"
+repo = "https://github.com/element-hq"
+matrix_room = "#element-ios:matrix.org"
+licence = "Apache-2.0"
+latest_release = "2022-11-02"
+featured = true
```
HarHarLinks: I'm concerned about the perceived neutrality of the Foundation having 3 out of 6 of the featured clients be from Element. My feeling is that actually Element recommends Element X over classic, so perhaps we can un-feature classic?

content/ecosystem/clients/element-x.md
```diff
@@ -24,4 +24,4 @@ google_play_store.app_id = "io.element.android.x"
 apple_app_store = { app_id = "id1631335820", org = "vector" }
```
HarHarLinks, line 26: Would you mind adding the f-droid package that EXA is offering?

content/ecosystem/clients/element-classic.md
mxandreas: Fair point, I will remove it from the list.

mxandreas:
@HarHarLinks 

> I noticed that compared to https://github.com/matrix-org/matrix.org/pull/2813, Element X is not renamed in this PR, despite marking it as stable. Is this intended?

Yes, correct. We'll do the actual rename of the app a bit later (e.g. within a few months) for a few good reasons.

I have done the tweaking:
* Added F-Droid link for EX.
* Removed the classic altogether.

HarHarLinks:
> @HarHarLinks
> 
> > I noticed that compared to #2813, Element X is not renamed in this PR, despite marking it as stable. Is this intended?
> 
> Yes, correct. We'll do the actual rename of the app a bit later (e.g. within a few months) for a few good reasons.

Okay, sounds fine for me.

> I have done the tweaking:
> 
>     * Added F-Droid link for EX.

Awesome, thank you!

>     * Removed the classic altogether.

I would be lying if I said this didn't surprise me. As long as Element classic is maintained at all, I think it makes sense to have it on our list. Of course, being listed at all is in the end up to the app manufacturer. What I meant earlier was only to toggle the `featured = ` property to `false`, which means it still appears on the page, but only in the lower section that contains all clients, and not anymore among the "featured" clients at the top of the list.

In case this is unclear: you can preview your PR locally using the method described in the README, or our CI (https://decorate-mobile-legacy.matrix-website.pages.dev/ecosystem/clients/).

mxandreas:
> What I meant earlier was only to toggle the featured =  property to false, which means it still appears on the page, but only in the lower section that contains all clients, and not anymore among the "featured" clients at the top of the list.

Oh, I did not understand that there was this in between option. It is likely a better fit for the time being. I'll do the updates.

mxandreas:
Ok, should be good now. Element Classic appears but not in the featured top section. Thanks.

PR #2833, opened 2025-07-19
feat: include < polycule > into the client list

TheOneWithTheBraid:
I feel like now that we became an ecosystem foundation member, it might slowly be time to list < polycule > on the client list.

I'm not sure how to deal with the following change of the metadata : < polycule > targets Linux but only uses conventional packaging rather than Flathub. Is there any way to propagate Linux support without being on Flathub ? It'd be highly confusing since < polycule > *mainly* targets Linux.

*This PR is contributed on behalf of < polycule > as foundation ecosystem member.*

### DIFF
--- a/content/ecosystem/clients/polycule.md
+++ b/content/ecosystem/clients/polycule.md
@@ -1,0 +1,29 @@
++++
+title = "< polycule >"
+template = "ecosystem/client.html"
+[extra]
+thumbnail = "polycule.svg"
+maintainer = "The one with the braid"
+licence = "EUPL-1.2"
+language = "Flutter"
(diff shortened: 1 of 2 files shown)


HarHarLinks:
> Is there any way to propagate Linux support without being on Flathub ?

There is this yet unresolved discussion about it: #2694

TheOneWithTheBraid:
Okay, I removed the Flathub link in favor of the `other_linux_link`. Correctly shows `Linux` as platform but obviously there's no download shield before #2694 yet. From my side, the PR is therefore ready.

HarHarLinks (approved):
nice, thanks!

PR #2895, opened 2025-08-27
fix: use type scale for headline sizes

### LINKED ISSUE #2688: Some <h>eadings are smaller than regular text

pReya:
closes #2688 

The old values from h1 to h4 were using a scaling factor of ~1.25 (major third), but somehow h5 was completely out of line. I've updated the values to match a major third scale and corrected the h5 value. h6 is still not usable (this would require for everything to become much bigger, or choose a different scaling factor).

- Check for common mistakes:
- For blog posts:
- [ ] Let us know if you are contributing in a specific role, such as on behalf of an organisation or team, [for example](https://github.com/matrix-org/matrix.org/pull/2788).
- [ ] Your individual commits or pull request is [signed off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md).

### DIFF
--- a/sass/_fonts.scss
+++ b/sass/_fonts.scss
@@ -31,33 +31,30 @@ h5 {
     line-height: 1.3;
 }
 
-// These headlines are using a "minor third" type scale (scaling factor is 1.2)
+// These headlines are using a "major third" type scale (scaling factor is 1.25) - h6 is not usable
 h1 {
     margin-top: 0;
-    font-size: 2.986rem;
+    font-size: 3.0517rem;
 }
 
 h2 {
-    font-size: 2.488rem;
+    font-size: 2.4414rem;
     margin-block: 2rem .5rem;
 }
 
 h3 {
-    font-size: 2.074rem;
+    font-size: 1.9531rem;
     margin-block: 1.5rem .5rem;
 }
 
 h4 {
-    font-size: 1.728rem;
+    font-size: 1.5625rem;
 }
 
 h5 {
-    font-size: 1.44rem;
+    font-size: 1.25rem;
 }
 
-h6 {
-    font-size: 1.2rem;
-}
 
 h2+h3 {
     margin-block-start: 1rem;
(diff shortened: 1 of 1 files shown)


HarHarLinks:
This is how the legal pages (prominently using h5) look with this PR:
| before | after |
|---|---|
| image | image |

And the blog (which uses lots of h1s and h2s):
| before | after |
|---|---|
| image | image |

So I think this looks reasonable? I don't really have a reference of what is best practice.

Here is a random naive experiment of simply bumping the h5 on current main from 0.8 to 1:
image

Hm.

I would tend to accept the PR as-is, but would also love any references to learn a bit more about this topic. @thibaultamartin any thoughts?

pReya:
> Here is a random naive experiment of simply bumping the h5 on current main from 0.8 to 1:

Typographically this is considered "bad style". Headlines should always be larger than body text.

MTRNord:
Hm, I think the one sticking out to me on the comparison given by @HarHarLinks  is the "dept of spec "header. It feels fairly large. I am less annoyed by the h1 actually as that one seems fine. Maybe this is just a thing I got used to, though?

I wonder what other blogs commonly use. Thinking Mozilla or GitHub perhaps as ones where I like their sizing.

I also wonder if possibly another approach here would be to "clamp" h5+ to h4 level size. So essentially making nothing smaller than h4.

pReya:
I see 3 different options how to proceed:

1. Have 5 visually distinctive headline levels with scaling factor 1.25 (this is the current state of this PR)
2. Have 5 visually distinctive headline levels with scaling factor 1.2 (would make the big headlines a little smaller)
3. Have 4 visually distinctive headline levels with scaling factor 1.25 and make "higher" levels never smaller than h4 (suggestion by @MTRNord)

My personal preference would be 2. But someone from the team should make a decision here.

MTRNord:
So we had finally time to have a talk and decided 2 works best for us @pReya 

If you want to you can update the PR otherwise we would do it.

pReya:
> So we had finally time to have a talk and decided 2 works best for us @pReya
> 
> If you want to you can update the PR otherwise we would do it.

Done

PR #2897, opened 2025-08-29
Update How It Works to match v3 sync instead and fix codeblock styles.

MTRNord:
| Before | After |
|-----------|---------|
| grafik | grafik |

- Check for common mistakes:
- For blog posts:
- [ ] Let us know if you are contributing in a specific role, such as on behalf of an organisation or team, [for example](https://github.com/matrix-org/matrix.org/pull/2788).


sass/_howitworks.scss
```diff
@@ -18,9 +18,17 @@
...
-.legend pre {
-    font-size: 14px;
-    white-space: pre-wrap;
+#legendBox {
+    min-width: 0;
+}
+
+.legend {
+
+    pre {
```
HarHarLinks, line 27: why isn't the code block style centrally defined for the whole site?

templates/shortcodes/howitworks.html
```diff
@@ -62,26 +62,35 @@
...
+                            "origin_server_ts": 1417731086797,
+                            "sender": "@alice:alice.com"
+                        }
+                    ]
+                    ...
+                }
+            }
+        }
+    }
+    ...
```
HarHarLinks: You are of course technically correct, but I think taking some stylistic liberty to compact this horizontally for readability seems warranted here. 2 spaces instead of 4 already go quite a ways. I find it hard to argue for 4 spaces when half of my big screen is whitespace because we limit content to 1000px or so width. wdyt?

sass/_howitworks.scss
MTRNord: because I fear it may do weird stuff with the markdown generated code blocks which are like this:

```
<pre style="background-color: #1e1e1e;color: #dcdcdc;"><code>
<span>This is a single line in the code block</span>
<span>This is another</span>
</code></pre>
```

Including the weird formatting. Making it readable breaks the rendered version too. So this would be matched too and I just dont want things to explode here.

templates/shortcodes/howitworks.html
MTRNord: Fair enough. I can update that.

sass/_howitworks.scss
HarHarLinks: I see: because this is html in the template instead of markdown-generated here, and we don't currently use https://www.getzola.org/documentation/content/syntax-highlighting/#inline-vs-classed-highlighting we cannot easily apply that.

templates/shortcodes/howitworks.html
```diff
@@ -62,26 +64,35 @@
         </div>
         <div class="legend hidden" id="legend6">
             Destination clients receive Alice's message with a long-lived GET request.
-            (Clients are free to implement more efficient transports than polling as desired).
             <pre>
-curl "https://matrix.bob.com/_matrix/client/v2/sync?access_token=ACCESS_TOKEN"
+curl "https://matrix.bob.com/_matrix/client/v3/sync?access_token=ACCESS_TOKEN"
```
HarHarLinks: in terms of outdated: token by get is deprecated too https://spec.matrix.org/v1.15/client-server-api/#using-access-tokens

MTRNord: good catch. I totally forgot...

templates/shortcodes/howitworks.html
```diff
@@ -62,26 +64,40 @@
...
             Destination clients receive Alice's message with a long-lived GET request.
-            (Clients are free to implement more efficient transports than polling as desired).
             <pre>
-curl "https://matrix.bob.com/_matrix/client/v2/sync?access_token=ACCESS_TOKEN"
+curl "https://matrix.bob.com/_matrix/client/v3/sync" \
+    -H "Authorization: Bearer ACCESS_TOKEN"
+</pre>
+
+            Response:
+            <pre>
```
HarHarLinks: This breaks:
image
And it's also inconsistent with the other code blocks now that don't separate request from response.

MTRNord: Yup welcome markdown parser... the empty line does that. Fixed in [a45676b](https://github.com/matrix-org/matrix.org/pull/2897/commits/a45676bfe92234f942f35d0e1f0d1756b65c2676)

HarHarLinks (approved):

PR #2902, opened 2025-09-03
Add status page to footer

### LINKED ISSUE #2901: Add link from Matrix.org home page to https://status.matrix.org/
My Matrix web client went haywire (told me I was disconnected from some server but not which one, maybe I should file an issue against Element).  My browser's debugger console had multiple complaints about state synchronization (```"MatrixClient sync state => ERROR"```).  I tried various URLs scraped from the debugger console and got 429 errors and other unhelpful results.

I looked on downdetector.com (which doesn't track Matrix.org) and on Matrix.org and didn't see any outage announcements.  So I tried dumping cookies and local storage for element.io and logging in again... which strongly indicated something was wrong with the Matrix.org homeserver.

Then I tried a web search and came across https://status.matrix.org/ which makes it clear that database recovery of the Matrix.org homeserver is in progress.  When I first looked at the Matrix.org web page I was surprised not to see a link to a system-status page, but at that point I had figured maybe there just wasn't one.

The request: since there **_is_** a system-status site, it would be great if there were a link to it from the Matrix.org page.  I am willing to try a PR to the website repo, but the repo README strongly cautions against unsolicited PRs, hence this filing.

HarHarLinks:
:tophat: Website & Content WG

By evaluating I ended up effectively solving this already so here goes.
This PR changes the footer to use a flexbox instead of grid so we can center any remainder items easily. It then adds the status page to the footer so it is easy to find.

Resolves #2901

- Check for common mistakes:
- For blog posts:
- [x] Mention any issues related to the PR. Use [GitHub keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) as appropriate.

### DIFF
--- a/navigation.toml
+++ b/navigation.toml
@@ -18,7 +18,7 @@ footer_internal = [
     { title = "Legal", href = "/legal" },
     { title = "Law Enforcement Guidelines", href = "/legal/law-enforcement-guidelines" },
     { title = "Contact", href = "/contact" },
-    { title = "Status", href = "https://status.matrix.org" },
+    { title = "Matrix.org Status", href = "https://status.matrix.org" },
     { title = "Site Source", href = "https://github.com/matrix-org/matrix.org/" },
 ]
 


navigation.toml
```diff
@@ -18,6 +18,7 @@ footer_internal = [
     { title = "Legal", href = "/legal" },
     { title = "Law Enforcement Guidelines", href = "/legal/law-enforcement-guidelines" },
     { title = "Contact", href = "/contact" },
+    { title = "Status", href = "https://status.matrix.org" },
```
turt2live: To make it clear this isn't status for the whole federation:
```suggestion
    { title = "Homeserver Status", href = "https://status.matrix.org" },
```

turt2live (commented):
otherwise lgtm - I haven't verified the CSS

navigation.toml
HarHarLinks: But it is also *not only* the status for only the homeserver (see ticket)

HarHarLinks: On that note, perhaps the status page should use a clearer title than "[Matrix - An open network for secure, decentralized communication](https://matrix.org/)" or even a description if possible?

de0u: I was about to suggest that adding a link from the status page back to matrix.org might clarify what the status page is the status of... but now I understand that the page title **_is_** such a link.  However, it doesn't look like one, so I would never have thought to click on it.

Maybe it would be enough to just change the word "Matrix" in the status page's title to "Matrix.org"?  That way even if the title doesn't look clickable people will feel they know what to manually type?

turt2live: I feel we need to make the clarification in both places.

thibaultamartin:
```suggestion
    { title = "Matrix.org Status", href = "https://status.matrix.org" },
```

Issue #2913, opened 2025-09-18
Document success stories


MTRNord:
We might want to document success stories in the matrix ecosystem.

Current possible ideas are:

- Frontpage
- Ecosystem
- Blog post series

There has not yet been a decision on which of these is best.

yncyrydybyl:
This should include:
### successful deployments 
starting points are: 
* https://element.io/en/matrix-in-germany/projects
* https://www.fairkom.eu/en/references-and-partners
* https://element.io/en/matrix-in-germany/projects/universities
* https://servers.joinmatrix.org/
* https://en.wikipedia.org/wiki/Matrix_%28protocol%29#Adoption
### successful products
* some TIM-Products (famedly, Barmer, AOK, etc)
* products which have integrated matrix (beeper, element pro)

PR #2907, opened 2025-09-06
Add distributions page

HarHarLinks:
:tophat: Website & Content WG

Fixes #2068. Based on #2182.

Signed-off-by: HarHarLinks <2803622+HarHarLinks@users.noreply.github.com>

**:heavy_check_mark: Checklist**

- Check for common mistakes:
  - [x] Wrap plain URLs in `<>` to linkify them ([learn more](https://github.com/matrix-org/matrix.org/blob/main/CONTENT.md#publishing-to-the-blog)).
  - [x] Use the right level of headings: The page title will use a level 1 headings, so your headings should use level 2 and below.
  - [x] Use internal links: when linking to another page on <https://matrix.org>, use the Zola `[label](@/target.md)` [syntax](https://www.getzola.org/documentation/content/linking/#internal-links).
- For blog posts:
  - [ ] Verify the date and post ordering on the `/blog` page, especially for multiple posts on the same day. Prefer UTC format, e.g. `2025-12-01T14:00:00Z` for Dec 1st, 2025, 2pm UTC.
  - [ ] Set the correct author and category. Browse existing ones at <https://matrix.org/author/> and <https://matrix.org/category/> to match them.
- [x] Let us know if you are contributing in a specific role, such as on behalf of an organisation or team, [for example](https://github.com/matrix-org/matrix.org/pull/2788).
- [ ] Let us know if your PR is time-sensitive in any way.
- [x] Mention any issues related to the PR. Use [GitHub keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) as appropriate.
- [x] Your individual commits or pull request is [signed off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md).

PR #2122, opened 2024-01-03
Updating Federated Host Information

rossthemusicteacher:
Updating our hosting information to include a new link, and updated description.

PR #2915, opened 2025-09-19
Add in details of the Synapse Pro homeserver

neilisfragile:
Adding in Synapse Pro since it is at this point distinct to Synapse and important to the eco-system since it backs the matrix.org homeserver as well as multiple large scale and high profile government installations. 

Separately, I think we need a deterministic ordering for projects for all eco-system pages and I would support the approach Thib outlines [here](https://github.com/matrix-org/matrix.org/issues/1584), but for now, adding as is. 

Signed-off-by: Neil Johnson [neil@element.io](mailto:neilj@element.io)

- Check for common mistakes:
  - [x] Wrap plain URLs in `<>` to linkify them ([learn more](https://github.com/matrix-org/matrix.org/blob/main/CONTENT.md#publishing-to-the-blog)).
  - [x] Use the right level of headings: The page title will use a level 1 headings, so your headings should use level 2 and below.
  - [x] Use internal links: when linking to another page on <https://matrix.org>, use the Zola `[label](@/target.md)` [syntax](https://www.getzola.org/documentation/content/linking/#internal-links).
- For blog posts:
- [x] Mention any issues related to the PR. Use [GitHub keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) as appropriate.

### DIFF
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -19,6 +19,17 @@ repository = "https://github.com/matrix-construct/tuwunel"
 room = "#tuwunel:matrix.org"
 
 [[servers]]
+name = "Synapse Pro"
+description = "Synapse for Enterprise. Backs the Foundation's matrix.org homeserver "
+author = "Element"
+maturity = "Stable"
+language = "Python and Rust"
+licence = "Element Commercial License"
+repository = "https://element.io/server-suite/synapse-pro"
+room = "#synapse:matrix.org"
+
+
+[[servers]]
 name = "Dendrite"
 description = "Dendrite is a second-generation Matrix homeserver written in Go!"
 author = "Element"
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -20,7 +20,7 @@ room = "#tuwunel:matrix.org"
 
 [[servers]]
 name = "Synapse Pro"
-description = "Synapse for Enterprise. Backs the Foundation's matrix.org homeserver."
+description = "Synapse for Enterprise, with a focus on performance, scalability and compliance."
 author = "Element"
 maturity = "Stable"
 language = "Python and Rust"
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -1,6 +1,6 @@
 [[servers]]
 name = "Synapse"
-description = "Synapse is a Matrix homeserver written in Python/Twisted. "
+description = "Synapse is a Matrix homeserver written in Python/Twisted."
 author = "Element"
 maturity = "Stable"
 language = "Python"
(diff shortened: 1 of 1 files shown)


content/ecosystem/servers/servers.toml
```diff
@@ -18,6 +18,16 @@ licence = "Apache-2.0"
 repository = "https://github.com/matrix-construct/tuwunel"
 room = "#tuwunel:matrix.org"
 
+[[servers]]
+name = "Synapse Pro"
+description = "Synapse for Enterprise. Backs the Foundation's matrix.org homeserver."
```
HarHarLinks: I'm not sure if this is the place and wording to advertise that fact. It would be very welcome though to share further info about the setup for example on https://matrix.org/homeserver/about/.

Also, in light of https://github.com/matrix-org/matrix.org/pull/2907#discussion_r2329757646 I would like @thibaultamartin's opinion here.

MTRNord: https://github.com/matrix-org/matrix.org/pull/2122/files#r1444654419 Is another data point here for prior decisions on similar wording.

MTRNord: This wording feels maybe more applicable to #2913 instead

neilisfragile: I'm not super high temperature on adding the matrix.org use-case, though I thought it would be a useful fact to add since I often have to explain that matrix.org is backed by Synapse Pro rather than Synapse Community. 

I'll rewrite it instead and may follow up with details to the homeserver 'about' page. I like the idea of using the matrix.org homeserver as a success story case study.

HarHarLinks: Thank you, I'm excited about that idea!

content/ecosystem/servers/servers.toml
```diff
@@ -1,6 +1,6 @@
 [[servers]]
 name = "Synapse"
-description = "Synapse is a Matrix homeserver written in Python/Twisted. "
+description = "Synapse is a Matrix homeserver written in Python/Twisted."
 author = "Element"
 maturity = "Stable"
 language = "Python"
```
HarHarLinks, line 6: I just saw this - while we have your attention ;) do you not consider synapse to also be written in rust? Sure, it doesn't have the rust federation worker, but...

neilisfragile: There is a little rust, but it is minimal i.e. 1.1% Rust vs 0.4% HTML. 

We have no short/medium term intention to increase, so given this is a very high-level overview, it seems misleading to say it is written in Rust. The github repo does reference rust, though in that context it probably _is_ reasonable to give the extra detail.

PR #2916, opened 2025-09-19
Remove EMS from hosting page

tulir:
https://element.io/pricing only has "Self-hosted" and "Air-gapped" options, which sounds like they no longer offer hosting services.

Signed-off-by: Tulir Asokan <tulir@maunium.net>


HarHarLinks (changes requested):
Thanks for noticing, lgtm. Please check the PR template https://github.com/matrix-org/matrix.org/blob/main/.github/pull_request_template.md?plain=1, I think we're good to merge once we have the signoff :)

tulir:
Nothing was added, so there's technically nothing to sign off on 🤔

HarHarLinks (approved):
sounds reasonable to me but ianal. idk what exactly is a "contribution" in the sense of the DCO, so to be safe I assume it means "patch". anyway thanks again for the PR and for signing off :)

PR #2973, opened 2025-10-25
fix: language and punctuation in conference banner

pReya:
- fix language (repetition of the word "recordings")
- fix punctuation (two punctuation marks next to each other)


HarHarLinks:
Now we have to choose between #2970 😅

MTRNord:
Chosen #2970 because it came first :)

Issue #3013, opened 2025-11-20
Add an entry in the navbar to browse members


thibaultamartin:
It should be easier to find the members of the Foundation from the navbar.

We can add it under "Foundation".

<img width="407" height="243" alt="Image" src="https://github.com/user-attachments/assets/f0133c1c-ea0f-4fad-a7b7-412785a3ee8b" />

We must ensure the current link `https://matrix.org/membership/` keeps working.

PR #3042, opened 2025-12-06
Add Foundation membership page to Foundation section issue #3013

Ankit-singh-dot:
### Problem
The Membership page (`/membership/`) was not appearing in the Foundation dropdown menu despite:
- Having `extra.navigation = "foundation"` set in its frontmatter
- Being added to `navigation.toml` as a child entry with `parent = "foundation"`

### Root Cause
The Foundation header was using `section = "foundation"`, which automatically generates navigation items from the `/content/foundation/` directory. Since the Membership page is located at `/content/membership/` (outside the foundation directory), it was not included in the auto-generated navigation.

Additionally, the template doesn't support the `parent` field in `[[header.children]]` entries - it expects `children` to be a direct property of the header entry when using manual navigation.

### Solution
Changed the Foundation navigation from using `section` (auto-generated) to manual `children` entries, similar to how the Blog section is structured. This allows explicit control over which pages appear in the Foundation dropdown.

### Changes
- Replaced `section = "foundation"` with `id = "foundation"` and `href = "/foundation/"`
- Added explicit `[[header.children]]` entries for:
  - About (`/foundation/about/`)

  - Governing Board (`/foundation/governing-board/`)
  - Working Groups (`/foundation/working-groups/`)
  - Members (`/membership/`) ← **This fixes the issue**

### Result
The Membership page now appears in the Foundation dropdown menu as "Members", making it easily accessible to users browsing the Foundation section.

- [x] I have signed off my commit
- [x] I tested the site locally using `zola serve`
- [x] The Membership link correctly appears under Foundation
- [x] I did not modify unrelated files or assets

Screenshot 2025-12-06 at 8 53 54 PM

### DIFF
--- a/content/membership/_index.md
+++ b/content/membership/_index.md
@@ -1,7 +1,10 @@
 +++
-title = "Why become a member?"
-aliases = ["/members/"]
+title = "Members"
+aliases = ["/members/", "/membership/"]
 template = "membership.html"
+
+extra.navigation = "foundation"
+
 extra.summary = """
 Becoming a member of the Matrix.org Foundation allows you to support Matrix and
 its mission both financially and by providing guidance regarding where the
--- a/navigation.toml
+++ b/navigation.toml
@@ -4,7 +4,24 @@ href = "https://spec.matrix.org"
 
 [[header]]
 title = "Foundation"
-section = "foundation"
+id = "foundation"
+href = "/foundation/"
+
+[[header.children]]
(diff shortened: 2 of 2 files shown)


awtj8o81ryywg793:
I'm not a maintainer, but I must emphasise that you shouldn’t remove the PR checklist. Your commit needs to be signed off, or the PR as a whole does.

Issue #917, opened 2021-01-04
Settle on "centralization" or "centralisation"


benparsons:
Looks like we will use "centralisation", but not completely agreed yet.

kegsay:
We're British, so centralisation seems like the obvious choice.

thibaultamartin:
en_GB ftw, including on the front page where we say Matrix is "An open network for secure, decentralised communication"

PR #3080, opened 2025-12-15
Set spell checker to en-gb

HarHarLinks:
### Description

Change the spell checker from all English to British English specifically, since we are British and this is general policy.

Let's see how hard CI explodes.

### Related issues

#917

### Role

🎩 Website & Content WG

### Signoff

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -14,7 +14,7 @@ jobs:
       uses: actions/checkout@v6
 
     - name: Check spelling
-      uses: crate-ci/typos@v1.20.3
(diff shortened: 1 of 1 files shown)


HarHarLinks:
> Let's see how hard CI explodes.

0 issues, that can't be right :thinking:

HarHarLinks:
there we go...

So I guess for some content, such as user generated TWIM, this does not make so much sense, and it also seems to be confused about inline HTML in MD.

HarHarLinks:
Website office hour 2026-01-22: decision to not enforce British English this hard in CI

PR #3074, opened 2025-12-14
Fix: Hero heading isn't horizontally centred on small viewports

awtj8o81ryywg793:
### Description

On viewports that are `767px` wide or less, the heading contained within the hero section on the homepage is not horizontally centred. This is due to a redefinition of the `margin` property from `0 auto` to `0 30px`. While this redefinition adds spacing around the heading (on the x-axis), it causes the element to no longer be horizontally centred. 

To address this issue, a base (i.e., outside of any media queries) `padding` property has been added to the heading, set to `0 30px`, which provides spacing around the heading on the x-axis. Additionally, the value of the `max-width` property has been changed to `40rem` to accommodate the new `padding`, and the second `margin` property has been removed.

### Related issues

No related issues, as this is a fairly minor change.

### Role

Individual not affiliated with any project relevant to this PR.

### Timeline

There’s no immediate deadline, so feel free to review this PR whenever you can.

### Signoff

Please [sign off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md) your individual commits or whole pull request.

PR #3092, opened 2025-12-17
Update: Change unit for hero heading max-width

awtj8o81ryywg793:
### Description

Currently, we use `rem` as the unit for the `max-width` of the hero heading. However, that isn't the best unit to use. Instead, we should use `em` so the `max-width` is set relative to the `font-size` of the heading itself, not the root `font-size`.

To maintain the current computed `max-width` of `640px`, let's use `max-width: 16em`, which also equals `640px` when computed.

### Related issues

No related issues, as this is a fairly minor change. However, this PR does address [this](https://github.com/matrix-org/matrix.org/pull/3074#pullrequestreview-3582914712) comment.

### Role

### Timeline

### Signoff

### DIFF
--- a/sass/_index.scss
+++ b/sass/_index.scss
@@ -28,7 +28,7 @@
                 color: #fff;
                 font-size: 2.5rem;
                 font-weight: 700;
-                max-width: 40rem;
+                max-width: 16em;
                 margin: 0 auto;
                 padding: 0 30px;
                 text-align: center;


HarHarLinks:
Maybe I'm lacking the webdesign smarts, but why would I want the width (basically inverse margin?) to depend on the font size?

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3092#issuecomment-3666686912):

> Maybe I'm lacking the webdesign smarts, but why would I want the width (basically inverse margin?) to depend on the font size?

The main reason for this approach is accessibility, as some users might set their default font size to something other than `16px`, like `20px`. In design situations like the one we’re discussing, the link between `font-size` and `max-width` is pretty clear (at least to myself). The `max-width` is set up to let text wrap across two lines while still looking uniform. Using a static value for the heading's `max-width` could lead to more cases where the layout breaks if accessibility settings are changed, resulting in wrapping to more than two lines. On the other hand, using a relative value helps us maintain the ratio between `font-size` and `max-width`, allowing for two-line wrapping.

Using `rem` isn’t necessarily an issue, but I think `em` makes a bit more sense since it shows the connection between the `font-size` of the heading and the `max-width` more clearly. After checking other parts of the website, I noticed we sometimes use different units, like `px`, for `max-width` to achieve text wrapping. Depending on how this discussion goes, I wonder if it’s better to complete this PR as it is and open a new one, or if we should expand the scope of this one to tackle other issues too. Maybe we should turn this discussion into a separate issue? I think I _may_ have misunderstood your [comment](https://github.com/matrix-org/matrix.org/pull/3074#pullrequestreview-3582914712).

thibaultamartin:
I'm still struggling to see the added value of using `em` instead of `rem`, since both take into consideration the font-size the user has set :)

Unless I've missed something, it is functionally the same from an accessibility perspective, and this change only has an impact on developer experience. Is this correct?

If so, I would advocate to keep `rem` instead of `em` since it makes it easier to convert to a value in pixels to get a sense of the dimension we're setting.

If we relied on webcomponents we should probably use `em` more, but we don't use any in this context.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3092#issuecomment-3684090182):

> I'm still struggling to see the added value of using `em` instead of `rem`, since both take into consideration the font-size the user has set :)
>
> Unless I've missed something, it is functionally the same from an accessibility perspective, and this change only has an impact on developer experience. Is this correct?

Yes, that’s correct. The main reason for switching from `rem` to `em` is for the reasons mentioned in [this comment](https://github.com/matrix-org/matrix.org/pull/3092#issuecomment-3667532072), which would fall under "developer experience" I'd say.


Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3092#issuecomment-3684090182):

> If so, I would advocate to keep `rem` instead of `em` since it makes it easier to convert to a value in pixels to get a sense of the dimension we're setting.

Okay. Let's stick with `rem` for now. We can revisit this later if it makes sense to do so.

Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3092#issuecomment-3684090182):

> If we relied on webcomponents we should probably use `em` more, but we don't use any in this context.

Don't tempt me! 😆

PR #3113, opened 2025-12-21
Fix: Add names to external links in footer

### LINKED ISSUE #3091: Accessibility issues in the footer
[WAVE](https://wave.webaim.org/) identified that the footer icons inside of the div with "external-links" only have svg icons but no name. We should add names to ensure these links are not invisible with for example screenreaders.

Image

awtj8o81ryywg793:
### Description

Add names (via the `aria-label` attribute) to external links in footer.

### Related issues

Closes #3091.

### Role

### Timeline

### Signoff

### DIFF
--- a/templates/includes/footer.html
+++ b/templates/includes/footer.html
@@ -6,7 +6,7 @@
     </div>
     <div class="external-links">
         {% for item in navigation.footer_external %}
-        <a href="{{ item.href }}" {% if item.newTab %}target="_blank"{% endif %} {% if item.relMe %}rel="me"{% endif %} {% if item.alt %}aria-label="{{ item.alt }}" title="{{ item.alt }}"{% endif %}><span aria-hidden="true">{{ load_data(path="/assets/" ~ item.image) | safe }}</span></a>
+        <a href="{{ item.href }}" {% if item.newTab %}target="_blank"{% endif %} {% if item.relMe %}rel="me"{% endif %} {% if item.alt %}aria-label="{{ item.alt }}"{% endif %}><span aria-hidden="true">{{ load_data(path="/assets/" ~ item.image) | safe }}</span></a>
         {% endfor %}
     </div>
     <div class="subfooter-links">


awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3113#pullrequestreview-3605457991):

> Firefox a11y devtools day that icons should also have alt text which they don't. I don't know the best practices but my feeling here is since the link labels give the context already, the icons should probably be aria-hidden?
>
> Let me know if you wanna tweak

Let me tweak it. See [`1c0aa85`](https://github.com/matrix-org/matrix.org/pull/3113/commits/1c0aa85b840a2d8e4e10c265e79d6fc453a80bc3).

HarHarLinks (approved):
I think this makes sense. Do you happen to have any references that pointed to this being the best practice solution? It must be a common case, right?

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3113#pullrequestreview-3606148666):

> Do you happen to have any references that pointed to this being the best practice solution?

I'm afraid I do not. I just agreed with your rationale. The SVGs are decorative, and since the link itself contains an `aria-label` attribute, there isn't really a need for anything else. Thinking about it, I think adding a `title` attribute would be a good idea too. As the A11y Collective states:

> The key takeaway here is that aria-label is for assistive technology and title is for visual aids. Each serves its purpose, but they should never be swapped or relied upon interchangeably.

Source: https://www.a11y-collective.com/blog/aria-label-vs-title/

Whilst we’re at it, I wonder if the alt text should be added too?

HarHarLinks:
We should try and find a solid reference for how to handle this best to avoid disproving the situation. E.g. I would imagine that getting icon alt text read out to me as a screen reader users is just annoying when it also reads the important info, which is the link label (which for seeing users is the icon).

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3113#issuecomment-3738311322):

> We should try and find a solid reference for how to handle this best to avoid disproving the situation.

Good shout! I'll see what I can find.

Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3113#issuecomment-3738311322):

> [...] I would imagine that getting icon alt text read out to me as a screen reader users is just annoying when it also reads the important info, which is the link label (which for seeing users is the icon).

Agreed. The best option would probably be to use only `aria-label`, as not all screen readers seem to support `title`.

awtj8o81ryywg793:
> Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3113#issuecomment-3738311322):
> 
> > We should try and find a solid reference for how to handle this best to avoid disproving the situation.
> 
> Good shout! I'll see what I can find.

There’s "aria-label or title? Screen Reader Behaviour Explained" by the A11Y Collective, which states:

> However, the title attribute has some serious drawbacks when it comes to accessibility. While it works well for sighted users, it doesn’t always translate well to screen readers. The content of the title attribute is often ignored or read inconsistently, leaving users who rely on assistive technology at a disadvantage. Also, it’s dependent on mouse hover or keyboard focus. It won’t be [accessible on mobile devices or touchscreens](https://www.a11y-collective.com/blog/mobile-accessibility/), where there’s no concept of hovering.

Source: https://www.a11y-collective.com/blog/aria-label-vs-title/

TheOneWithTheBraid:
LGTM

PR #3114, opened 2025-12-21
Fix: Multiple labels in the navigation bar

### LINKED ISSUE #3090: Accessibility issues in the navigation bar
[WAVE](https://wave.webaim.org/) identified that we have apparently labels of the mobile nav dropdown "leaking" through to desktop.

This causes "Multiple form labels":

Image

and "Empty form label":

Image

as the input field and checkbox are `aria-hidden` while the label apparently is not.

We should follow the guidelines for this properly while keeping the functions working.

awtj8o81ryywg793:
### Description

Add `aria-label` attributes to both `label` elements in the header to ensure neither is seen as empty by assistive technologies. Style the `.page-overlay` `label` so it is not displayed (e.g., `display: none`) on desktop viewports to avoid accessibility issues.

### Related issues

Closes #3090.

### Role

Member of The Matrix.org Foundation Website & Content Working Group.

### Timeline

### Signoff

### DIFF
--- a/templates/includes/header.html
+++ b/templates/includes/header.html
@@ -2,9 +2,9 @@
     <a href="/" class="brand">
         <img src="/images/matrix-logo-white.svg" alt="Matrix logo">
     </a>
-    <input id="site-header-dropdown-checkbox" type="checkbox" class="dropdown-checkbox" aria-hidden="true">
-    <label for="site-header-dropdown-checkbox" class="dropdown-button">&#xe602;</label>
-    <label for="site-header-dropdown-checkbox" class="page-overlay"></label>
+    <input id="site-header-dropdown-checkbox" type="checkbox" class="dropdown-checkbox" aria-hidden="true" aria-labelledby="dropdown-button">
+    <label for="site-header-dropdown-checkbox" class="dropdown-button" id="dropdown-button">&#xe602;</label>
+    <label for="site-header-dropdown-checkbox" class="page-overlay" aria-hidden="true"></label>
     <nav>
         {% for link in navigation.header %}
         {% if link.section %}
--- a/templates/includes/header.html
+++ b/templates/includes/header.html
@@ -3,7 +3,7 @@
         <img src="/images/matrix-logo-white.svg" alt="Matrix logo">
     </a>
     <input id="site-header-dropdown-checkbox" type="checkbox" class="dropdown-checkbox" aria-hidden="true" aria-labelledby="dropdown-button">
-    <label for="site-header-dropdown-checkbox" class="dropdown-button" id="dropdown-button">&#xe602;</label>
+    <label for="site-header-dropdown-checkbox" class="dropdown-button" id="dropdown-button" aria-label="Toggle Site Header Dropdown">&#xe602;</label>
     <label for="site-header-dropdown-checkbox" class="page-overlay" aria-hidden="true"></label>
     <nav>
         {% for link in navigation.header %}
--- a/templates/includes/header.html
+++ b/templates/includes/header.html
@@ -2,8 +2,8 @@
     <a href="/" class="brand">
         <img src="/images/matrix-logo-white.svg" alt="Matrix logo">
     </a>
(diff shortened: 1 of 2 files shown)


templates/includes/header.html
```diff
@@ -2,9 +2,9 @@
     <a href="/" class="brand">
         <img src="/images/matrix-logo-white.svg" alt="Matrix logo">
     </a>
-    <input id="site-header-dropdown-checkbox" type="checkbox" class="dropdown-checkbox" aria-hidden="true">
-    <label for="site-header-dropdown-checkbox" class="dropdown-button">&#xe602;</label>
-    <label for="site-header-dropdown-checkbox" class="page-overlay"></label>
+    <input id="site-header-dropdown-checkbox" type="checkbox" class="dropdown-checkbox" aria-hidden="true" aria-labelledby="dropdown-button">
+    <label for="site-header-dropdown-checkbox" class="dropdown-button" id="dropdown-button" aria-label="Toggle Site Header Dropdown">&#xe602;</label>
```
awtj8o81ryywg793: I'm not sure if there is a preference in terms of capitalisation for use cases like this (e.g., `aria-label`). Any thoughts?

HarHarLinks (commented):
Running the WAVE browser plugin on a local build still raises the same original issue.

I'm not quite sure what's going on here functionally.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3114#pullrequestreview-3605497155):

> Running the WAVE browser plugin on a local build still raises the same original issue.

So, it does. I was actually working backwards by using the web application directed at matrix.org instead of the browser plug-in directed at my local copy of the website. Let me actually install the plug-in... \*facepalm\*

I thought a combination of labelling the `input` with `aria-labelledby` and the second `label` with `aria-hidden="true"` would help, but it appears not. I think the only way we can make that error go away is either removing the second `label`, which is currently used to allow the dropdown to be toggled by clicking outside of it, or restructuring the markup even more and using a JavaScript-only solution. Which isn't ideal.

Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3114#pullrequestreview-3605497155):

> I'm not quite sure what's going on here functionally.

In short: the `.dropdown-button` `label` is acting like a button, and is directly linked, via the `for` attribute,  to the `.dropdown-checkbox` `label` so when it is clicked, it toggles the `checked` state of the `input` (without any JavaScript). Because of this button-like functionality, we should make sure it is seen as a button by assistive technology. Adding the `aria-labelledby` label specifies which label the element is labelled by, which given my earlier point, may not be needed.

awtj8o81ryywg793 (commented):
~I welcome any thoughts on this.~ See [`5b15b8b`](https://github.com/matrix-org/matrix.org/pull/3114/commits/5b15b8b6e5fadf2e43d8723b9df1b9d512e1be84).

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3114#issuecomment-3684676238):

> I think the only way we can make that error go away is either removing the second label, which is currently used to allow the dropdown to be toggled by clicking outside of it, or restructuring the markup even more and using a JavaScript-only solution. Which isn't ideal.

Unless I am mistaken, as long as we retain the two `<label>` elements, we cannot eliminate this error. To resolve the issue, we would need to remove the second toggle and replicate its functionality using JavaScript as a progressive enhancement. However, marking the second `<label>` element with `aria-hidden` will hide it from screen readers and will mitigate any confusion.

templates/includes/header.html
awtj8o81ryywg793: I've refactored the `aria-label` to read slightly differently and to be in sentence case. See [`5b15b8b`](https://github.com/matrix-org/matrix.org/pull/3114/commits/5b15b8b6e5fadf2e43d8723b9df1b9d512e1be84).

awtj8o81ryywg793:
I'll keep this Pull Request as a draft until I get feedback regarding how best to proceed.

awtj8o81ryywg793:
This pull request is no longer needed. See #3130.

PR #3116, opened 2025-12-22
Fix: Remove the unnecessary closing tag for img

awtj8o81ryywg793:
### Description

Re-factor the `img` tag used in the "Platinum member" section.

### Related issues

No related issues, as this is a fairly minor change.

### Role

### Timeline

### Signoff

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -51,7 +51,7 @@
         <div class="row">
             <div class="membership-card">
                 <a target="_blank" rel="noreferrer noopener" href="https://element.io/built-on-matrix" class="membership-card__content">
-                    <img alt="Element.io Logo" src="images/element-creators-of-matrix--dark.svg" />
+                    <img alt="Element.io Logo" src="images/element-creators-of-matrix--dark.svg">
                 </a>
             </div>
         </div>


templates/index.html
```diff
@@ -51,7 +51,7 @@ <h2>Platinum member</h2>
         <div class="row">
             <div class="membership-card">
                 <a target="_blank" rel="noreferrer noopener" href="https://element.io/built-on-matrix" class="membership-card__content">
-                    <img alt="Element.io Logo" src="images/element-creators-of-matrix--dark.svg"></img>
+                    <img alt="Element.io Logo" src="images/element-creators-of-matrix--dark.svg" />
```
HarHarLinks: If https://validator.w3.org/nu/?doc=https%3A%2F%2Fmatrix.org%2F is to be believed, the following would be correct.
```suggestion
                    <img alt="Element.io Logo" src="images/element-creators-of-matrix--dark.svg">
```

However we have in similar cases "Info"s all over the page:
> Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.

Gonna ask the @matrix-org/website-wg what kind of HTML we are targeting.

thibaultamartin: Having a void element without a trailing slash feels very wrong. I'm no expert here, what are the downsides of keeping a trailing slash in real life?

HarHarLinks: As far as I can see, either of the 3 variants work, because practically browsers tend to be very forgiving. In that case I'm not sure why having a full closing tag is worse than the original PR though.

awtj8o81ryywg793: When I refactored the `` element, I quickly checked [this MDN link](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img), which demonstrates `` elements with a trailing slash, and made the change. However, upon further investigation, I can see that there are good reasons to drop the trailing slash (`/`). The main reason is that including the slash can unintentionally lead to it being included in the `src` attribute if it is not surrounded by quotation marks, which results in a broken URL. The use of a closing `</img>` is a non-starter for me since the HTML spec states that `` has no end tag (see [here](https://html.spec.whatwg.org/#the-img-element)).

Perhaps it would be useful to move this discussion to a dedicated issue?

awtj8o81ryywg793: I've signed off on the suggestion from @HarHarLinks. As stated [previously](https://github.com/matrix-org/matrix.org/pull/3116#discussion_r2641126992), there are many similar improvements we can make in terms of the markup. Let's discuss that elsewhere. But in the meantime, let's get this PR merged since a closing `</img>` isn't valid, even if browsers don't seem to have an issue with it.

PR #3117, opened 2025-12-22
fix and minify bluesky.svg

HarHarLinks:
### Description

Used https://validator.w3.org/nu/about.html to validate the site and fixed issues by hand.

SVG minification assisted by https://www.svgminify.com/.

### Role

Website & Content WG

### Signoff


HarHarLinks:
This is a bit of an odd one. We are embedding the SVG inline, but keep it as separate file in the repo. Inlining means the content is HTML, not XML(?), so technically a subpart of the document should not suddenly declare being XML(?), and so mostly we don't. However the separate SVG file is maybe not technically correct without the XML header?

PR #3122, opened 2025-12-25
Fix: Set width and height attributes for Bluesky SVG

### LINKED ISSUE #3121: Bug: Bluesky icon is missing from the footer
Although the icon for Bluesky was added with #3107 and later minified with #3117, it is not visually displayed. However, the SVG is still present in the HTML. It appears that the minification has broken the SVG, as the icon was visible between the merging of both pull requests.

Here’s an image of what the footer currently looks like:

Image

awtj8o81ryywg793:
### Description

Add `width` and `height` attributes to the Bluesky icon SVG. The lack of these attributes is causing a visual bug. See #3121.

### Related issues

Closes #3121.

### Role

### Timeline

There’s no immediate deadline, but considering that the bug prevents users from clicking the Bluesky link, the sooner, the better.

### Signoff


awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3122#pullrequestreview-3612125987):

> I wonder why these aren't styled using CSS instead anyway?

That wouldn't be a bad idea, though it doesn't hurt to have a "default" value defined as an attribute.

HarHarLinks:
I tried in devtools and it seemed you can't override with css if an attribute is set 😕

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3122#issuecomment-3691320896):

> I tried in devtools and it seemed you can't override with css if an attribute is set 😕

I've just tried overwriting the styles via "Developer Tools," and I'm able to do so just fine.

PR #2435, opened 2024-07-22
add CONTRIBUTING.md

joshsimmons:
adding a CONTRIBUTING.md file to the repo to bring it in line with our contribution policies. of note, the language *does not require legally identifiable or "real" names,* which is the approach we're implementing across the board for contribution policies.

Issue #2198, opened 2024-02-20
General Design review


thibaultamartin:
Now that we have actual data in production, we can see that some pages don't really work with the design they have. We need a design review to potentially change fundamentally how some pages display information.

- Global
  - Make sure everything is on brand
  - General accessibility review: link colours, border colours, layouts without borders, etc
  - On very wide screens, the text in the black header looks very off center
  - The navbar is quite crowded, which makes responsiveness difficult to handle (the navbar collapses into a hamburger menu too early)
- Front page
  - We removed the sponsors page for lack of sponsors in the platinum tier, we will hopefully add it back at some point. We need a proper design for it
- About page
  - Doesn't use the black header as the rest of pages
  - Need to define what the page is for: Matrix, the Foundation, both? What do other websites and organisations usually do?
- Blog
  - Looks very "engineer design"
  - The previews are rather large, and difficult to browse
- Blog Post
  - The column is off center 
  - The title sizes are inconsistent (especially in the quotes)
  - TWIM posts look very messy
  - The ToC has inconsistent link colour (it looks like normal text until you hover it)
  - Redesign of the YouTube Player placeholder
- Documentation
  - Docs are not searchable (but [could be](https://www.getzola.org/documentation/getting-started/configuration/))
  - When scrolling through the docs you don't see in the ToC where you're at
  - _Ideally_ having proper styling of the code depending on the language à la starlight
- Ecosystem
  - All the subsections are inconsistent
  - The cards layout doesn't really work with the density of information we want to display
- Try Matrix
  - Do a full review of a user journey to see if it makes sense

MTRNord:
Idea for the sponsors:

What if we instead of having them under "Donate" put them under Ecosystem?

Meaning we would populate https://matrix.org/ecosystem, make the link in the menu clickable (though we need to rethink the mobile menu then) and have a flow like this on it:

1. These are the big players in the community: Lists Sponsors
2. These are all the things the whole community is working on and proud to show: Linking off to clients, servers, etc.

imho it feels natural to consider the sponsors as part of the matrix ecosystem :thinking:

MTRNord:
For discoverability, https://github.com/matrix-org/matrix.org/issues/1115 (dark mode) is another design related wish by people.

awtj8o81ryywg793:
[...]

PR #3130, opened 2025-12-28
Enhancement: Improve the accessibility of the site header

### LINKED ISSUE #3090: Accessibility issues in the navigation bar

### LINKED ISSUE #3115: Enhancement: Improve the accessibility of the site header
Currently, there are a few issues with the site's header:

- `<label>` elements are used as buttons but do not contain the required attributes (e.g., `role`, `aria-pressed`) to mark them as such.
  - A minimal amount of JavaScript will be required to emulate button functionality.
- `<label>` elements cannot be navigated using the keyboard.
  - Adding the attribute `tabindex=0` would fix this and allow all users to navigate to and from it with ease.
- `<a>` elements of first-level navigation items are child elements of toggles (i.e., they are contained within a `<label>` element), rather than sibling elements.
  - Such a structure seems a little misguided. Even if it doesn't cause issues, at least for mouse users.
- `<label>` elements are missing interactive visuals (e.g., `cursor: pointer`) that demonstrate their button-like functionality.

awtj8o81ryywg793:
### Description

Currently, the accessibility of the site header is lacking. For instance, users cannot navigate to the toggle button or change its state via the keyboard. Additionally, while the toggle uses a `<label>` element to allow toggling the drop-down without JavaScript, which is a commendable approach, the element is not marked as a button, despite functioning as one.

This Pull Request introduces a new file, `header.js`, which adds support for interacting with the drop-down toggle via the keyboard, as well as setting `aria-pressed`, `aria-expanded`, `tabindex`, and `role` attributes. Here's a breakdown for each attribute as to why they are added to the element via JavaScript:

- `aria-pressed` - Its value necessarily depends on being updated via JavaScript, which, without it, may lead to confusion for users, as the value of the attribute wouldn't match the toggled state of the drop-down.

- `aria-expanded` - Its value necessarily depends on being updated via JavaScript, which, without it, may lead to confusion for users, as the value of the attribute wouldn't match the toggled state of the drop-down.

- `tabindex` - Setting the attribute's value to `0` allows users to navigate via the keyboard to the drop-down toggle. However, without JavaScript, there is no means by which the user can toggle the visibility of the drop-down. As such, the attribute should only be added via JavaScript to avoid any confusion.

- `role` - Setting the attribute's value to `"button"` signals to assistive technology that the element is a button. However, on its own, such an attribute doesn't add any additional functionality to the `<label>` element that native `<button>` elements have since this functionality must be added manually via JavaScript. As such, the attribute should only be added via JavaScript to avoid any confusion.

For further reading regarding these attributes, I'd suggest checking out: https://www.a11y-collective.com/blog/aria-button/.

In addition to those changes, this pull request also changes the header's drop-down arrows to be `<label>` elements, rather than `<div>` elements, as well as adding `cursor: pointer` to the top-most toggle.

### Related issues

Closes #3115. Closes #3090.

### Role

Member of The Matrix.org Foundation Website & Content Working Group.

### Timeline

### Signoff

### DIFF
--- a/static/js/header.js
+++ b/static/js/header.js
@@ -54,7 +54,7 @@ const handleToggle = (element) => {
     const relatedToggles = getRelatedToggles(element);
     relatedToggles.forEach(relatedElement => {
         toggleAriaAttributes(relatedElement);
-        relatedElement.setAttribute("aria-pressed", false);
+        relatedElement.setAttribute("aria-pressed", "false");
     });
 
     const checkbox = getCheckbox(element);
--- a/static/js/header.js
+++ b/static/js/header.js
@@ -25,7 +25,7 @@ const getRelatedToggles = (element) =>
 /**
  * Toggles the ARIA attributes of a given label element.
  *
- * @param {HTMLElement} element - The label element to toggle attributes.
+ * @param {HTMLElement} element - The label element to toggle attributes for.
  */
 const toggleAriaAttributes = (element) => {
     const isExpanded = element.getAttribute("aria-expanded") === "true";
@@ -39,7 +39,7 @@ const toggleAriaAttributes = (element) => {
  * Handles the toggle action for a given label element, including updating
  * ARIA attributes and checkbox state.
  *
- * @param {HTMLElement} element - The label element to handle toggle for.
+ * @param {HTMLElement} element - The label element to handle the toggle for.
  */
 const handleToggle = (element) => {
     toggleAriaAttributes(element);
@@ -63,9 +63,8 @@ toggles.forEach(element => {
     element.setAttribute("aria-pressed", "false");
 
     const checkbox = getCheckbox(element);
-    if (checkbox) {
+    if (checkbox)
         checkbox.checked = false; // Ensure a consistent state
-    }
 
     element.addEventListener("click", (event) => {
         event.preventDefault(); // Prevent the default browser behaviour (i.e., toggling).
--- a/static/js/header.js
+++ b/static/js/header.js
@@ -12,7 +12,7 @@ const getCheckbox = (element) =>
     );
 
 /**
- * Gets all related toggles for a given label element, excluding itself.
+ * Retrieves all related toggles for a given label element, excluding itself.
  *
  * @param {HTMLElement} element - The label element.
  * @returns {HTMLElement[]} An array of related label elements.
--- a/static/js/header.js
+++ b/static/js/header.js
@@ -22,8 +22,6 @@ function toggleCheckbox(element) {
     element.checked = !element.checked;
 }
 
-console.log(dropdownToggle);
-
 document.addEventListener("DOMContentLoaded", () => {
     /**
      * Set the aria-pressed attribute on the dropdown toggle as a progressive enhancement
--- a/static/js/header.js
+++ b/static/js/header.js
@@ -1,11 +1,13 @@
-// SPDX-FileCopyrightText: 2026 The Matrix.org Foundation C.I.C.
-// SPDX-FileContributor: awtj8o81ryywg793
-//
-// SPDX-License-Identifier: Apache-2.0
+/*
+ * SPDX-FileCopyrightText: 2026 The Matrix.org Foundation C.I.C.
+ * SPDX-FileContributor: awtj8o81ryywg793
+ * 
+ * SPDX-License-Identifier: Apache-2.0
+ */
 
 const toggles = document.querySelectorAll(".site-header label");
 
-/**
+/*
  * Retrieves the checkbox associated with a given label element.
  *
  * @param {HTMLElement} element - The label element.
@@ -16,7 +18,7 @@ const getCheckbox = (element) =>
         `.site-header input[type="checkbox"]#${element.getAttribute('for')}`
     );
 
-/**
+/*
  * Retrieves all related toggles for a given label element, excluding itself.
  *
  * @param {HTMLElement} element - The label element.
@@ -27,7 +29,7 @@ const getRelatedToggles = (element) =>
         `.site-header label[for="${element.getAttribute('for')}"]`
     )).filter(related => related !== element);
 
-/**
+/*
  * Toggles the ARIA attributes of a given label element.
  *
  * @param {HTMLElement} element - The label element to toggle attributes for.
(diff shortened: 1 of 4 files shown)


awtj8o81ryywg793:
Whilst I initially planned to just tackle accessibility issues in this PR, I moved to also refactor the header as a whole. However, I've now scaled back the scope of this PR to its original goal. A refactor will be made if (or when) we move to a more component/macro-based workflow, as suggested in this [comment](https://github.com/matrix-org/matrix.org/issues/2198#issuecomment-3704285534).

awtj8o81ryywg793:
Given the introduction of `header.js` and the outcome of #3135, I'd like to include the required REUSE information. Before doing so, could it be clarified _who_ the copyright holder is? Would it be permissible to use a GitHub username for either copyright holders or contributors?

HarHarLinks:
> Given the introduction of `header.js` and the outcome of #3135, I'd like to include the required REUSE information. Before doing so, could it be clarified _who_ the copyright holder is? Would it be permissible to use a GitHub username for either copyright holders or contributors?

IANAL, but aiui yes:
https://reuse.software/faq/#copy-pseudonym
https://github.com/matrix-org/matrix.org/pull/2435#issue-2423900118

HarHarLinks:
> For instance, users cannot navigate to the toggle button via the keyboard or change its state.

agreed. in terms of screen reader, would it be feasible/preferable to not have to toggle anything in the first place?

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3130#issuecomment-3797093326):

> agreed. in terms of screen reader, would it be feasible/preferable to not have to toggle anything in the first place?

I'm not sure it would be, no. How would you see that working?

static/js/header.js
```diff
@@ -0,0 +1,87 @@
...
+        `.site-header label[for="${element.getAttribute('for')}"]`
+    )).filter(related => related !== element);
+
+/*
+ * Toggles the ARIA attributes of a given label element.
+ *
+ * @param {HTMLElement} element - The label element to toggle attributes for.
+ */
+const toggleAriaAttributes = (element) => {
+    const isExpanded = element.getAttribute("aria-expanded") === "true";
```
MTRNord, line 38: Hm I am a bit unsure on this. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded reads like the aria-expanded should be on the thing that is being expanded and not on the button. However, this looks like we set it on the button instead.

In fact the below citation to me makes it sound like we don't need this at all unless the items we toggle are children of this element. 

> For example, aria-expanded is applied to the parent item in a DOM tree to indicate whether its child branch is shown. The parent controls the visibility of the associated child branch, as well.

Though if we want to use it the page goes on with:

> There are two declarations that can be applied to objects that control the visibility of another object: [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls), or [aria-owns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-owns) combined with aria-expanded. The aria-controls and aria-owns indicate the relationship between the controlling element and the controlled element. The aria-expanded indicates to assistive technology whether the controlled element is expanded or collapsed.

So we likely also would need/want some variant of aria-owns/aria-controls on the button then.

Though I am not entirely sure if this is correct or not. Feel free to tell me I am saying bullshit here :D

static/js/header.js
```diff
@@ -0,0 +1,87 @@
...
+
+    const checkbox = getCheckbox(element);
+    if (checkbox)
+        checkbox.checked = !checkbox.checked;
+};
+
+// Initialise toggles
+toggles.forEach(element => {
+    element.setAttribute("role", "button");
+    element.setAttribute("tabindex", 0);
```
MTRNord, line 68: I wonder if this is a case of the warning at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex#accessibility_concerns

In general I am starting to wonder  why we use a label here and not a button? Is there anything preventing us to use a button and then styling it? Wouldn't that give us better accessibility?

awtj8o81ryywg793: Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3130#discussion_r3094063235):

> I wonder if this is a case of the warning at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex#accessibility_concerns

Thanks for raising the concern. This isn't an issue here because we use ARIA to mark the elements as buttons and handle button-like functionality with JavaScript. Without `tabindex` set to `0` users wouldn't be able to keyboard-focus those elements. `<label>` elements are also listed as interactive content. See: https://developer.mozilla.org/docs/Web/HTML/Guides/Content_categories#interactive_content

Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3130#discussion_r3094063235):

> In general I am starting to wonder why we use a label here and not a button? Is there anything preventing us to use a button and then styling it? Wouldn't that give us better accessibility?

We use `<label>` elements instead of `<button>` elements because they (when linked to a checkbox) provide toggle functionality without JavaScript.

However, using multiple `<label>` elements with a single `<input>` is not valid. Let's switch the overlay to a `<button>` instead and treat it as a progressive enhancement.

static/js/header.js
awtj8o81ryywg793: Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3130/changes#r3094036949):

> Though I am not entirely sure if this is correct or not. Feel free to tell me I am saying bullshit here :D

As your link shows, setting `aria-expanded` on the button is fine. The attribute may be used whether or not the element it is applied to is the parent of the toggled element. What is important, and what I missed, is using either `aria-controls` or `aria-owns`.

PR #3171, opened 2026-01-22
Ensure distributions are sorted by maturity and then name within its maturity group

MTRNord:
### Description

Changes the distributions page to a more deterministic sorting
image

### Role

Website WG Hat

### Timeline

Whenever

### Signoff

See Commits :)

### DIFF
--- a/templates/shortcodes/all_distributions.html
+++ b/templates/shortcodes/all_distributions.html
@@ -96,7 +96,17 @@
 <div id="filters-overlay"></div>
 
 <div id="all-distributions" class="projects-card-deck">
-    {% for distribution in distributions_data.distributions | sort(attribute="maturity") | reverse %}
+    {# Group distributions by maturity server-side and sort by name within each group #} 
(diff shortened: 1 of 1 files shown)


HarHarLinks:
Imo, (prior to [caa1bee](https://github.com/matrix-org/matrix.org/commit/caa1bee03ea7c7cb1ce0e27c1d3d0d2da42b04c9) adding ESS Pro in the front), these were sorted in a deliberate way by hand, of course with the subjective bias that brings.

I feel like currently, this page is making recommendations towards community distros, and then afterwards also mentioning somewhat special audience distros whose audience however i estimate is much more likely to discover them through different means. This isn't to say I'm apposed to presenting this content in a way that is easier to understand than these implications. See our handful of issues about the topic.

If we're merging this, I would shuffle the toml so synapse and synapse pro are next to each other.

MTRNord:
From the meeting: Maybe we need to split it between community and "professional use"/enterprise.

Issue #2405, opened 2024-06-19
show project maturity in the ecosystem overviews


HarHarLinks:
e.g. browsing https://matrix.org/ecosystem/clients/ you'd assume that EX is stable

HarHarLinks:
that would be a quick workaround, but to be clear I'm not against keeping it in the recommendations in general, but it can't stay like how it is now.

PR #3172, opened 2026-01-22
medium style overhaul for the ecosystem to have better consistency across the pages and fix some issues

MTRNord:
### Description

| Before | After |
|----------- |-------- |
|image|image|
|image|image|
|image|image|
|image|image|

This fixes:

- Better title wrapping
- The pill doesn't cause overflow issues with the title anymore
- The license is much more distinct
- The title is a bit smaller now
- Maturity of Clients is shown as well
- Cards are now the same across these pages (though noticeably Integrations can't utilize all of it)

### Related issues

- https://github.com/matrix-org/matrix.org/issues/2405
- I think also some of our general design overhaul tasks

### Role

Website WG Hat :)

### Timeline

Soon? Never? Now? When the moon phase turned 23 times and its the second blood moon after the third wave of a silent night?

### Signoff

See commits :)

### DIFF
--- a/content/ecosystem/bridges/X/bridges.toml
+++ b/content/ecosystem/bridges/X/bridges.toml
@@ -16,7 +16,7 @@ featured = true
 privilege.platform = "User" # Free text
 privilege.matrix = "Homeserver Admin" # Any of Homeserver Admin, Room Admin, None
 [bridges.supports]
-dm = true
+direct_messages = true
 message_media = true
 redactions = true
 reactions = true
--- a/sass/_bridges.scss
+++ b/sass/_bridges.scss
@@ -22,6 +22,7 @@
 
         border: 1px solid #d2d2d2;
         border-radius: 12px;
+        transition: border-color 0.3s ease, background-color 0.3s ease;
 
         img {
             width: 64px;
@@ -51,6 +52,7 @@
 
         &:hover {
             background-color: #e9e9e9;
+            border-color: #000;
         }
     }
 }
--- a/sass/_bridges.scss
+++ b/sass/_bridges.scss
@@ -5,7 +5,7 @@
 .platforms-grid {
     display: grid;
     grid-template-columns: repeat(auto-fill, 17.6rem);
-    row-gap: .8rem;
+    gap: .8rem;
     margin-block: 2rem 3rem;
     margin-inline: auto;
 
@@ -16,10 +16,12 @@
         align-items: center;
         padding-block: .5rem;
         padding-inline: .5rem 1rem;
-        max-width: fit-content;
         cursor: pointer;
 
         color: var(--color-text);
+
+        border: 1px solid #d2d2d2;
+        border-radius: 12px;
 
         img {
             width: 64px;
@@ -49,7 +51,6 @@
 
         &:hover {
             background-color: #e9e9e9;
-            border-radius: 12px;
         }
     }
 }
@@ -81,12 +82,16 @@
     }
 
     .title-row {
+        display: flex;
+        flex-direction: row;
         align-items: center;
         justify-content: flex-start;
         gap: .5rem;
+        width: auto;
 
         h3 {
-            margin-block: 1rem;
+            margin-block: .5rem;
+            text-align: left;
         }
     }
 
--- a/sass/_bridges.scss
+++ b/sass/_bridges.scss
@@ -11,6 +11,8 @@
     justify-content: center;
 
     .platform {
+        @extend .clickable-card;
+
         display: flex;
         flex-direction: row;
         gap: 1rem;
--- a/sass/_bridges.scss
+++ b/sass/_bridges.scss
@@ -8,6 +8,7 @@
     gap: .8rem;
     margin-block: 2rem 3rem;
     margin-inline: auto;
+    justify-content: center;
 
     .platform {
         display: flex;
--- a/sass/_hosting.scss
+++ b/sass/_hosting.scss
@@ -13,11 +13,12 @@
     }
 
     .hosting_providers {
-        display: flex;
-        flex-direction: row;
-        justify-content: center;
-        flex-wrap: wrap;
+        display: grid;
+        grid-template-columns: repeat(auto-fit, minmax(300px, 1fr));
         gap: 3.125rem;
+        justify-content: center;
+        align-items: stretch;
+        width: 100%;
 
         .provider_card {
             display: flex;
@@ -25,16 +26,15 @@
             justify-content: space-between;
 
             padding: 24px;
-
-            flex: 1 0 400px;
+            width: 100%;
             max-width: 400px;
+            margin: 0 auto;
+
             @media (max-width: 767px) {
-                flex: 1 0 300px;
                 max-width: 300px;
             }
 
-
-            border: 1px solid #D2D2D2;
+            border: 1px solid #d2d2d2;
             border-radius: 16px;
 
             .details {
@@ -57,16 +57,15 @@
 
                 .operating_since {
                     height: 1.5rem;
-                    font-size: .9rem;
+                    font-size: 0.9rem;
                 }
-
             }
 
             .call-to-action {
                 background-color: #000;
                 color: #fff;
                 border-radius: 1000px;
-                padding: .4rem 2rem;
+                padding: 0.4rem 2rem;
                 align-self: center;
             }
         }
--- a/sass/_hosting.scss
+++ b/sass/_hosting.scss
@@ -1,6 +1,6 @@
 .hosting_page {
     margin: 0 auto;
-    padding: 5.625rem 0 2rem 0;
+    padding-block: 5.625rem 2rem;
     @media (max-width: 767px) {
         padding: 5.625rem 32px 2rem 32px;
     }
@@ -25,14 +25,9 @@
             flex-direction: column;
             justify-content: space-between;
 
-            padding: 24px;
+            padding: 32px;
             width: 100%;
-            max-width: 400px;
             margin: 0 auto;
-
-            @media (max-width: 767px) {
-                max-width: 300px;
-            }
 
             border: 1px solid #d2d2d2;
             border-radius: 16px;
--- a/sass/_hosting.scss
+++ b/sass/_hosting.scss
@@ -21,6 +21,8 @@
         width: 100%;
 
         .provider_card {
+            @extend .hoverable-card;
+
             display: flex;
             flex-direction: column;
             justify-content: space-between;
--- a/sass/_hosting.scss
+++ b/sass/_hosting.scss
@@ -21,8 +21,6 @@
         width: 100%;
 
         .provider_card {
-            @extend .hoverable-card;
-
             display: flex;
             flex-direction: column;
             justify-content: space-between;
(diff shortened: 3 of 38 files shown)


HarHarLinks:
Preview: https://mtrnord-overhaul-ecosystem.matrix-website.pages.dev/ecosystem/

HarHarLinks:
Looks like the filters stopped working on my phone on this PR deployment.

MTRNord:
> Looks like the filters stopped working on my phone on this PR deployment.

Seems like its caused by cloudflare and cors:

>    Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://bb593a6f.matrix-website.pages.dev/js/servers.js?h=1cb6dd610dd7ad0025a8. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200. 

Using https://bb593a6f.matrix-website.pages.dev/ecosystem/servers/ instead works fine

HarHarLinks:
Screenshots look good, so does testing on mobile. 

I think cards could be a bit wider on mobile? 
Screenshot_20260122-211840

templates/ecosystem/clients.html
```diff
@@ -40,9 +40,15 @@ <h2>Featured clients</h2>
...
-                <span>{{ client.content | safe }}</span>
-                <div class="platform-links">
+                <div class="title-row">
+                    <h3>{{ client.title }}</h3>
+                    <div class="maturity {{ client.extra.maturity | lower }}">{{ client.extra.maturity }}</div>
+                </div>
+                <div class="metadata-pills">
+                    <div class="pill licence">{{ client.extra.licence }}</div>
+                </div>
+                <p class="description">{{ client.content | safe }}</p>
```
awtj8o81ryywg793:
```suggestion
                <p class="description">{{ client.content  | markdown(inline=true) | safe }}</p>
```

Without using `markdown(inline=true)` the output isn't correct.

Without `markdown(inline=true)` :

```html
<p class="description"><p>...</p></p>
```

With `markdown(inline=true)` :

```html
<p class="description">...</p>
```

Source: https://www.getzola.org/documentation/templates/overview/#markdown

templates/ecosystem/clients.html
```diff
@@ -202,9 +208,15 @@ <h2>All known clients</h2>
...
-                <span>{{ client.content | regex_replace(pattern=`<a.*>(?P<text>.*)</a>`, rep=`$text`) | safe  }}</span>
-                <div class="platform-links">
+                <div class="title-row">
+                    <h3>{{ client.title }}</h3>
+                    <div class="maturity {{ client.extra.maturity | lower }}">{{ client.extra.maturity }}</div>
+                </div>
+                <div class="metadata-pills">
+                    <div class="pill licence">{{ client.extra.licence }}</div>
+                </div>
+                <p class="description">{{ client.content | regex_replace(pattern=`<a.*>(?P<text>.*)</a>`, rep=`$text`) | safe }}</p>
```
awtj8o81ryywg793:
```suggestion
                <p class="description">{{ client.content | markdown(inline=true) | regex_replace(pattern=`<a.*>(?P<text>.*)</a>`, rep=`$text`) | safe }}</p>
```

Without using `markdown(inline=true)` the output isn't correct.

Without `markdown(inline=true)` :

```html
<p class="description"><p>...</p></p>
```

With `markdown(inline=true)` :

```html
<p class="description">...</p>
```

Source: https://www.getzola.org/documentation/templates/overview/#markdown

awtj8o81ryywg793 (changes requested):
Just a few small suggestions.

templates/ecosystem/clients.html
MTRNord: Kinda applied this. However needed to also explicitly strip links from the descriptions as nested a tags are disallowed in html and breaks the parsing in browsers like firefox

awtj8o81ryywg793 (commented):
Everything so far looks good to me. Thank you for implementing my suggested changes.

There are other areas in files not currently part of this request that exhibit the same issue. Would you mind including fixes for those in this PR?

See:

- [line 21](https://github.com/matrix-org/matrix.org/blob/b6b1445791964ffc1add7feb410e4a8885f7f37e/templates/ecosystem/bridge_implementations.html#L21) of `templates/ecosystem/bridge_implementations.html`
- [line 12](https://github.com/matrix-org/matrix.org/blob/97cbab6a83e3d01cb69ff5d748edebef3ca2485c/templates/shortcodes/hosting_providers.html#L12) of `templates/shortcodes/hosting_providers.html`

In addition to these files, do we know why we use a `<span>` element for the description of each [integration](https://e062d81a.matrix-website.pages.dev/ecosystem/integrations/)? Please refer to [line 6](https://github.com/matrix-org/matrix.org/blob/97cbab6a83e3d01cb69ff5d748edebef3ca2485c/templates/shortcodes/all_integrations.html#L6) of `templates/shortcodes/all_integrations.html`. I would suggest switching the `<span>` to a `<p>` element, with the `.description` class assigned so its styling matches the other ecosystem pages.

awtj8o81ryywg793:
I assume the answer to this question will be "yes," but I want to confirm. Is it intended for the hosting provider card to look similar yet different from the other ecosystem cards? I'm mainly referring to the larger font used for the hosting provider card's description compared to that of the other ecosystem cards.

I also noticed a responsive styling bug on the [hosting providers page](https://matrix.org/ecosystem/hosting/). If you set your browser's viewport to `850px` and gradually shrink the width, you'll see the layout cycling between a two-in-a-row and one-in-a-row layout. I think we should update the responsive styling to align with the styling of other ecosystem pages too.

sass/_hosting.scss
awtj8o81ryywg793:
```suggestion
    padding-block: 5.625rem 2rem;
```

Currently, we have an issue in viewports between `767px` and `910px` in width, where there is no padding on the left and right sides of `.content.hosting_page`, which means things look rather cramped. Given that, on `.hosting_page` let's swap out `padding: 5.625rem 0 2rem 0` for `padding-block: 5.625rem 2rem` since `.content` already defines `padding-inline: 1rem`.

sass/_hosting.scss
```diff
@@ -13,28 +13,28 @@
...
+        justify-content: center;
+        align-items: stretch;
+        width: 100%;
 
         .provider_card {
             display: flex;
             flex-direction: column;
             justify-content: space-between;
 
             padding: 24px;
```
awtj8o81ryywg793:
```suggestion
            padding: 32px;
```

sass/_hosting.scss
```diff
@@ -13,28 +13,28 @@
...
-
-            flex: 1 0 400px;
+            width: 100%;
             max-width: 400px;
```
awtj8o81ryywg793:
```suggestion
           
```

Given the use of `grid-template-columns` by `.hosting_providers`, we don't need to define a `max-width`.

This change also helps to ensure that the gap between all `.provider_card` elements within the grid is actually `3.125rem`, as defined for `.hosting_providers`.

sass/_hosting.scss
```diff
@@ -13,28 +13,28 @@
...
+            margin: 0 auto;
+
             @media (max-width: 767px) {
-                flex: 1 0 300px;
                 max-width: 300px;
```
awtj8o81ryywg793:
```suggestion
                
```

Given the use of `grid-template-columns` by `.hosting_providers`, we don't need to define a `max-width`.

This change also helps to ensure that the gap between all `.provider_card` elements within the grid is actually `3.125rem`, as defined for `.hosting_providers`.

awtj8o81ryywg793 (dismissed):

HarHarLinks:
this looks a bit unfortunate on mobile, but I would also be happy to tackle it in a separate PR if you prefer
image

similar I'm not sure why we use different h- and v-gaps for download buttons:
image

HarHarLinks:
As discussed in https://github.com/matrix-org/matrix.org/pull/3219#discussion_r2807342149 I do like the idea of `transition: border-color 0.3s ease;` to black for clickable cards in general.

HarHarLinks:
image
The bridge platform buttons are invisible on mobile while they have a hover background on desktop. The buttons are about the size of the purple debug marker depending on the size of each individual one. I think it would be better if the clickable areas were more clearly visible.

I think the grid would also benefit from `justify-content: center;`?

HarHarLinks:
There are some spacing and alignment issues on the bridge details:
image

I also think we should rather spell out DM as Direct Messages.

HarHarLinks:
Similar situation of invisible buttons on mobile: 
image

HarHarLinks:
Hosters in 3 columns doesn't work for me as long as the amount of text remains so large.

There is a lot of wasted space with 2 or 1 columns:
image

Maybe it would be better to have an individual page for each hoster if we grant them that much text.

HarHarLinks (changes requested):
No need to implement all my comments in this PR, but at least this one should not introduce regressions.

MTRNord:
@HarHarLinks please take another look. I did some fixes but not all:

- I did try to fix the hosting page a bit but choose 1 column on mobile for now. I believe making it have detail cards is another PR. We already are feature creeping here a bit I think. I think its still a small improvement over what we had at the time you did the review.
- I fixed the detail pages for clients and bridges.
- I made DM to "Direct Messages"
- I did try to make the click actions on the bridges more obvious. Not sure how well though
- I added the transition animation

I also applied all suggestions by @awtj8o81ryywg793 

I didnt fix the invisible buttons on mobile for the SDKs. No idea how to make that better right now -> defer to an issue please?

Either way lets have another round of reviews @HarHarLinks and @awtj8o81ryywg793 ?

sass/_bridges.scss
```diff
@@ -5,7 +5,7 @@
 .platforms-grid {
     display: grid;
     grid-template-columns: repeat(auto-fill, 17.6rem);
-    row-gap: .8rem;
+    gap: .8rem;
     margin-block: 2rem 3rem;
     margin-inline: auto;
```
HarHarLinks, line 10:
```suggestion
    margin-inline: auto;
    justify-content: center;
```

content/ecosystem/bridges/X/bridges.toml
```diff
@@ -16,7 +16,7 @@ featured = true
 privilege.platform = "User" # Free text
 privilege.matrix = "Homeserver Admin" # Any of Homeserver Admin, Room Admin, None
 [bridges.supports]
-dm = true
+direct_messages = true
```
HarHarLinks, line 19: this needs to make it into CONTENT.md

sass/_projects.scss
HarHarLinks, line 1: server cards are not clickable, yet they highlight (explicitly per servers.scss).
a lot of project cards are not clickable yet highlight.

clients and bridges are the only ones that are actually clickable.

meanwhile hosting cards are not clickable and do not highlight.

if we decide to give clickable cards or cards in general this visual feature, I think this PR should include for /branding cards to also highlight.

perhaps the highlighting feature should a ~globally available css class?

MTRNord: I just gave everything a hover highlight now.

HarHarLinks (approved):
i still find confusing for cards to animate on hover if then they are not clickable.

the css would do great to attempt to follow some kind of component approach instead of redefining most things for each category.

overall i'd rather have this major improvement today than wait another half year though.

Issue #2837, opened 2025-07-21
Update footer socials


HarHarLinks:
we have at least a linkedin and a bluesky. idk if our twitter is still active.

@thibaultamartin help?

pReya:
@thibaultamartin If you let me know which profiles should be included, I'd be happy to update the code/icons.

thibaultamartin:
Our Twitter is semi-active, in the sense that our [Buffer](https://buffer.com/) is still pushing posts to it but I don't believe anyone is actively monitoring it.

Our currently mostly active socials are:

- Fediverse - https://mastodon.matrix.org/@matrix
- LinkedIn - https://www.linkedin.com/company/matrix-org/
- Bluesky - https://bsky.app/profile/matrix.org
- YouTube - https://www.youtube.com/@Matrixdotorg

HarHarLinks:
Please follow the pattern/precedence from 09ecbea6605dc78a360784033ce259fb1871a611 when implementing.

HarHarLinks:
I think we're currently only missing LinkedIn out of the above.

PR #3184, opened 2026-01-30
Add linkedin to our socials in the footer

dannycolin:
### Description

Add linkedin to our socials in the footer

### Related issues

https://github.com/matrix-org/matrix.org/issues/2837

### Role

Matrix Hackathon 2026

### Timeline

No timeline

### Signoff


HarHarLinks:
Could you annotate the picture following the pattern of https://github.com/matrix-org/matrix.org/blob/main/static/assets/bluesky.svg and https://github.com/matrix-org/matrix.org/blob/main/static/assets/bluesky.svg.license? We haven't written https://reuse.software/ into the contributing policy yet, however the repo is claiming Apache or CC on all content which isn't accurate if we include others' assets without their express permission, so we started to track this using the principle of https://reuse.software/. It can be a bit annoying to figure out licenses sometimes, so let me know if you run into trouble.

Companies often provide a press kit or similar with their official logo and certain permissions that ca be reference, which might be helpful here.

dannycolin:
No worries. I don't think it was mentioned in the documentation so I want with the default copyright notice in the svg file at first. It should follow the pattern you suggested now.

dannycolin:
@HarHarLinks As mentioned on the Thursday meeting, I updated this patch with the license inforamtions. Unless we're really against using fontawesome, this should be good to merge now.

PR #3194, opened 2026-02-03
Update info for Commet client

Airyzz:
### Description

Updating status for the Commet client, due to recent update

### Role

I'm the maintainer of Commet

### Timeline

At your nearest convenience

### Signoff

Signed-off-by: Airyz <airyz@proton.me>

### DIFF
--- a/content/ecosystem/clients/commet.svg.license
+++ b/content/ecosystem/clients/commet.svg.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2026 Airyz
+
+SPDX-License-Identifier: CC-BY-SA-3.0


content/ecosystem/clients/commet.svg
HarHarLinks, line 1: Thank you! We have an ongoing effort of [REUSE](https://reuse.software/)-ising any images on the website to ensure we publish them under the correct license. I take this contribution as implicit agreement to use this image at all - would you mind clarifying under what specific license you grant The Foundation to use it? By default [we suggest](https://github.com/matrix-org/matrix.org/blob/main/LICENSE.md) CC BY-SA for "content". Could you add the appropriate metadata in the same way as it was done e.g. here: https://github.com/matrix-org/matrix.org/blob/main/content/ecosystem/clients/extera-next.svg.license

jermanuts:
@Airyzz why you didn't mention that it is available for mobile too?

PR #3239, opened 2026-02-26
Improve the Try Matrix Page

MTRNord:
### Description

Well its a whole redesign. I suggest having a look at the deployment for all the things. Trying is better than words in this case

### Role

Website and Content WG

### Timeline

Whenever

### Signoff

See commits


static/js/providers-carousel.js
HarHarLinks, line 1: there is a bug where reaching the end makes it jump to the start instead of rotating endlessly.

this seems like it would be possible in pure css: https://codepen.io/cbolson/pen/ZYEdGQX

It would be awesome to be able to scroll the carousel by mouse on top of the animation.

sass/_try-matrix.scss
HarHarLinks, line 1: i think both client and hosting cards should use the same outline ui pattern from https://github.com/matrix-org/matrix.org/pull/3172#discussion_r2856821094

HarHarLinks (commented):
looks very promising to me!

sass/_try-matrix.scss
MTRNord: Right. I can do that. But also keeping this open since I guess I need to patch/update that to have no duplicated code once #3172 is merged eventually :D

sass/_hosting.scss
```diff
@@ -1,74 +1,83 @@
...
+}
+
+.provider_card {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+
+    padding: 24px;
+
+    flex: 1 0 400px;
```
weeman1337: If it has basis 400px and is only allowed to grow but not shrink… then you could just set the width. Also for the other places.

sass/_hosting.scss
```diff
@@ -1,74 +1,83 @@
-.hosting_page {
-    margin: 0 auto;
-    padding: 5.625rem 0 2rem 0;
+.hosting_providers {
+    display: flex;
+    flex-direction: row;
```
weeman1337: `row` is the default…

```suggestion
```

templates/try-matrix.html
```diff
@@ -9,58 +12,147 @@ <h1>{{ page.title }}</h1>
...
+                            <p class="open-project">View app details</p>
+                        </a>
+                    </div>
+                    {% endif %}
+                    {% endfor %}
+                </div>
                 <div class="cta-wrapper">
-                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
-                    <a href="/ecosystem/clients" class="call-to-action secondary">Browse clients</a>
+                    <a href="/ecosystem/clients" class="call-to-action secondary">Browse all apps</a>
```
weeman1337: Link leads to a partially broken page with no icons because of  missing trailing `/`


```suggestion
                    <a href="/ecosystem/clients/" class="call-to-action secondary">Browse all apps</a>
```

weeman1337 (commented):
All the buttons are CTA buttons except those two:

Image

Maybe they can also be CTAs „Browse server software“ / „Browse Distributions“.

---

Other than that the content is now longer and offers more options. In my opinion that makes the onboarding looking more complicated. The ultimate try matrix solution would be one button and you are in. Of course that doesn't work. Maybe this is a topic for discussion and should not block the PR.

sass/_try-matrix.scss
weeman1337: And they should be aligned :D

Image

MTRNord: Gone now anyway :)

sass/_hosting.scss
MTRNord: I reverted the changes to this file entirely now since its not needed anymore

sass/_hosting.scss
MTRNord: I reverted the changes to this file entirely now since its not needed anymore

MTRNord:
@weeman1337 

> All the buttons are CTA buttons except those two:
> Image
> 
> Maybe they can also be CTAs „Browse server software“ / „Browse Distributions“.
> 
> Other than that the content is now longer and offers more options. In my opinion that makes the onboarding looking more complicated. The ultimate try matrix solution would be one button and you are in. Of course that doesn't work. Maybe this is a topic for discussion and should not block the PR.

I reverted the hosting.scss changes now entirely as they got redundant.
As discussed in the meeting there is now a TLDR at the top.
I did make the buttons a CTA now as well.

I hope that addresses some of your concerns.

static/js/providers-carousel.js
```diff
@@ -0,0 +1,82 @@
+(function () {
```
awtj8o81ryywg793, line 7: I think we should include REUSE license information, especially since it is a new file.

See: [github.com/awtj8o81ryywg793/matrix.org/blob/fe43450cf31c426564bf925e06711565cb2c6472/static/js/header.js](https://github.com/awtj8o81ryywg793/matrix.org/blob/fe43450cf31c426564bf925e06711565cb2c6472/static/js/header.js#L1)

static/js/providers-carousel.js
```diff
@@ -0,0 +1,82 @@
...
+    const carousel = document.querySelector('.providers-carousel');
+    if (!carousel) return;
+
+    const items = [...carousel.querySelectorAll('.provider-compact')];
+    const NUM_ITEMS = items.length;
+    const DURATION_S = 30; // keep in sync with CSS --carousel-duration
```
awtj8o81ryywg793: Given the comment, shouldn't we use the CSS variable (--carousel-duration) as the single source of truth?

static/js/providers-carousel.js
```diff
@@ -0,0 +1,82 @@
...
+    const DRAG_THRESHOLD = 5; // px of movement before treating as drag, not click
+
+    function baseDelay(i) {
+        return -(DURATION_S / NUM_ITEMS) * i;
+    }
+
+    function getItemCyclePx() {
+        const remPx = parseFloat(getComputedStyle(document.documentElement).fontSize);
```
awtj8o81ryywg793: Similar to this [comment](https://github.com/matrix-org/matrix.org/pull/3239/changes#r2861954751), I think we should use the CSS variable (--item-gap) as the single source of truth. If only to make the relationship explicit.

This is less of a concern that the other comment, however, given that we still get the same value.

static/js/providers-carousel.js
```diff
@@ -0,0 +1,82 @@
...
+        return 190 + remPx; // --item-width (190px) + --item-gap (1rem)
```
awtj8o81ryywg793: Similar to this [comment](https://github.com/matrix-org/matrix.org/pull/3239/changes#r2861954751), I think we should use the CSS variable (--item-width) as the single source of truth. If only to make the relationship explicit.

awtj8o81ryywg793 (commented):
Alongside the points raised in this review, I've noticed that the "Get an app" table isn't responsive.

templates/try-matrix.html
```diff
@@ -8,59 +11,240 @@ <h1>{{ page.title }}</h1>
...
+                    <label for="pf-all" class="platform-filter-btn">All</label>
+                    <label for="pf-ios" class="platform-filter-btn">iOS</label>
+                    <label for="pf-android" class="platform-filter-btn">Android</label>
+                    <label for="pf-web" class="platform-filter-btn">Web</label>
+                    <label for="pf-windows" class="platform-filter-btn">Windows</label>
+                    <label for="pf-macos" class="platform-filter-btn">macOS</label>
+                    <label for="pf-linux" class="platform-filter-btn">Linux</label>
+                </div>
+                <div class="client-comparison-wrapper">
+                    <table class="client-comparison">
```
HarHarLinks: I think introducing a table here complicates things and answers the wrong questions. 

Important is that one can just try some clients without issue at any time to find the most comfy one, and probably custom criteria like "try Cinny if you like discord"

MTRNord: My proposal would be:

grafik

MTRNord: Pushed it now :)

thibaultamartin:
This is an amazing redesign! Thanks for tackling it, the Try Matrix page has needed some love for a long time, and this redesign is made tastefully and efficiently!

## Simple instructions

On desktop, I like that the simple "Get Element X and create an account on matrix.org" instructions take the whole. It keeps things simple for people who just want to give it a go.

On mobile, the instructions are slightly larger than the screen as seen from an iPhone 11 Pro, a reasonably not too old device. Fiddling with the font-size, I can make the instructions fit in. Maybe it would make sense to `clamp` the font-size page-wise depending on the screen width? I don't mind a media or container query either.

## Clients list

Seeing all the featured clients in a simple and compact list is really great. It's nicely presented and less overwhelming than the regular view we have on /ecosystem/clients.

For the screenshots, I think we should enforce a specific aspect ratio. 4:3 seems to be an interesting one. For desktop, it's seamless. For mobile, we can ask to have a collage of 2 or 3 screenshots in the same image to achieve the same aspect ratio.

I think it could be interesting to have an anchor for each client as well, so it's possible to link to `/try-matrix#fluffychat`. It would make it easy to turn the "Get Element X" CTA into an anchor that points the user to the Element X entry on that same page.

## Bells & whistles

Two lines of CSS could go a long way:

- Adding `scroll-snap-type: y proximity;` would let the window scroll nicely to the most relevant content
- Adding `scroll-behavior: smooth;` would give us smooth scrolling when clicking the arrow. It would make it more obvious that we're staying on the same page.

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+                    <label for="pf-all" class="platform-filter-btn">All</label>
+                    <label for="pf-ios" class="platform-filter-btn">iOS</label>
+                    <label for="pf-android" class="platform-filter-btn">Android</label>
+                    <label for="pf-web" class="platform-filter-btn">Web</label>
+                    <label for="pf-windows" class="platform-filter-btn">Windows</label>
+                    <label for="pf-macos" class="platform-filter-btn">macOS</label>
+                    <label for="pf-linux" class="platform-filter-btn">Linux</label>
+                </div>
+
+                <div class="client-cards">
```
HarHarLinks, line 60: the client cards and their text is surprisingly small on mobile, and partially also on desktop

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
         </div>
     </div>
-    <div class="step-background coloured">
-        <div class="step choose-a-server">
+
+    <div class="step-background" id="get-an-app">
+        <div class="step choose-a-client no-image">
             <div class="instructions">
-                <h3>Choose a server</h3>
+                <h3>Get an app</h3>
```
HarHarLinks: these are also surprisingly small on mobile, and don't really distinguish from the `h4`s

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
-                    We recommend Element as a safe default, but there are a variety of clients available.
+                    Open Element in your browser — no download needed. You'll be on
+                    <strong>matrix.org</strong>, the server run by the Matrix.org Foundation,
+                    which is a fine place to start. If you're planning longer-term use or want
+                    your own domain, read through the steps below first.
                 </p>
                 <div class="cta-wrapper">
-                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
-                    <a href="/ecosystem/clients" class="call-to-action secondary">Browse clients</a>
+                    <a href="https://app.element.io" class="call-to-action">Open Element Web</a>
```
HarHarLinks: this is really not useful on a phone :sweat_smile:

we could try to just hide it in these cases, but then we are missing to give people the info that you can also just try on a pc browser. is that relevant?

sass/_try-matrix.scss
```diff
@@ -54,10 +64,802 @@
...
+        border-color: #7acc8a;
+    }
+    .platform-pill--web {
+        background: #ede9fe;
+        color: #5b21b6;
+        border-color: #c4b5fd;
+    }
+    .platform-pill--windows {
+        background: #e0f0ff;
+        color: #0061a4;
```
HarHarLinks: this is very close to the iOS color

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+                <input type="radio" name="pf" id="pf-linux" class="platform-filter-radio">
+                <div class="platform-filters">
+                    <label for="pf-all" class="platform-filter-btn">All</label>
+                    <label for="pf-ios" class="platform-filter-btn">iOS</label>
+                    <label for="pf-android" class="platform-filter-btn">Android</label>
+                    <label for="pf-web" class="platform-filter-btn">Web</label>
+                    <label for="pf-windows" class="platform-filter-btn">Windows</label>
+                    <label for="pf-macos" class="platform-filter-btn">macOS</label>
+                    <label for="pf-linux" class="platform-filter-btn">Linux</label>
+                </div>
```
HarHarLinks: @yncyrydybyl says
> I think the button list ontop is not needed.

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+                                    {% if feat.multi_account %}<span class="client-feat-pill">Multi-account</span>{% endif %}
+                                    {% if feat.oauth %}<span class="client-feat-pill">OAuth 2.0</span>{% endif %}
+                                    {% if feat.sso %}<span class="client-feat-pill">SSO</span>{% endif %}
+                                </div>
+
+                                <div class="client-card-downloads">
+                                    {{ clients::install_links(packages=client.extra.packages) }}
+                                </div>
+
+                                <a href="{{ client.path }}" class="client-card-more-link">Full details &rarr;</a>
```
HarHarLinks: @yncyrydybyl says
> The blue link in the details needs more style.

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+
+    <div class="step-background" id="get-an-app">
+        <div class="step choose-a-client no-image">
             <div class="instructions">
-                <h3>Choose a server</h3>
+                <h3>Get an app</h3>
                 <p>
-                    Matrix is a decentralised protocol: like email, you need to have an account provider to start using
-                    matrix.
+                    To use Matrix you need an app. Here are the well-established featured
```
anoadragon453:
```suggestion
                    To use Matrix you need a client. Here are the well-established featured
```

We should be consistent and use one term to describe them, otherwise folks will wonder if there's a difference between a "client" and an "app".

Some people may also be confused and think "app" only means "an app on my phone", rather than a website.

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+                                    <li>100 MB data / day</li>
+                                </ul>
+                                <a href="/docs/chat_basics/matrix-for-im/#creating-a-matrix-account" class="call-to-action secondary">How to create an account</a>
+                            </div>
+                            <div class="plan-card featured">
+                                <h5>Premium</h5>
+                                <ul>
+                                    <li>100 MB max attachment size</li>
+                                    <li>1 GB data / day</li>
+                                    <li>Faster room invitations</li>
```
thibaultamartin:
```suggestion
                                    <li>Lower room invite limits</li>
```

This one might be a bit misleading, as it can be read as "faster room joins"

templates/try-matrix.html
```diff
@@ -8,59 +17,250 @@ <h1>{{ page.title }}</h1>
...
+                                    <li>Foundation members: included</li>
```
thibaultamartin:
```suggestion
                                    <li>Free for Foundation members</li>
```

I think it makes more sense like this?

templates/try-matrix.html
MTRNord: Yeah maybe it is too much for so little amount.

This was mainly inspired by https://joinmastodon.org/apps

templates/try-matrix.html
MTRNord: Yeah I applied that now but I wonder about the title if that makes sense or not. I can a reason for both. App simplifies it a little I guess but I can see your arguments too.

templates/try-matrix.html
MTRNord: grafik

templates/try-matrix.html
MTRNord: Got rid of it now

sass/_try-matrix.scss
MTRNord: Fixed

templates/try-matrix.html
MTRNord: Hidden on mobile now but both shown on desktop.

templates/try-matrix.html
MTRNord: I hope this is better now.

templates/try-matrix.html
MTRNord: Need more details here I think. Not sure what exactly you mean. It feels fine on my side

templates/try-matrix.html
HarHarLinks: Should match https://matrix.org/homeserver/pricing/

templates/try-matrix.html
HarHarLinks: Yeah I think it's fine as long as we have only a handful

templates/try-matrix.html
MTRNord: yeah thats where "faster" is from. So now I am confused :D thanks for mentioning it again since I overlooked that when applying this fix.

thibaultamartin: Yes, the item is "Speed at which you can invite others to a room", or in other words: "How aggressively you're rate-limited" :)

A friend read "Faster room invitation" as "Faster room joins" not as "Lower rate-limits." I don't think we can use "Lower rate-limits" because it's rather technical.

There might be other more straightforward ways to phrase it though!

HarHarLinks: "room invite limits" also sounds like "you can only have up to 10 ppl in your rooms".

matching the wording here means that also /pricing could be adjusted.

thibaultamartin: That's a very good point. Maybe something along the lines of "Invite more people at once", which is not technically accurate but close enough from a product perspective?

templates/try-matrix.html
anoadragon453: Oh, good point. I think the same applies actually; I have a slight preference for "client" there as well. And "Switching between apps ..."

Edit: And the "Browse All Apps" button :D

sass/_try-matrix.scss
```diff
@@ -54,10 +78,771 @@
...
+            img {
+                height: 36px;
+                width: auto;
+                max-width: none;
+                margin: 0;
+                border-radius: 6px;
+
+                @media (max-width: 700px) {
+                    height: 30px;
+                }
```
awtj8o81ryywg793, line 519: I think the platform buttons are too small on mobile, which makes them hard to read.

Image

sass/_try-matrix.scss
```diff
@@ -54,10 +78,771 @@
...
+            font-size: 0.75rem;
+            line-height: 1;
+            background: rgba(0, 0, 0, 0.55);
+            color: #fff;
+            padding: 0.2rem 0.35rem;
+            border-radius: 4px;
+            pointer-events: none;
+            opacity: 0.85;
+            z-index: 1;
+        }
```
awtj8o81ryywg793, line 357: I think we should increase the size of the trigger, especially on small viewports.

Image

awtj8o81ryywg793 (commented):
Here are a few suggestions, focused on the experience on small viewports.

Other than the included comments, we also have a text positioning issue on CTA buttons:

Image

sass/_try-matrix.scss
MTRNord: I increased it a bit. But in practice clicking anywhere on the image should work. The icon in the lower right is not actually the touch point. The image is.

sass/_try-matrix.scss
MTRNord: should be better now hopefully

MTRNord:
@awtj8o81ryywg793 

> Other than the included comments, we also have a text positioning issue on CTA buttons:

Thats fixed by using clamp now so it doesnt overflow as likely. I also centered the text within now.

MTRNord:
Note: Missing/blocking is that I also add the missing screenshots.

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
+                        </p>
+                        {% if page.extra.matrixorg_premium %}
+                        <div class="matrixorg-plans">
+                            <div class="plan-card">
+                                <h5>Free</h5>
+                                <ul>
+                                    <li>10 MB max attachment size</li>
+                                    <li>100 MB data / day</li>
+                                </ul>
+                                <a href="/docs/chat_basics/matrix-for-im/#creating-a-matrix-account" class="call-to-action secondary">How to create an account</a>
```
weeman1337: Most other buttons are actions. This one is not. Maybe it should be something like „Read account creation details“.

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
+                    Once you have an account, explore <a href="/docs/chat_basics/public-rooms">public
+                    rooms and Spaces</a> to find communities built around topics you care about.
+                    You can message anyone on Matrix directly using their address
+                    (<code>@username:server</code>) — no matter which server or app they use.
                 </p>
                 <div class="cta-wrapper">
-                    <a href="/docs/chat_basics/matrix-for-im/#creating-a-matrix-account" class="call-to-action">Create
-                        an account</a>
-                    <a href="/docs/chat_basics" class="call-to-action secondary">Read the docs</a>
+                    <a href="/docs/chat_basics/matrix-for-im" class="call-to-action">Chat basics guide</a>
```
weeman1337: This is one of the other buttons that not sound like an action. Maybe „Read the full guide“ or something like this.

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
-                    We recommend Element as a safe default, but there are a variety of clients available.
+                    Open Element in your browser — no download needed. You'll be on
+                    <strong>matrix.org</strong>, the server run by the Matrix.org Foundation,
+                    which is a fine place to start. If you're planning longer-term use or want
+                    your own domain, read through the steps below first.
                 </p>
                 <div class="cta-wrapper">
-                    <a href="/ecosystem/clients/element/" class="call-to-action">Install Element</a>
-                    <a href="/ecosystem/clients" class="call-to-action secondary">Browse clients</a>
+                    <a href="https://app.element.io" class="call-to-action hide-on-mobile">Open Element Web</a>
```
weeman1337: I would open this in a `target="_blank"`. If I have clicked that button it would be great to still have the „Try Matrix“ page open to read the rest.

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
+                    <a href="/ecosystem/clients/element-x/" class="call-to-action secondary show-primary-on-mobile">Get Element X (mobile)</a>
```
weeman1337: Same with `target="_blank"` here

weeman1337 (commented):
I like the new TL;DR

Other than that only some smaller notes.

templates/try-matrix.html
MTRNord: We did discuss this in the last meeting. I applied another adjustment for this. I hope this is a good middle ground now :)

templates/try-matrix.html
MTRNord: Hopefully this works now :D we may want a followup ticket for the other page though.

HarHarLinks: ideally we want alignment with the payment service on account.matrix.org, too.

HarHarLinks:
can we get more anchors so i can send people e.g. to the pick a server section?

templates/try-matrix.html
HarHarLinks: The font size seems to change somewhat arbitrarily all over the place, and there is even some inconsistency (e.g. between "good for" text and the m.org homeserver details.

Maybe let's take a step back:

1. What is the goal?
2. Can it be ~1 consistent size?
3. (Should text in general be larger on m.org?)

sass/_try-matrix.scss
```diff
@@ -54,10 +78,778 @@
...
+    }
+    .platform-pill--linux {
+        background: #fff4e5;
+        color: #92400e;
+        border-color: #fcd9a0;
+    }
+
+    // ── Account tiers (Create an account section) ─────────────────────────────
+
+    .account-tiers {
```
HarHarLinks, line 589: this is starting to get nitpicky, but the nested boxes on mobile are wasting space with double padding, wondering if that could be avoided responsively.

image

sass/_try-matrix.scss
```diff
@@ -54,10 +78,778 @@
...
+            margin-bottom: 0.5rem;
+        }
+
+        > p {
+            max-width: 42rem;
+            font-size: clamp(1rem, 3.5vw, 1.2rem);
+        }
+    }
+
+    .scroll-down-indicator {
```
HarHarLinks, line 137: would it be within reason so try and always have both the CTA and this on screen on mobile? The iPhone 11 simulation with my custom root font size of 18 gives me this thanks to the header wasting 50% of v-space:
image

HarHarLinks: The no download needed text is also BS on mobile.

templates/try-matrix.html
anoadragon453: > client app

Love it. Happy with the current state.

thibaultamartin:
@MTRNord This is a great PR already and I wouldn't want to let it bitrot, do you have some bandwidth to tackle it or would you rather someone else takes over :) ?

MTRNord:
> @MTRNord This is a great PR already and I wouldn't want to let it bitrot, do you have some bandwidth to tackle it or would you rather someone else takes over :) ?

I would love if someone can help with the client screenshots. I am kinda lacking motivation for that currently :/

So yes help appreciated!

HarHarLinks:
(https://github.com/AlexanderVanhee/Gradia looks like it might create very pretty screenshots)

content/ecosystem/clients/cinny.md
```diff
@@ -11,6 +11,9 @@ repo = "https://github.com/cinnyapp/cinny"
 website = "https://cinny.in"
 matrix_room = "#cinny:matrix.org"
 featured = true
+featured_order = 4
+screenshots = []
+good_for = "Users coming from Discord who want a familiar layout"
```
HarHarLinks, line 17: please add to CONTENT.md

sass/_try-matrix.scss
```diff
@@ -54,10 +76,778 @@
...
+        max-height: 100%;
+        width: auto;
+        height: auto;
+        object-fit: contain;
+        object-position: center;
+        border-radius: 4px;
+    }
+
+    // Multi-screenshot layout — images sit side by side, each fills its slot
+    // with cover so the 4:3 container is always full
```
HarHarLinks, line 372: Hm obviously contain isn't great either but cover means images are cut off 
image

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
+                                    {% if feat.multi_account %}<span class="client-feat-pill">Multi-account</span>{% endif %}
+                                    {% if feat.oauth %}<span class="client-feat-pill">OAuth 2.0</span>{% endif %}
+                                    {% if feat.sso %}<span class="client-feat-pill">SSO</span>{% endif %}
+                                </div>
+
+                                <div class="client-card-downloads">
+                                    {{ clients::install_links(packages=client.extra.packages) }}
+                                </div>
+
+                                <a href="{{ client.path }}" class="client-card-more-link">Full details &rarr;</a>
```
HarHarLinks: How about
```suggestion
                                <a href="{{ client.path }}" class="client-card-more-link">View more details</a>
```

content/ecosystem/clients/cinny.md
MTRNord: Added

templates/try-matrix.html
```diff
@@ -8,59 +17,242 @@ <h1>{{ page.title }}</h1>
...
-            <div class="instructions">
-                <h3>Choose a client</h3>
+    <div class="step-background quickstart">
+        <div class="step no-image">
+            <div class="instructions quickstart-instructions">
+                <h2>Just want to try it?</h2>
                 <p>
-                    We recommend Element as a safe default, but there are a variety of clients available.
+                    Open Element in your browser — no download needed. You'll be on
+                    <strong>matrix.org</strong>, the server run by the Matrix.org Foundation,
```
HarHarLinks: pls ctrl+f that the text conforms to the /branding spelling guide

sass/_try-matrix.scss
MTRNord: Fixed

sass/_try-matrix.scss
MTRNord: Kinda improved. Not perfect but better than before I think?

templates/try-matrix.html
```diff
@@ -8,59 +17,248 @@ <h1>{{ page.title }}</h1>
...
 <div class="try_matrix">
-    <div class="step-background">
-        <div class="step choose-a-client">
+    <div class="step-background quickstart">
+        <div class="step no-image">
+            <div class="instructions quickstart-instructions">
+                <h2>Just want to try it?</h2>
+                <p class="desktop-only">
+                    Open Element in your browser — no download needed. You'll be on
+                    <strong>matrix.org</strong>, the server run by the The Matrix.org Foundation C.I.C.,
```
HarHarLinks:
```suggestion
                    <strong>matrix.org</strong>, the server run by The Matrix.org Foundation,
```
bruh

CIC is also ok but using the fill name seems a bit extra here

templates/try-matrix.html
```diff
@@ -8,59 +17,248 @@ <h1>{{ page.title }}</h1>
...
+                    which is a fine place to start. If you're planning longer-term use or want
+                    your own domain, read through the steps below first.
+                </p>
+                <p class="mobile-only">
+                    Get started using Element X. You'll be on
+                    <strong>matrix.org</strong>, the server run by the The Matrix.org Foundation C.I.C.,
```
HarHarLinks:
```suggestion
                    <strong>matrix.org</strong>, the server run by The Matrix.org Foundation,
```

templates/try-matrix.html
```diff
@@ -155,7 +161,7 @@ <h3>matrix.org homeserver</h3>
                             <span class="tier-badge foundation">Foundation-run</span>
                         </div>
                         <p>
-                            The Matrix.org Foundation hosts a server open to everyone.
+                            The Matrix.org Foundation C.I.C. hosts a server open to everyone.
```
HarHarLinks: as above, some nuance is justified imo

HarHarLinks (changes requested):

templates/try-matrix.html
```diff
@@ -8,59 +17,248 @@ <h1>{{ page.title }}</h1>
...
+    <div class="step-background" id="get-an-app">
+        <div class="step choose-a-client no-image">
             <div class="instructions">
-                <h3>Choose a client</h3>
+                <h2>Get a client app</h2>
                 <p>
-                    We recommend Element as a safe default, but there are a variety of clients available.
+                    To use Matrix you need a client. Here are the well-established featured
+                    clients — pick whichever fits your platform and needs. Switching between
+                    apps is easy since they all use the same account.
```
HarHarLinks: I just thought of this, maybe it could be worded more eloquently. @awtj8o81ryywg793 any ideas?
```suggestion
                    apps is easy and can be done at any time, since they all use the same account.
                    You can try multiple at the same time no problem until you find what works best for you.
```

HarHarLinks (commented):
Trying it on my phone, it feels pretty good. 

I noticed the "close" x in the lightbox is always top right of the screen, i.e. sometimes very far from the actual image or even over it's border. Could maybe improve the placement some time.
Screenshot_20260409-174408


Also the lightbox seems to make it into browsing history, so browser navigation after checking out all the lightboxes is pretty annoying.

These are pretty minor issues I wouldn't block on though.

templates/try-matrix.html
awtj8o81ryywg793: Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3239/changes#r3059011751):

> I just thought of this, maybe it could be worded more eloquently. 

How does the following read?

> To use Matrix you need a client. Below are well‑established, featured clients — choose the one that best suits your platform and needs. Switching between them is effortless, as all clients use the same account, and you can run multiple clients simultaneously until you find your preferred option.

HarHarLinks: My stab at iterating further:
> To use Matrix you need a client. Below are some well‑established, featured clients — choose the one that best suits your platform and needs. You can switch between them at any time, as all clients use the same account, and you can run multiple clients simultaneously to find your preferred option.

MTRNord: Comitted Kim's variant in 09430a1

HarHarLinks (approved):

PR #3262, opened 2026-03-06
Add a basic css-only lightbox to the figure shortcode

MTRNord:
### Description

Adds the lightbox from #3239 to the figure shortcode

### Related issues

None

### Role

Website & Content WG

### Timeline

None

### Signoff

See commits.

### DIFF
--- a/sass/_lightbox.scss
+++ b/sass/_lightbox.scss
@@ -1,0 +1,91 @@
+// ── CSS-only lightbox via :target ─────────────────────────────────────────────
+//
+// Usage (in a Zola shortcode or template):
+//
+//   <!-- trigger -->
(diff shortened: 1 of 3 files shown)


HarHarLinks:
this isn't working out on https://88e0cbc6.matrix-website.pages.dev/homeserver/official/#/ :thinking:

MTRNord:
> this isn't working out on https://88e0cbc6.matrix-website.pages.dev/homeserver/official/#/ 🤔

Because they are too low res or why exactly?

HarHarLinks:
Yeah, arguably the images in that case should just be bigger anyway. But I don't think we can generalise that pictures are always significantly bigger than displayed? In which cases the experience is a bit ridiculous. I can't think of a solution, except
- manual flag for lightbox
- lots of JS?
- (bigger pictures for /homeserver/official)

PR #3265, opened 2026-03-06
Make build.sh executable

MTRNord:
### Description

Does what the title says. It annoyed me :D

### Role

Website & Content WG

### Signoff

Commits!

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ "$CF_PAGES_BRANCH" == "main" ]; then
   # build production using base_url from the site config.toml


HarHarLinks:
Why is this even needed, and why does the shebang contain a space?

MTRNord:
> Why is this even needed, and why does the shebang contain a space?

I have no idea :/ I am guessing cloudflare uses it? I dont know really. I just saw it and got annoyed by it not having bee 755 yet.

MTRNord:
The shebang being broken probably always was a bug but doesnt matter the way cloudflare calls that file?

build.sh
```diff
@@ -1,4 +1,4 @@
-# !/bin/bash
+#!/bin/bash
```
weeman1337: Let's fix it properly:

```suggestion
#!/usr/bin/env bash
```

weeman1337 (approved):

HarHarLinks:
(I suggest not to merge this without anyone who actually touches cloudflare in the loop about it)

guillaumevillemont:
For some reason, we configured the build to run `bash build.sh` on Cloudflare so I don't think the shebang matters much.
This branch build was successful, no warnings in the logs. so :+1:

PR #3267, opened 2026-03-06
Improve Blog Post layout

MTRNord:
### Description

This does a few things:

- The flexbox is replaced with a grid. This fixes some of the spacing weirdness we got due to using `flex: 3;`. This is minor and not visually really obvious
- The space to the aside is a bit smaller
- The metadata is a bit more muted
- The blockquotes are modernized to be less... blocky? Idk it feels a lot nicer especially on twim where we got lots of them.

### Role

Website & content WG

### Signoff

In the commits

### DIFF
--- a/templates/blog.html
+++ b/templates/blog.html
@@ -31,7 +31,7 @@
             <small>Last update: {{ page.updated | date(format="%Y-%m-%d %H:%M") }}</small>
             {%- endif %}
         </header>
-        <div>
+        <div class="post-content">
             {% if page.summary -%}
             {{ page.summary | safe }}
             <p><a href="{{ page.path | safe }}#continue-reading">Continue reading…</a></p>


sass/_blog.scss
```diff
@@ -144,29 +154,53 @@ article.post {
...
+        padding-left: 2rem;
+        margin: 0 0 1rem 0;
+
+        &::before,
+        &::after {
+            position: absolute;
+            color: #777;
+        }
+
+        // "//" rotated 90deg sits naturally at top-left
```
HarHarLinks, line 187: Does this mean anything that is going over my head or is it just stylish?

It's now inconsistent with pages such as /blog

MTRNord: Its just stylish.

I didnt want to use `"` as it doesnt really feel right for twim.

and I did overlook /blog indeed oops. Good catch.

MTRNord: The inconsistency on /blog is now fixed.

HarHarLinks: are you intentionally not applying this globally?

https://d29b8cfb.matrix-website.pages.dev/foundation/working-groups/website-content/#charter

sass/_blog.scss
```diff
@@ -144,29 +154,53 @@ article.post {
...
+            left: 0;
+        }
+
+        // Thin vertical line below the slashes, clipped by overflow: hidden
+        &::after {
+            content: "";
+            left: 5px;
+            top: 40px;
+            width: 1px;
+            height: 100%;
```
HarHarLinks, line 200: this line at 100% height has the combined lineheight of the text plus 10px top and bottom padding as defined somewhere according to my devtools. you are moving it 10px+30px=40px down, which means it exceeds the blockquote's container.
```suggestion
            height: calc(100% - 50px);
```
would bring that back so the line actually ends with the last line instead of the blockquote element. imo it looks a bit leaner. wdyt?

image

sass/_blog.scss
```diff
@@ -144,29 +154,53 @@ article.post {
...
+
+        &::before,
+        &::after {
+            position: absolute;
+            color: #777;
+        }
+
+        // "//" rotated 90deg sits naturally at top-left
+        &::before {
+            content: "//";
```
weeman1337, line 189: This looks a bit strange imho. Is this a new common pattern? I think that quotes with just the | make the layout look smoother.

PR #3268, opened 2026-03-06
Anchor improvements

MTRNord:
### Description

| Before | After |
| :--------: | :-----: |
| grafik | grafik |

### Role

Website & Content WG

### Signoff

See commits

### DIFF
--- a/sass/_blog.scss
+++ b/sass/_blog.scss
@@ -196,7 +196,8 @@ h3 a {
 }
 
 .zola-anchor {
-    font-size: 1rem;
-    margin-right: 0.5rem;
-    color: inherit;
+    font-size: 0.85em;
+    margin-left: 0.4rem;
+    color: #aaa;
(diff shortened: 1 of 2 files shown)


MTRNord:
(Codeowners auto added legal team but I dont think their review is required :) )

HarHarLinks:
I would be ok with this, certainly an improvement. I noticed it does not completely match the behavior from your proposal https://github.com/matrix-org/matrix.org/pull/3239/files#diff-a33aab1446a824fc5a4e0bce56ae5d9c1ba7b1a3ddd245538ad003f04c9f4dd9R272-R274

What about putting them into the left margin (on desktop view, i.e. when there's space), but to the right on mobile, like MDN does?

MTRNord:
Note from meeting: maybe use pseudo elements to have things still linksbündig

Issue #1758, opened 2023-04-29
improve inline code style


HarHarLinks:
# [zola]

currently inline code markup uses red color monospace font plus a red frame around it.

meanwhile code blocks/parapgraphs use a "dark mode" look.

- i think red is not the best choice, because, being a signal color, it implies an error or something similarly bad.
- i think the looks of inline and block code should be consolidated to a common style basis.

nathanael-h:
Hi, I support this issue. At first I thought these red texts were errors. Would you be interested in a small PR just for a color change? Or would you prefer to address this issue as part of a broader enhancement to the inline code style?

AdilDyer:
Hey would you pls provide the steps to reproduce ?

hughns:
A big +1 on improving this from me. I think colour is most important.

> Hey would you pls provide the steps to reproduce ?

Some examples of pages withe a inline code:

https://matrix.org/homeserver/about/

<img width="998" height="180" alt="Image" src="https://github.com/user-attachments/assets/efe4d705-9213-4a01-9b6c-0188c5f9470b" />

https://matrix.org/ecosystem/hosting/

<img width="1036" height="280" alt="Image" src="https://github.com/user-attachments/assets/34288579-7b5e-44c6-8e3b-3d682f4143da" />

https://matrix.org/homeserver/official/

<img width="997" height="635" alt="Image" src="https://github.com/user-attachments/assets/f3fa8b70-9e03-4a02-b0a8-75a328d31958" />

hughns:
For the colour, something like this looks "okay":

```
diff --git a/sass/_base.scss b/sass/_base.scss
index b370a4998..fc80c6b1f 100644
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -262,8 +262,9 @@ pre table {
 }
 
 :not(pre)>code {
-    color: #f54029;
-    border: 1px solid #f54029;
+    color: color-mix(in srgb, currentColor, transparent 10%);
+    background-color: color-mix(in srgb, currentColor, transparent 95%);
+    border: 1px solid color-mix(in srgb, currentColor, transparent 50%);
     border-radius: 4px;
     padding: 0 .4rem;
 }
```

But I don't know if the [availability](https://caniuse.com/?search=color-mix) of `color-mix` is sufficient.

Looks like this:

<img width="912" height="587" alt="Image" src="https://github.com/user-attachments/assets/ac370798-ac2c-469a-9509-a7fdda08a704" />

[...]

PR #3269, opened 2026-03-06
Better inline code

MTRNord:
### Description

I am not sure if this is actually better or not. Happy to hear feedback on this though.

Personally the red always felt a bit out of place. This is now a bit closer to the big blocks but not sure if thats a good thing.

I also tried to size match the text around it a bit better than before.

grafik

### Role

Website & Content WG.

### Signoff

See commits

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -1,3 +1,6 @@
+$color-code-background: #272822;
+$color-code-color: #f8f8f2;
+
 :root {
     --page-max-width: 57rem;
     --page-with-toc-max-width: 81rem;
@@ -117,9 +120,10 @@ ol:last-child {
 }
 
 pre {
+    background-color: $color-code-background;
     padding: 1rem;
     overflow: auto;
-    border-radius: 4px;
+    border-radius: 0.3rem;
 }
 
 // The line numbers already provide some kind of left/right padding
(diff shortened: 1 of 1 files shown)


sass/_base.scss
```diff
@@ -206,15 +210,31 @@ pre table {
...
-    color: #f54029;
-    border: 1px solid #f54029;
-    border-radius: 4px;
-    padding: 0 .4rem;
+code {
+    background-color: $color-code-background;
+    color: $color-code-color;
+    border-radius: 0.3rem;
+    padding: 1px 5px 2px;
+    font-size: 0.8em;
```
awtj8o81ryywg793, line 218:
```suggestion
```

I don't think we need to decrease the size of the text. Keeping it the same size ensures consistency, which I believe is something we should strive for in this situation.

awtj8o81ryywg793 (commented):
> Personally the red always felt a bit out of place. This is now a bit closer to the big blocks but not sure if thats a good thing.

I'd say it's a good thing. It helps make everything feel more cohesive. Currently, the red styling suggests that something is broken.

sass/_base.scss
MTRNord: The reason here is that with all the border and background it always felt larger than the rest. Hence this tried to visually match the surroundings. But maybe thats just me

HarHarLinks:
Pls check https://github.com/matrix-org/matrix.org/issues/1758 and #2945 too

HarHarLinks:
Some vertical misalignment on my Android phone:
Screenshot_20260307-164854

sass/_base.scss
```diff
@@ -206,15 +210,31 @@ pre table {
...
-:not(pre)>code {
-    color: #f54029;
-    border: 1px solid #f54029;
-    border-radius: 4px;
-    padding: 0 .4rem;
+code {
+    background-color: $color-code-background;
+    color: $color-code-color;
+    border-radius: 0.3rem;
+    padding: 1px 5px 2px;
```
weeman1337, line 217: We need to talk about units in general. If you use absolute units for the paddings and relative units for the font size, then people with other base font sizes will have misaligned code blocks.

sass/_base.scss
weeman1337: Maybe this one is subjective. For me the .8 makes inline code hard(er) to read in text blocks because the font size changes in the flow.

HarHarLinks:
Note that we are due a zola upgrade in which we will lose the currently employed theme, changing the look of code blocks generated from markdown.

I think apart from my comment there, I prefer #2945.

PR #3285, opened 2026-03-16
Add Zendrite homeserver

bones-was-here:
### Description

Adds the Zendrite homeserver

### Role

Individual

### Timeline

### Signoff

### DIFF
--- a/content/ecosystem/servers/servers.toml
+++ b/content/ecosystem/servers/servers.toml
@@ -39,6 +39,17 @@ repository = "https://github.com/element-hq/dendrite"
 room = "#dendrite:matrix.org"
 
 [[servers]]
+name = "Zendrite"
+description = "Zendrite is a second-generation Matrix homeserver, forked from Dendrite with new MSCs and optimisations."
+author = "Patrick Schratz and Dendrite/Zendrite contributors"
+maturity = "Stable"
+language = "Go"
+licence = "AGPL-3.0-or-later"
+website = "https://zendrite.pat-s.me/"
+repository = "https://codefloe.com/pat-s/zendrite"
+room = "#dendrite:matrix.org"
+
+[[servers]]
 name = "Conduit"
 description = "Conduit is a simple, fast and reliable chat server written in Rust"
 author = "Timo Kösters"


bones-was-here:
@pat-s what do you think? I reckon it could be marked Stable, based on comparison with some of the other listed servers (would need to update the docs site too).

HarHarLinks:
Thanks for the PR!

@pat-s, does this make sense to you? :)

pat-s:
Yes, I think (and hope!) it falls under the "stable" category, given all the fixes we put on top of Dendrite. Thanks @bones-was-here!

content/ecosystem/servers/servers.toml
```diff
@@ -38,6 +38,17 @@ licence = "AGPL-3.0-or-later OR Element Commercial License"
...
+[[servers]]
+name = "Zendrite"
+description = "Zendrite is a second-generation Matrix homeserver, forked from Dendrite with new MSCs and optimisations."
+author = "Patrick Schratz and Dendrite/Zendrite contributors"
+maturity = "Stable"
+language = "Go"
+licence = "AGPL-3.0-or-later"
+website = "https://zendrite.pat-s.me/"
+repository = "https://codefloe.com/pat-s/zendrite"
+room = "#dendrite:matrix.org"
```
MTRNord, line 50: I would rather leave it out than reusing another server's room here even though it is a fork. This feels misleading and will cause confusion to people expecting people ask about dendrite but then end up asking about Zendrite.

MTRNord (dismissed):

content/ecosystem/servers/servers.toml
pat-s: I have no strong opinion on that. Just wanna note that the channel members were asked if they wanna have both in the same room and the consensus was "yes". 

But yes, it's likely confusing and maybe a distinct room would be better long-term.

HarHarLinks: I think it's fine for projects to share a room, *if* it is recognised officially enough as both projects' room to avoid the confusion MTRNord mentions. A definite way to do so, besides room name, could be room description.

clokep: Rooms can also have multiple aliases to solve that.

thibaultamartin: Got a grünes Licht from Matthew that the project can reuse the room alias

thibaultamartin (approved):

HarHarLinks:
Sorry this appears to be stuck so long; for visibility we'd like to share that we are waiting on an ongoing decision by the Governing Board about how to classify server project maturity.

bones-was-here:
> Sorry this appears to be stuck so long; for visibility we'd like to share that we are waiting on an ongoing decision by the Governing Board about how to classify server project maturity.

Why weren't the other homeservers subject to this requirement?

HarHarLinks:
The only server added since the start of this classification project is Venator, which both claims the lowest maturity at Alpha and whose description is abundantly clear about being experimental.

We have to make the cut somewhere and hoped it would be quicker. I'd be much less reluctant if the claim here wasn't Stable. It would also help if there was proof that Zendrite is at least as stable as Dendrite.

pat-s:
If Dendrite is considered stable, Zendrite is stable^2. 

I don't care much about the classification tbh. Feel free to change is to one that fits in your eyes. 
At the same time I personally don't want to fill out a checklist or go through a possible multi-stage evaluation just to proof a "status". The main point for this effort was to make users aware in general and improve discoverability. 

An alternative in my eyes could also be to mention that the status is "being evaluated" or similar and that users must evaluate on their own if they wanna use it (until an "official" status has been determined).

bones-was-here:
Other homeservers were added with the "stable" maturity when they still had significant stability problems, problems of the kind zendrite doesn't have. Dendrite was more stable than some of the other servers too, at least at the time they were added with that status (it's also much more stable than early versions of dendrite from back when its "beta" status was chosen).

I'm not sure the maturity is "being evaluated", it sounds to me like they don't want to do that work or don't have time (and don't even have criteria), but also don't want to take anyone else's word. Blocking indefinitely would be double standards though.

A more pragmatic way could be adding a line to this page noting that the maturity status of each server is as claimed by its maintainer, and isn't validated by matrix.org.

content/ecosystem/servers/servers.toml
```diff
@@ -38,6 +38,17 @@ licence = "AGPL-3.0-or-later OR Element Commercial License"
 repository = "https://github.com/element-hq/dendrite"
 room = "#dendrite:matrix.org"
 
+[[servers]]
+name = "Zendrite"
+description = "Zendrite is a second-generation Matrix homeserver, forked from Dendrite with new MSCs and optimisations."
+author = "Patrick Schratz and Dendrite/Zendrite contributors"
+maturity = "Stable"
```
HarHarLinks, line 45:
```suggestion
maturity = "Beta"
```

HarHarLinks (changes requested):
As you know, neither us nor Element claims Dendrite to be stable, in fact I'm not sure it has fallen behind enough to not even call it beta anymore. Part of the problem is that we don't have a definition and hence everyone has different expectations what "maturity" even means. This, mixed with taking someone's word is at the core of the issue we are trying to resolve, particularly with CoI involved. Unfortunately simply putting up a disclaimer would mostly help us deflect responsibility instead of helping readers get a realistic impression of the project landscape.

Neither project being evaluated right now, but the Governing Board is working on defining high level criteria which are as objective as possible to make it easier for readers to evaluate which project would fit their needs best. Unfortunately, determining how well and reliably a server fulfills the expectations one might have towards a "homeserver" software isn't as simple as it may sound. We are happy to have added more voices with expertise in the area of homeservers to the board recently, which will help us ensure the criteria will be as fair as can be.

Unfortunately the maturity project, while evolving, is as much on the "soon" horizon as it was in May. As you say, we do not want to delay indefinitely but were hoping this would be quicker, so thanks for bringing it up again. I'm sorry that your project fell into this crack. We propose to merge Zendrite to the list classified as Beta, same as Dendrite.

Issue #3222, opened 2026-02-18
Advertise TWIM on the index page


HarHarLinks:
TWIM is an amazing resource where you get all the latest info but also the room is a great place to learn about stuff.

However it is hard to discover, and since we only call it TWIM, not easy to get started with.

We maybe want a small section about the blog and what's in there on the homepage.

MTRNord:
From the meeting at 2026W08-4:

There should be a CTA and @weeman1337 is going to make a proposal

PR #3290, opened 2026-03-20
Rework the frontpage to be more concise and useful(?)

MTRNord:
### Description

This is mostly an idea right now. I am not sure if it is actually an improvement. Its a change. Open for opinions. Happy to drop if not the right direction.

image

### Related issues

no idea

### Role

Individual or Website & Content WG

### Timeline

whenever

### Signoff

see commits.

### DIFF
--- a/sass/_index.scss
+++ b/sass/_index.scss
@@ -530,12 +530,17 @@
         .membership-card {
             max-width: 648px;
             padding: 28px;
-            border-radius: 16px;
+            border-radius: 14px;
             border: 1px #d2d2d2 solid;
             display: flex;
             flex-direction: column;
             text-align: center;
             gap: 28px;
+            transition: border-color 0.2s ease;
+
+            &:hover {
+                border-color: #888;
+            }
         }
 
         .membership-card__content {
--- a/templates/index.html
+++ b/templates/index.html
@@ -132,11 +132,11 @@
     <div id="platinum-sponsor" class="section">
         <h2>Platinum member</h2>
         <div class="row">
-            <div class="membership-card">
-                <a target="_blank" rel="noreferrer noopener" href="https://element.io/built-on-matrix" class="membership-card__content">
+            <a target="_blank" rel="noreferrer noopener" href="https://element.io/built-on-matrix" class="membership-card">
+                <div class="membership-card__content">
                     <img alt="Element.io Logo" src="images/element-creators-of-matrix--light.svg">
-                </a>
-            </div>
+                </div>
+            </a>
         </div>
     </div>
 


HarHarLinks:
#3222

MTRNord:
@HarHarLinks commented:

> #3222

Have a look at [09c3195](https://github.com/matrix-org/matrix.org/pull/3290/commits/09c3195e7f16558f9da250dfcc702605d20a28df) :)

awtj8o81ryywg793:
I wonder if the background of the "Platinum member" section should be light grey (`#f8f8f8`) rather than white? Perhaps the "Become a member" section instead?

HarHarLinks:
The platinum member being a clickable tile should also use the new clickable tile UX? (if only this was a component)

templates/index.html
```diff
@@ -8,48 +8,133 @@
...
+                <h2>This Week in Matrix</h2>
+                <p>A community-curated weekly digest of the latest news, releases, and projects from across the Matrix ecosystem. Find out what's happening, discover new apps and servers, and share your own work in <a href="https://matrix.to/#/#thisweekin:matrix.org">#thisweekin:matrix.org</a>.</p>
+                <div class="cta-row twim__ctas">
+                    <a href="/category/this-week-in-matrix/" class="call-to-action">Browse all editions</a>
+                    <a href="/twim-guide" class="call-to-action secondary">How to contribute</a>
+                </div>
+            </div>
+            {% set_global twim_found = false %}
+            {% for page in blog.pages %}
+            {% if not twim_found and page.taxonomies.category and "This Week in Matrix" in page.taxonomies.category %}
```
HarHarLinks, line 92: neat!

MTRNord:
> I wonder if the background of the "Platinum member" section should be light grey (`#f8f8f8`) rather than white? Perhaps the "Become a member" section instead?

The problem is the graphics at the bottom only work this way so the idea was to have the last 2 sections before that the same way since they are technically meant to be like one block. Its kinda cheating of course. The alternative is to add another section I guess :p open for ideas though.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3290#issuecomment-4101125011):

> The problem is the graphics at the bottom only work this way so the idea was to have the last 2 sections before that the same way since they are technically meant to be like one block. Its kinda cheating of course. The alternative is to add another section I guess :p open for ideas though.

Here's what I was thinking, at least for my first [suggestion](https://github.com/matrix-org/matrix.org/pull/3290#issuecomment-4098817349).

image

templates/index.html
```diff
@@ -8,50 +8,135 @@
...
-            alt="A conversation between four people on Element Call, a VoIP conference tool based on Matrix." />
     </div>
+
+    <div id="blog-preview" class="section">
+        <h2>Latest from the Blog</h2>
+        <p class="blog-preview-description">Stay up to date with the latest news, updates, and insights from the Matrix community.</p>
+        <div class="blog-preview-grid">
+            {% set_global blog_count = 0 %}
+            {% for page in blog.pages %}
+            {% if page.extra.hide_from_blog %}{% continue %}{% endif %}
```
HarHarLinks, line 111: have you considered also excluding TWIMs given they have their own section and don't give a huge benefit here?

templates/index.html
```diff
@@ -8,50 +8,135 @@
...
-            <a href="/docs/communities" class="call-to-action inverted">Reclaim your Space</a>
+
+    <div id="use-cases" class="section">
+        <h2>Built for everyone</h2>
+        <div class="use-cases-grid">
+            <div class="use-case-card">
+                <img src="assets/frontpage/personal-graphic.svg" alt="" />
+                <h3>Individuals &amp; families</h3>
+                <p>Chat securely with friends, family, and co-workers. Your conversations, on your terms, with full end-to-end encryption.</p>
+                <a href="/try-matrix" class="call-to-action">Start a conversation</a>
```
HarHarLinks, line 61: ctas in the all of new sections are missing the center alignment in mobile views

templates/index.html
```diff
@@ -8,50 +8,135 @@
...
+                    {% if page.taxonomies.category %}
+                    <span class="blog-preview-tag">{{ page.taxonomies.category | first }}</span>
+                    {% endif %}
+                </div>
+                <h3><a href="{{ page.permalink }}">{{ page.title }}</a></h3>
+                {% if page.description %}
+                <p>{{ page.description }}</p>
+                {% endif %}
+                <a href="{{ page.permalink }}" class="blog-preview-readmore">Read more →</a>
+            </div>
```
HarHarLinks, line 126: I would keep the text as is, but make the whole tile clickable, like the single TWIM card above. Or drop the hover animation on the card, which seems inconsistent.

sass/_index.scss
```diff
@@ -102,186 +122,395 @@
...
+            display: flex;
+            flex-direction: column;
+            gap: 16px;
+            text-align: left;
+            background-color: #fff;
+            transition: border-color 0.2s ease;
+
+            &:hover {
+                border-color: #888;
+            }
```
HarHarLinks, line 472: not needed here.

sass/_index.scss
```diff
@@ -102,186 +122,395 @@
...
 
-        .use-case-personal-content {
-            display: flex;
-            flex-direction: row;
-            justify-content: space-between;
-            max-width: 1000px;
-            margin-inline: auto;
+            &:hover {
+                border-color: #666;
+            }
```
HarHarLinks, line 163: not needed

templates/index.html
```diff
@@ -8,50 +8,135 @@
...
+                <h3>Own Your Conversations</h3>
+                <p>Host your own server or choose one you trust. No single corporation owns Matrix—it's a global open network where you control your data and who can access it.</p>
+            </div>
+            <div class="feature-item">
+                <h3>One Network, Many Apps</h3>
+                <p>Choose from dozens of compatible apps and servers built by the community. Matrix bridges to other platforms so you can reach everyone, everywhere.</p>
+            </div>
+            <div class="feature-item">
+                <h3>Community Powered</h3>
+                <p>Built by its community, for its community. An open, transparent project governed by the Matrix.org Foundation—not a corporate walled garden.</p>
```
HarHarLinks, line 49:
```suggestion
                <p>Built by its community, for its community. An open, transparent project governed by The Matrix.org Foundation—not a corporate walled garden.</p>
```
this also wraps somewhat unfortunately

MTRNord:
Converting back to draft after talking internally due to ongoing discussions first :) This will be put on the waitinglist to at least end of april or later.

PR #3299, opened 2026-03-24
Fix Issues found by zizmor

MTRNord:
### Description

This fixes issues found by the [zizmor linter](https://zizmor.sh/).

This is a tool I discovered via https://github.com/curl/curl/pull/21085

While we are luckily not affected this time this for example will bring more safety against things like https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23

### Related issues

None

### Role

Website & Content WG

### Timeline

Whenever the moon is green, blue, red or a shade of pink. If neither applies use any moonphase when Saturn did half a rotation /s

### Signoff

See commits

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,5 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
+    labels:
+      - dependencies


MTRNord:
> with my staff hat: please also add the action (https://github.com/zizmorcore/zizmor-action) as a workflow.

Oh didnt even see that. That seems like a good idea :)

.github/dependabot.yml
```diff
@@ -6,6 +6,8 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions" # See documentation for possible values
-    directory: "/.github" # Location of package manifests
+    directory: "/.github/workflows" # Location of package manifests
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
```
HarHarLinks, line 13:
```suggestion
      default-days: 7
    labels:
      - meta
```

.github/workflows/zizmor-lint.yaml
```diff
@@ -0,0 +1,25 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+    push:
+        branches: ["main"]
+    pull_request:
+        branches: ["**"]
```
HarHarLinks, line 7: i see this was copied from https://github.com/matrix-org/policyserv-setup-bot/pull/20/files, but what's the point?

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

leaving out this line should already make it run on all branches, e.g. .github/workflows/linter.yaml#L6

MTRNord: Actually this was copied from https://github.com/zizmorcore/zizmor-action?tab=readme-ov-file#usage-with-github-advanced-security-recommended

and I am in fact not quite sure what the actual difference is. I assumed that people writing a linter for this actually know what they are doing.

The commit adding it to their readme sadly is not very descriptive https://github.com/zizmorcore/zizmor-action/commit/665dc7917f47f3251e94749574feaabd466f6325

My best guess here is that this is just meant to be explicit rather than having implicit behaviour here.

.github/dependabot.yml
MTRNord: Wouldnt we want to still have "dependencies" also as a label?

```suggestion
      default-days: 7
    labels:
      - meta
      - dependencies
```

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#labels-- says the default behaviour is overridden if setting it.

HarHarLinks: i didn't realise that, again https://github.com/matrix-org/policyserv-setup-bot/pull/20/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28 sets it explicitly. I'm not opposed to it, but I didn't expect it since I think we haven't used it before but the meta label instead. If we're keeping it I would separate meta and dependencies, actually, so this would only use the dependencies label (maybe explicitly?)
```suggestion
      default-days: 7
    labels:
      - dependencies
```

.github/workflows/zizmor-lint.yaml
HarHarLinks: consistency as an argument works for me, both with the template and the rest of the org, but I was curious.

.github/dependabot.yml
MTRNord: I guess that works for me. Applying.

PR #3331, opened 2026-04-03
Rename Synapse Admin (etke.cc fork) to Ketesa

aine-etke:
### Description

Rename Synapse Admin (etke.cc fork) to Ketesa.
Following up https://github.com/etkecc/ketesa/releases/tag/v1.0.0

### Role

Project maintainer

### Timeline

No hard deadline

### Signoff

Signed-off-by: Aine <97398200+aine-etke@users.noreply.github.com>

### DIFF
--- a/content/ecosystem/integrations/integrations.toml
+++ b/content/ecosystem/integrations/integrations.toml
@@ -142,18 +142,18 @@ licence = "MIT"
 repository = "https://gitlab.com/jpypi/rustix"
 
 [[integrations]]
-name = "Synapse Admin"
+name = "Ketesa"
 description = """
-Admin console for synapse Matrix homeserver with additional features
+Admin UI for Matrix servers, formerly Synapse Admin.
+Drop-in replacement with extended features, multi-backend support, and visual customization.
 
-(Maintained by etke.cc fork)
 CDN version: https://admin.etke.cc
 """
 maintainer = "etke.cc"
 language = "TypeScript"
 licence = "Apache-2.0"
-repository = "https://github.com/etkecc/synapse-admin"
-room = "#synapse-admin:etke.cc"
+repository = "https://github.com/etkecc/ketesa"
+room = "#ketesa:etke.cc"
 
 [[integrations]]
 name = "matrix-hookshot"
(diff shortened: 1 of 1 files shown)


content/ecosystem/integrations/integrations.toml
```diff
@@ -142,18 +142,18 @@ licence = "MIT"
 repository = "https://gitlab.com/jpypi/rustix"
 
 [[integrations]]
-name = "Synapse Admin"
+name = "Ketesa"
 description = """
-Admin console for synapse Matrix homeserver with additional features
+Admin UI for Matrix servers, formerly Synapse Admin.
```
HarHarLinks: I feel odd about generalising this statement to "Matrix servers" when apparently still requires synapse- (and MAS?) specific admin APIs.

Saying this means we recommend to any admin to install this, so a continuwuity admin might and then be confused that it's not working.

aine-etke: from the discussion in the TWIM room: this change was intentional, but we don't want to share more details _just yet_.

copy of my TWIM message:

the reason is simple: for the last few weeks we worked hard as hell to sort out everything (including Ketesa), and the team just doesn't have  the will to live enough gas in the tank right now to continue pushing with the same speed. We're tired as hell, and making a proper announcement right now means people will be interested and will _wait_ for that to happen.

So, instead we will take some time to rest, sort out other stuff (the todo pile is still growing 😭 ), and then, refreshed (hopefully), make an announcement with the actual start of implementation so people can play with nightly builds while it's in progress.

With the v1.0.0 release we heavily refactored the project overall, and the APIs (providers) part specifically as a groundwork for the _future improvements_, but that's it _for now_

---

If you insist on changing the wording, sure!

thibaultamartin: The description should reflect the current reality. The best solution to manage user expectations and not put too much pressure on the etke.cc team would be to reword like Kim suggested, and update the wording later once Synapse is no longer needed :)

HarHarLinks: (I realise that hasn't always been handled well in the Matrix ecosystem but hence my effort to improve the clarity in such situations when I can - thanks for understanding)

aine-etke: Sure, updated the description using Ketesa's readme

MTRNord (approved):

Issue #3230, opened 2026-02-22
Clarify ecosystem features


HarHarLinks:
#3197 overhauled the client feature table of the ecosystem section of the website. [This](https://github.com/matrix-org/matrix.org/pull/3197#discussion_r2758888830) and other PRs trying to maintain client feature flags keep reminding us that these features are not well explained.

We should try to find clearer definitions of each feature and document the in places such as [templates/macros/clients.html](https://github.com/matrix-org/matrix.org/pull/3197/files/97732f460de191a03e3bd75d208f6941db2977a8#diff-63e40598efc5bf3e9cfc2c820ad3e4493473e1e1dc7ff7fc937d331a2ad8e6d3), CONTENT.md and where ever else relevant (maybe deduplicate and referencing just a single place for sanity?).

We should also add a feature to the ecosystem pages to learn more about the definition of a feature. I think something like a question mark icon for each feature entry which when hovering or clicking will display the explanation text would work.

awtj8o81ryywg793:
Another point to clarify is what is meant by "support" in regard to a particular feature. Are we implying that the most recent formal release of a project includes support for such a feature, or that code sitting in a remote repository, without a formal release (yet), enables support for a certain feature? I guess _both_, but I'm not sure how to communicate that.

HarHarLinks:
tbd: what does "supported" generally mean?

a frequent argument is about features being hidden behind feature flags, i.e. it is confusing to users since they're not enabled by default.
on the other hand, features might be technically available, so they're technically supported.

Imo we should assume defaults but widen the capabilities of our table to indicate features available behind a flag. We should have symbols for common states outside yes/no. This should be implemented in a way such that determining the detailed state is easy to a viewer.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/3230#issuecomment-4150712393):

> Imo we should assume defaults but widen the capabilities of our table to indicate features available behind a flag. We should have symbols for common states outside yes/no. This should be implemented in a way such that determining the detailed state is easy to a viewer.

I agree with this, given that the page that lists clients seems to be targeted to end-users rather than developers (or other types of users who may be more interested in what a client _technically_ supports), and that is probably the assumption most viewers have. However, with your idea of iterating on the table to display the availability of support, I think all types of viewers will be happy.

In terms of what "support" means, I guess you have:
- Support, by default - feature is supported by default, without the need of enabling any configuration setting(s).
[...]

Issue #3068, opened 2025-12-12
Update: Change the maturity status of Chooj from "Alpha" to "Obsolete"


awtj8o81ryywg793:
Chooj hasn't been updated in a while and has also officially been abandoned as of August 3, 2025. As a result, the maturity status of the client should be changed to "Obsolete".

@osousa, thank you for the [heads-up](https://github.com/matrix-org/matrix.org/pull/3054#issuecomment-3641744473)!

TheArcaneBrony:
Why are you opening issues for things that  you PR immediately anyways? seems kinda pointless to allocate 2 issue numbers?

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/3068#issuecomment-3644672084):

> Why are you opening issues for things that you PR immediately anyways? seems kinda pointless to allocate 2 issue numbers?

See [`CONTRIBUTING.md`](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md#how-we-take-decisions).

I _may_ be misunderstanding the "How we take decisions" section, as it might be aimed at _just_ the content that makes up blog posts, support articles, etc. However, this repository hosts more than just that content, and there is no other guidance in that document for changes such as those made in #3069. For the record, I don't necessarily disagree with you that it can be seen as a bit pointless.

TheArcaneBrony:
Honestly all of the pull requests I've made have been without open issues - I never noticed that particular line in the contributing docs, but at the same time the line in question also claims that said issue must have been accepted first.

I feel like it's probably fine as long as you document why you're making the change? Especially relevant is that *all pull requests* by themselves are issues due to how the entire system works.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/3068#issuecomment-3644785732):

> Honestly all of the pull requests I've made have been without open issues - I never noticed that particular line in the contributing docs, but at the same time the line in question also claims that said issue must have been accepted first.

[...]

PR #3351, opened 2026-04-13
issue template: invite people to PR if they're up for it

raboof:
### Description

Highlight people comfortable with GitHub can also just PR.

Not sure how to test if this renders correctly tbh... maybe let's just try it.

### Related issues

Discussed in https://github.com/matrix-org/matrix.org/issues/3329#issuecomment-4183416566

### Role

Independent

### DIFF
--- a/.github/ISSUE_TEMPLATE/update-ecosystem-info.yaml
+++ b/.github/ISSUE_TEMPLATE/update-ecosystem-info.yaml
@@ -46,4 +46,4 @@ body:
     attributes:
       value: |
         Note that if you're comfortable with using GitHub, you're also welcome to skip
-        creating an issue go straight to creating a PR to `./content/ecosystem`.
+        creating an issue and go straight to creating a PR to `./content/ecosystem`.
--- a/.github/ISSUE_TEMPLATE/update-ecosystem-info.yaml
+++ b/.github/ISSUE_TEMPLATE/update-ecosystem-info.yaml
@@ -42,3 +42,8 @@ body:
       description: Any other details that would help us update this entry?
     validations:
       required: false
+  - type: markdown
+    attributes:
+      value: |
+        Note that if you're comfortable with using GitHub, you're also welcome to skip
+        creating an issue go straight to creating a PR to `./content/ecosystem`.


.github/ISSUE_TEMPLATE/update-ecosystem-info.yaml
```diff
@@ -42,3 +42,8 @@ body:
       description: Any other details that would help us update this entry?
     validations:
       required: false
+  - type: markdown
+    attributes:
+      value: |
+        Note that if you're comfortable with using GitHub, you're also welcome to skip
+        creating an issue go straight to creating a PR to `./content/ecosystem`.
```
HarHarLinks: I think
```suggestion
        creating an issue and go straight to creating a PR to `./content/ecosystem`.
```

HarHarLinks (commented):
technically our CONTENT.md speaks of requiring an accepted issue before implementation, however my tendency is to allow some leeway to ecosystem updates. the only contentious bit i see there is our unclear definition of what it means to support some of the named features, which however we should resolve anyway.

raboof:
> technically our CONTENT.md speaks of requiring an accepted issue before implementation

Actually it's https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md#how-we-take-decisions :)

I personally dislike such a rule, and prefer the "if you PR without creating an issue you risk us rejecting your PR, that's your risk" approach - but that's up to you of course.

> the only contentious bit i see there is our unclear definition of what it means to support some of the named features, which however we should resolve anyway.

Yeah fair. Indeed ElementX does know about threads somewhat, but it's so basic I didn't feel comfortable suggesting that would be 'supported'. Maybe it would be good to have a 'somewhat (see comment)' option :D

HarHarLinks:
The respective issue is https://github.com/matrix-org/matrix.org/issues/3230

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3351#issuecomment-4236852469):

> I personally dislike such a rule, and prefer the "if you PR without creating an issue you risk us rejecting your PR, that's your risk" approach - but that's up to you of course.

I'm not against adopting such a change (or something similar) for pull requests that involve minor changes; it would help prevent situations like [this](https://github.com/matrix-org/matrix.org/issues/3068) from occurring. I also support clarifying when an issue _is_ needed (e.g., for significant changes).

PR #3354, opened 2026-04-16
Add basic pagefind setup

### LINKED ISSUE #19: Implement search
From https://matrix.org/jira/browse/ORG-57

[we have a search box but it only searches the wordpress blog afaict]

MTRNord:
### Description

This adds a basic pagefind setup with the default search box and some extra labeling in case we decide we want some more advanced search dialog later.

It also does not do any styling for now. I think the defaults are sufficient for matrix.org

cc @thibaultamartin This likely needs SRE to update cloudflare for it to work. I am not sure who to ping and whats needed. Iirc you already run something like this on cloudflare so maybe you can help here?

To build pagefind we need `npx -y pagefind --site public` added to cloudflare

---

#### Visual changes

**Search Bar in the Navigation**

grafik

**Search Modal Empty**

grafik

**Search bar when used**

grafik

### Related issues

Fixes #19 

### Role

Website & Content WG

### Timeline

N/A

### Signoff

See commits

### DIFF
--- a/pagefind.yml
+++ b/pagefind.yml
@@ -1,0 +1,10 @@
+site: public
+output_subdir: pagefind
+exclude_selectors:
+  # Ignore codeblocks
+  - "code"
+  # Blocked due to https://github.com/Pagefind/pagefind/issues/1134
+  # - "#that-s-all-i-know"
+  # - "#that-s-all-i-know-checkered-flag"
+  # - "#that-s-all-i-know+p"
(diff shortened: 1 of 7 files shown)


MTRNord:
Possibly worth deciding is if we want to exclude the author and category pages fully or just their post lists from the index to prevent duplicate post results.

MTRNord:
Notes from the meeting:

- Make it maybe more of a button in the navbar
- Filter out partial posts in categories, author pages and other places

PR #3374, opened 2026-04-26
Add matrirc to ecosystem/clients

pawelb0:
### Description

Adds matrirc to the clients ecosystem listing. matrirc is a local IRC
server that bridges to Matrix with native E2EE, so users can connect
any IRC client (irssi, weechat, hexchat) to their Matrix account.

Updated to match the latest spec.

### Related issues

None.

### Role

Independent. Author of matrirc.

### Timeline

No deadline.

### Signoff

On the commit.

### DIFF
--- a/content/ecosystem/clients/matrirc.md
+++ b/content/ecosystem/clients/matrirc.md
@@ -1,0 +1,29 @@
++++
+title = "matrirc"
+
+[extra]
+maintainer = "pawelb0"
+maturity = "Beta"
+repo = "https://github.com/pawelb0/matrirc"
+matrix_room = ""
+licence = "GPL-3.0-or-later"
+latest_release = "2026-04-25"
+featured = false
+good_for = "IRC users who want to use Matrix from their existing IRC client (irssi, weechat, hexchat)"
+
+[extra.features]
+e2ee = true
+spaces = false
+voip_1to1 = false
+voip_jitsi = false
+threads = false
+sso = false
+oauth = false
+multi_account = false
+multi_language = false
+
+[extra.packages]
+homebrew_formula = "matrirc"
++++
+
+A local IRC server that bridges to Matrix with native E2EE. Run matrirc on localhost; connect any IRC client (irssi, weechat, hexchat) and chat in your Matrix rooms and DMs.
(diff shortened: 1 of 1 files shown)


weeman1337 (changes requested):
Unfortunately the client info format changed. Can you please update it the [to current spec](https://github.com/matrix-org/matrix.org/blob/main/CONTENT.md#clients)?

awtj8o81ryywg793 (commented):
Given how this pull request’s description reads, I don’t think it should be listed as a client, unless I'm missing something. Perhaps it belongs either as a bridge or within the "other" category (see: https://github.com/matrix-org/matrix.org/pull/3602)?

pawelb0:
@weeman1337 Updated to the current client spec

@awtj8o81ryywg793 I think clients is the right category: matrirc is a user-run program you use to chat on Matrix with your own account — IRC is just the UI layer, similar to how Ement.el uses Emacs as its UI. It isn't a server-side bridge connecting a Matrix room to an external network, and it doesn't fit the per-network `bridges.toml` structure (there is no third-party platform account or privilege model involved). That said, if the maintainers prefer, I'm happy to move it to "other" once #3602 lands.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/pull/3374#issuecomment-5323744553):

> I think clients is the right category: matrirc is a user-run program you use to chat on Matrix with your own account — IRC is just the UI layer, similar to how Ement.el uses Emacs as its UI. It isn't a server-side bridge connecting a Matrix room to an external network, and it doesn't fit the per-network bridges.toml structure (there is no third-party platform account or privilege model involved). That said, if the maintainers prefer, I'm happy to move it to "other" once https://github.com/matrix-org/matrix.org/pull/3602 lands.

You’re right about Matrirc not being a bridge, at least when using the current suite of bridges listed on matrix.org as a benchmark. I should probably have rephrased that comment. Fairly regularly, we seem to circle back around to the "what is a client?" question. I think listing Matrirc under the "miscellaneous" category (see: https://github.com/matrix-org/matrix.org/pull/3602) would be best, given that the project self-describes as a "Local IRC server backed by Matrix. " Then, once we’ve sorted out what we mean by "client" (and other categories), we can circle back around and review MatriIrc again if needed.

PR #3395, opened 2026-05-04
Publish TWIM 2026-05-04

thibaultamartin:
### Description

Publish TWIM 2026-05-04. It misses ping section

### Role

🎩 Website WG

### Timeline

Now :)

### Signoff

Please [sign off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md) your individual commits.

✅

PR #3396, opened 2026-05-04
Add a small helper to auto generate internal links from matrix.org links

MTRNord:
### Description

Helper which should correctly convert matrix.org absolute urls to the internal zola format on a given file if you run it from the root of the matrix.org repo.

Not a python person so this may be wrong in places? Feel free to nit?

### Role

Website & Content WG

### Timeline

N/A

### Signoff

See commits

### DIFF
--- a/contrib/convert_matrix_org_links.py
+++ b/contrib/convert_matrix_org_links.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2026 The Matrix.org Foundation C.I.C.
+
+"""
+Convert https://matrix.org links to Zola internal link format (@/).
+
+This script converts absolute matrix.org links to relative Zola links using the
+@/ prefix, which references content from the content/ directory. It intelligently
+resolves file paths by checking the actual directory structure.
+
+Usage:
+    python contrib/convert_matrix_org_links.py <file_path>
+    python contrib/convert_matrix_org_links.py content/blog/2026/05/2026-05-04-twim.md
+"""
+
+import re
+import sys
+from pathlib import Path
+
+
+def find_file_in_content(path_segment: str, content_root: Path) -> str | None:
+    """
+    Find the actual file path in content/ directory for a given path segment.
+
+    Args:
+        path_segment: Path segment from URL (e.g., "blog/2026/05/conf-ticket-sales")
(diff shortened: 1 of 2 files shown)


contrib/convert_matrix_org_links.py
```diff
@@ -0,0 +1,199 @@
...
+# SPDX-FileCopyrightText: 2026 The Matrix.org Foundation C.I.C.
+
+"""
+Convert https://matrix.org links to Zola internal link format (@/).
+
+This script converts absolute matrix.org links to relative Zola links using the
+@/ prefix, which references content from the content/ directory. It intelligently
+resolves file paths by checking the actual directory structure.
+
+Usage:
```
HarHarLinks, line 12: what's the `content_root` here?

contrib/convert_matrix_org_links.py
```diff
@@ -0,0 +1,199 @@
...
+    python contrib/convert_matrix_org_links.py <file_path>
+    python contrib/convert_matrix_org_links.py content/blog/2026/05/2026-05-04-twim.md
+"""
+
+import re
+import sys
+from pathlib import Path
+
+
+def find_file_in_content(path_segment: str, content_root: Path) -> str | None:
```
HarHarLinks, line 22: this failed to find the right election centre in https://github.com/matrix-org/matrix.org/pull/3395/commits/ee69d53d0e97c56f63b8d4e1c39c49786c6f8af5

Was https://github.com/matrix-org/matrix.org/blob/ee69d53d0e97c56f63b8d4e1c39c49786c6f8af5/content/blog/2026/05/2026-05-04-twim.md?plain=1#L62 deliberate or also some bug?

HarHarLinks (changes requested):

contrib/convert_matrix_org_links.py
MTRNord: No that was regex accident before this script was done. This script wasnt in full used on the last twim.

contrib/convert_matrix_org_links.py
MTRNord: Its mostly a convenience feature for when I was testing which allowed me to have some test file in /tmp and then point at the correct content folder of matrix.org. Will update the docs to reflect that :)

weeman1337 (changes requested):
I wold suggest to include a call of `zola check` to the script after replacing the URLs. Otherwise, just go for it?

PR #3432, opened 2026-05-25
add Max to Room Directory WG members, list LogN as vice-chair

thetayloredman:
### Description

This PR adds Max (TheDarkWizard) to the members list of the Room Directory Working Group along with listing LogN as a vice chair.

### Related issues

N/A

### Role

Vice Chair of the Room Directory Working Group

### Timeline

No specific deadline.

### Signoff

Signed-off-by: Logan Devine <logan@zirco.dev>

CC: @ll-SKY-ll (WG chair)

### DIFF
--- a/content/foundation/working-groups/working_groups.toml
+++ b/content/foundation/working-groups/working_groups.toml
@@ -201,8 +201,8 @@ The membership of the Room Directory Working Group will be determined by the [Tr
 Members may be added or removed by a majority vote of the T&S Committee.
 """
 committee = "Trust & Safety Committee"
-members = ["Nico", "Sky", "mahdi", "Emma [it/its]", "Krille - Christian K.", "LogN"]
-chairs = ["Sky"]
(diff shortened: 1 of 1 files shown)


ll-SKY-ll:
interestingly the chairs arent even shown on the website :thinking:  
also, should we include both names? looking at other working groups it looks like people either have their matrix username or their real name. Or Krille who has a combination of both in his matrix user nickname :P 
edit: the change itself is correct though, LogN is vice chair as of yesterday and TheDarkwizard joined recently too :)

HarHarLinks:
> interestingly the chairs arent even shown on the website 🤔 also, should we include both names? looking at other working groups it looks like people either have their matrix username or their real name. Or Krille who has a combination of both in his matrix user nickname :P

We don't have strict rules, people can choose under which name they want to appear and also when to redact their membership from public display. I've considered unifying stuff before, but on my own example it wasn't clear to me how to handle it best. 🤷
We do have slightly related a ticket somewhere about blog author profiles which could use the same system if you have ideas about that.

HarHarLinks (approved):

PR #3446, opened 2026-06-05
Fix botched tags in the skel

thibaultamartin:
### Description

The opengraph tags were botched, because Zola tries to escape the URLs, which is the opposite of what we need it to do.

### Role

🫡 Website & Content WG

### Timeline

As soon as possible since it botches previews.

### Signoff

✅

### DIFF
--- a/templates/skel.html
+++ b/templates/skel.html
@@ -14,7 +14,7 @@
     <meta name="twitter:creator" content="@matrixdotorg" />
     <meta property="og:site_name" content="matrix.org" />
     {% if current_url %}
-    <meta property="og:url" content="{{ current_url }}" />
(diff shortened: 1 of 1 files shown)


config.toml
```diff
@@ -20,11 +20,8 @@ taxonomies = [
     { name = "category", feed = true, paginate_by = 10 },
 ]
 
-[markdown]
-# Whether to do syntax highlighting
-# Theme can be customised by setting the `highlight_theme` variable to a theme supported by Zola
-highlight_code = true
-highlight_theme = "visual-studio-dark"
+[markdown.highlighting]
+theme = "github-dark"
```
MTRNord, line 28: Hm whats this about ? 🤔

MTRNord (approved):

Issue #3422, opened 2026-05-18
Add sticker packs to list of client features


HarHarLinks:
https://github.com/matrix-org/matrix.org/blob/f758eca5b5e74423fe9edb5cf7da96cfe08a6b03/CONTENT.md?plain=1#L164-L173

MSC2545 is merged, and sticker packs is a feature of high interest to the community (as evidenced by all major community clients).

Unfortunately there are multiple follow-up MSCs implemented to different degree. I'm also not sure which implementations have stable support.

Related: #3230

Gernomaly:
I second this issue. We should try to find a consens on how to do this

PR #3449, opened 2026-06-06
More client features and feature levels

### LINKED ISSUE #3229: Update clients feature flags with new OAuth 2.0 API support
#3186 introduced a new feature flag to the clients list.

All clients need to be checked and set the feature to true where supported.

PRs implementing this would ideally point out the evidence for each client supporting it.

### LINKED ISSUE #3418: Cinny supports MatrixRTC
### Project Affected

https://github.com/cinnyapp/cinny

### Current Information

It claims no support for VOIP

### Suggested Correction

MatrixRTC has been included on March 7, by adding a bundled element-call

### Source

https://github.com/cinnyapp/cinny/pull/2680

### Additional Context

Ideally, the ecosystem would be able to distinguish between legagy calls and MatrixRTC calls

### LINKED ISSUE #3504: EPIC: improve client feature list
This epic collects issues related to improving the state of the client feature list on /ecosystem/clients/<client> pages.

Image

- tooltip on a feature could somehow allow you to find other clients who support that feature
- ordering of features
- display consistent list of features even when some clients are missing entries

HarHarLinks:
### Description

- restructures the client features to separate stable spec from upcoming spec from unrelated to spec
- adds additional feature levels "unknown" and "partial" to previous boolean

### Related issues

- resolves https://github.com/matrix-org/matrix.org/issues/3418
- resolves https://github.com/matrix-org/matrix.org/issues/3229

Follow-ups:
- https://github.com/matrix-org/matrix.org/issues/3422
- https://github.com/matrix-org/matrix.org/issues/3499
- https://github.com/matrix-org/matrix.org/issues/3503
- https://github.com/matrix-org/matrix.org/issues/3230
- tooltip on a feature could somehow allow you to find other clients who support that feature
- ordering of features
- legend and/or a11y of the icons #3462
- https://github.com/matrix-org/matrix.org/issues/3458

### Role

Website & Content WG

### Signoff


HarHarLinks:
Reference for icons license: https://fontawesome.com/license/free

content/ecosystem/clients/chatrix.md
```diff
@@ -11,15 +12,25 @@ repo = "https://github.com/Automattic/chatrix"
...
+
+[extra.features.1spec]
+e2ee = "unsupported"
+spaces = "unsupported"
+voip_1to1 = "unsupported" 
+threads = "unsupported"
+sso = "unsupported"
+
+[extra.features.2matrix2]
+oauth = "unknown"
```
HarHarLinks: this is actually in spec so should be under spec features or we should rename the section to Matrix 2.0 or so

static/assets/hourglass_partial.svg
HarHarLinks, line 1: the icons should ideally be more obvious by color or something. maybe this doesn't need to be a blocker?

HarHarLinks, line 1: the vertical alignment with the text isn't quite as good as for the other icons 
image

thibaultamartin: Doesn't have to be a blocker for me. I'm somewhat wary of adding too many colors, I wouldn't want it to be too distracting.

thibaultamartin: It doesn't strike me as too bad, I would definitely not block on it.

templates/ecosystem/client.html
```diff
@@ -45,20 +45,26 @@ <h3>{{ page.title }}<div class="maturity {{ page.extra.maturity | lower }}">{{ p
...
         <div class="features-matrix">
-            <ul>
-                {% for feature, supported in page.extra.features %}
-                <li {% if supported %}class="supported" {% endif %}>
-                    {{ clients::feature_label(feature_name=feature) | trim }}
-                </li>
-                {% endfor %}
-            </ul>
-            <p>Something wrong? Report the issue <a href="https://github.com/matrix-org/matrix.org/issues/new?template=update-ecosystem-info.yaml">here</a>.</p>
+            {% for category, my_features in page.extra.features %}
```
thibaultamartin: Nit, but `my_features` reads a bit odd here. Wouldn't `features` work?

content/ecosystem/clients/element-x.md
```diff
@@ -14,16 +14,23 @@ featured_order = 2
...
+spaces = "supported"
+voip_1to1 = "unsupported"
+threads = "partial"
+sso = "unsupported"
+
+[extra.features.2matrix2]
+oauth = "supported"
+voip_matrixrtc = "supported"
+sliding_sync = "supported"
+invisible_crypto = "supported"
```
HarHarLinks: actually idk if you could call that supported or rather partially supported. i'm not sure how far along the MSC actually is to call anything more than `"partial"` even in the MSC's reference implementation.

templates/ecosystem/client.html
```diff
@@ -45,20 +45,26 @@ <h3>{{ page.title }}<div class="maturity {{ page.extra.maturity | lower }}">{{ p
...
+                        </li>
+                    {% endfor %}
+                </ul>
+            </div>
+            {% endfor %}
         </div>
-    </div>
-    <div class="stores">
-        {{ clients::install_links(packages=page.extra.packages) }}
+        <p>Something wrong? Report the issue <a href="https://github.com/matrix-org/matrix.org/issues/new?template=update-ecosystem-info.yaml">here</a>.</p>
```
weeman1337: It is always easier to have a larger click area


```suggestion
        <p>Something wrong? <a href="https://github.com/matrix-org/matrix.org/issues/new?template=update-ecosystem-info.yaml">Report the issue here</a>.</p>
```

weeman1337 (commented):
I think there should be a legend or `title`/`aria-label` attribute for the feature status, so that a user can find out what it means. Also regarding a11y.

Image

HarHarLinks:
> I think there should be a legend or `title`/`aria-label` attribute for the feature status, so that a user can find out what it means. Also regarding a11y.

@weeman1337 yeah i had similar thoughts, even for seeing users this might needs some kind of legend. this means we need to refactor away from the custom list icon marker solution, doesn't it?

hate to be that guy but do we need to block this PR on that if it's not worsening the status quo?

templates/macros/clients.html
```diff
@@ -1,3 +1,16 @@
+{# Human-readable labels for feature categories #}
+{% macro category_label(category_name) %}
+{% if category_name == "1spec" %}
+Matrix Standard
+{% elif category_name == "2matrix2" %}
+Future Matrix Standard
```
Gernomaly: I am not sure if this is a good title for that, because it implies those feature to be 100% part of the standard in the future.

Anyone got other ideas how that section could be called?

Gernomaly: Idk if this is a good title for that section. I dont like the implication of finality about that features getting added. 

Anyone got a better idea on how to call this section?

HarHarLinks:
Office hour discussion:
image

static/assets/hourglass_partial.svg
HarHarLinks, line 1: alternatively could use an Erlenmeyer flask as symbol for "experiemental"/"brewing" support

Gernomaly: if you dont know it, its fine. But once you notice its for me impossible to unsee. I wouldnt block it either, but if someone got an idea on how to solve this i wouldnt mind

Gernomaly: i like that idea, id have to see it in practice to give a informed opinion

Gernomaly:
> > I think there should be a legend or `title`/`aria-label` attribute for the feature status, so that a user can find out what it means. Also regarding a11y.
> 
> @weeman1337 yeah i had similar thoughts, even for seeing users this might needs some kind of legend. this means we need to refactor away from the custom list icon marker solution, doesn't it?
> 
> hate to be that guy but do we need to block this PR on that if it's not worsening the status quo?

It feels very wrong to me to not block this for accessibility reasons. Id rather delay this pr for a bit to do it properly, than merge it and it doesnt get fixed at all in the end.

content/ecosystem/clients/element-x.md
HarHarLinks: for the purpose of Matrix 2.0, "invisible crypto" is just https://github.com/matrix-org/matrix-spec-proposals/pull/4153, but there remain also multiple related MSCs open

content/ecosystem/clients/chatrix.md
HarHarLinks, line 1: todo update CONTENT.md

weeman1337 (approved):

weeman1337:
> > > I think there should be a legend or `title`/`aria-label` attribute for the feature status, so that a user can find out what it means. Also regarding a11y.
> > 
> > 
> > @weeman1337 yeah i had similar thoughts, even for seeing users this might needs some kind of legend. this means we need to refactor away from the custom list icon marker solution, doesn't it?
> > hate to be that guy but do we need to block this PR on that if it's not worsening the status quo?
> 
> It feels very wrong to me to not block this for accessibility reasons. Id rather delay this pr for a bit to do it properly, than merge it and it doesnt get fixed at all in the end.

Noted. I will directly create a follow up issue.

weeman1337:
Follow up https://github.com/matrix-org/matrix.org/issues/3462

PR #3455, opened 2026-06-08
create iframes for embedding youtube videos

gian2dchris:
### Description

create iframe elements for embedding youtube videos

### Signoff

### DIFF
--- a/static/js/components.js
+++ b/static/js/components.js
@@ -133,14 +133,20 @@ class YoutubePlayer extends HTMLElement {
     loadPlayer() {
         const videoId = this.getAttribute("video-id");
         const start = this.getAttribute("start");
+
+        const iframe = document.createElement("iframe");
         let embedUrl = `https://www.youtube.com/embed/${videoId}`;
         if (start) embedUrl += `?start=${start}`;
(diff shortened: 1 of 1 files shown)


static/js/components.js
```diff
@@ -133,14 +133,20 @@ class YoutubePlayer extends HTMLElement {
     loadPlayer() {
         const videoId = this.getAttribute("video-id");
         const start = this.getAttribute("start");
+
+        const iframe = document.createElement("iframe");
```
weeman1337, line 137: Isn't that effectively the same than before?

Can you please describe why you have introduced these changes?

weeman1337: I can only guess. I think it is good to avoid setting `innerHTML` to prevent injections through `videoId`.

weeman1337 (approved):

static/js/components.js
gian2dchris: right, unlikely to happen, yet worth the simple fix.
thanks

HarHarLinks:
Hi @gian2dchris, in the future it would help us maintainers if the PR description stated the intention more clearly as well as your affiliation to provide context for this PR proactively. If you have concerns about that in cases where security is concerned you can also contact us privately, see our contributing guidelines.

Issue #3462, opened 2026-06-14
a11y and understandability for feature list on client details page


weeman1337:
Currently there are icons for the status of a feature on a client detail page

<img width="446" height="494" alt="Image" src="https://github.com/user-attachments/assets/f9daddd3-dc80-4872-ba87-3b481fc7a43d" />

- ✅  if supported
- ❌ if unsupported
- ❔ if unknown
- 🧪 if experimental

There are some issues with that:

- For persons that use assistive technologies (such as screen readers) there are no accessible texts for the items
- For other persons it may be unclear what the icons mean

Suggestions:

- Add a legend below the feature list that explains the icons
- Add `aria-label` to the items (ensure that child elements are aria hidden then)

HarHarLinks:
there is also 🧪 if partially supported

awtj8o81ryywg793:
I'm not going to say no to improving accessibility and helping more people understand what they're looking at.

Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/3462#issue-4658969090):

> Add a legend below the feature list that explains the icons

Instead of below the list, how about placing a horizontal legend just below the "Features" heading?

PR #3465, opened 2026-06-14
Add client feature descriptions

HarHarLinks:
### Description

Adds a (?) icon to the client feature list. Upon hover, an explanation for the respective feature is shown.

~~Unclear: a11y, mobile support~~

### Related issues

https://github.com/matrix-org/matrix.org/pull/3449

### Role

Website & Content WG

### Signoff


sass/_clients.scss
```diff
@@ -121,6 +121,12 @@ body:has(.projects-card-deck > div > .client-checkbox:checked) {
                 text-align: center;
             }
         }
+
+        img {
+            height: 1em;
```
weeman1337, line 126: In my opinion the icon is a bit too small. Also at least I would prefer them to be aligned.

andypiper:
lgtm 👍🏻

MTRNord (approved):
Looks good to me.

I dont think any of this is worth fixing but I noticed:

- Surprised about the lab icon on the Threads on Element X (but also didnt check. Maybe schildi next was further. I only used that lately)
- The question mark was confusing next to the x icon on cinny. I guess it is clear though that it is stuff where we have no data. I wonder if it could be also maybe have greyed out text or something to make it more obvious that it is missing data rather than missing feature.

Still happy to accept it as is.

HarHarLinks:
> * Surprised about the lab icon on the Threads on Element X (but also didnt check. Maybe schildi next was further. I only used that lately)

Last I checked you could not start threads on EXA, at least not without enabling some labs feature

> * The question mark was confusing next to the x icon on cinny. I guess it is clear though that it is stuff where we have no data. I wonder if it could be also maybe have greyed out text or something to make it more obvious that it is missing data rather than missing feature.

#3462

PR #3476, opened 2026-06-18
[codex] Fix hero overlap in mobile landscape

starkmarkus:
Prevents the front-page hero curves from overlapping the heading on mobile landscape viewports by relaxing the hero height and hiding the decorative curves in that orientation.\n\nThis keeps the hero readable on short screens without changing the desktop layout.


HarHarLinks:
Rejecting this PR on basis of our contribution policy (see above)

PR #3490, opened 2026-06-25
Trust and Safety Research and Documentation WG updates

andybalaam:
### Description

Two changes to the T&S R&D WG entry:

1. Add one new member
2. Set me as the chair, and remove Cat as vice-chair.

Both Niko and Cat are stepping down, and I have their support to take over.

Authority for this comes from this poll: https://matrix.discourse.group/t/chair-for-the-t-s-r-d-wg/391 and the support of Nico, our governing board sponsor.

### Role

I am contributing this as the newly-elected chair of the working group.

### Signoff

Commits contain signoffs.


HarHarLinks (approved):
As governing board member with visibility on the poll, lgtm

Issue #3488, opened 2026-06-23
Add Beeper Line bridge


HarHarLinks:
### Project Affected

n/a

### Current Information

missing

### Suggested Correction

add to index

### Source

https://github.com/beeper/line

### Additional Context

it was just announced going prod at https://blog.beeper.com/2026/06/23/beeper-june/

clins1994:
@HarHarLinks hi 👋 i'd like to work on this one so please assign it to me 😄

HarHarLinks:
We don't assign issues here, but please check our contribution guidelines and submit a PR referencing this issue

clins1994:
@HarHarLinks got it. thanks 👍

clins1994:
pr submitted and ready for review: https://github.com/matrix-org/matrix.org/pull/3498

clins1994:
@HarHarLinks can you review it for me? 🥺

MTRNord:
#3498 is now merged so this is done :)

clins1994:
@MTRNord cheers 🙌🏼

PR #3498, opened 2026-06-27
add beeper's line messenger bridge to the ecosystem bridge list

clins1994:
### Description

I've added the [beeper-line](https://github.com/beeper/line) bridge to the LINE bridge listing

I opted to leave the existing obsolete matrix-puppeteer-line entry. Let me know if it is a better idea to prune it, I can patch it in as well. If not, this is ready 🚢 

Screenshot 2026-06-27 at 21 44 07

### Related issues

- https://github.com/matrix-org/matrix.org/issues/3488
- https://github.com/beeper/line/issues/193

### Role

I contributed individually to the line bridge this year. I'm not affiliated with beeper or line.

### Timeline

No deadline

### Signoff

My commit is signed off

### DIFF
--- a/content/ecosystem/bridges/line/bridges.toml
+++ b/content/ecosystem/bridges/line/bridges.toml
@@ -1,4 +1,33 @@
 [[bridges]]
+name = "beeper-line"
+maintainer = "beeper, highesttt"
+summary = """
+A bridgev2 project for bridging Matrix and [LINE](https://www.line.me/en). Written in Go using [mautrix-go](https://github.com/mautrix/go) and intended for use with [Beeper](https://www.beeper.com).
+"""
+maturity = "Stable"
+language = "Go"
+license = "MIT"
+docs = "https://github.com/beeper/line#setup"
+repo = "https://github.com/beeper/line"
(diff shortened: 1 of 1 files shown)


MTRNord (approved):
Thanks for the PR and sorry that it slipped through the cracks!

I can't verify the supported features due to lack of line myself. So trusting the submission :)

Everything else seems to check out. :)

---

_For the others on the WG:_ I think featured here is justified considering the quality we know from beeper (and previously maunium) bridges. So vouching for that here :D (and we don't have another active bridge here so it probably doesn't make much of a difference here anyway)

---

regarding matrix-puppeteer-line we so far only marked things as obsolete rather than removing them. Ideally we should improve the UI for it a bit to hide it like we do in the other ecosystem categories I guess. The bridge pages UI makes it a bit harder than on the other pages where we have filters hiding them by default :) Will take that thought to https://github.com/matrix-org/matrix.org/pull/3518 to see if I can maybe fix that for the future :)

clins1994:
@MTRNord all good ! thank you 🪴

PR #3510, opened 2026-07-04
Fix client filters

### LINKED ISSUE #3509: *Features* filter does not work on `ecosystem/clients/`
On Linux Mint 22.3 Cinnamon Firefox.

If I remember correctly it used to work fine.

Personal notes: [Benjamin_Loison/matrix.org/issues/4](https://codeberg.org/Benjamin_Loison/matrix.org/issues/4)

weeman1337:
### Description

The client data was restructured in
686c08bb02bafb51b63ed952172b3f340953a8c6.
Update the filter CSS to match the new structure.

### Related issues

closes #3509

### Role

Website WG

### DIFF
--- a/templates/macros/clients.html
+++ b/templates/macros/clients.html
@@ -115,8 +115,10 @@ maturity-{{ client.maturity | lower }}
 
 licence-{{ client.licence | slugify }}
 
-{% for feature_name, feature_supported in client.features %}
-{% if feature_supported %}feature-{{ feature_name | replace(from="_", to="-") }}{% endif %}
+{% for feature_level_name, features in client.features %}
+    {% for feature_name, feature_state in features %}
+        {% if feature_state == "supported" %}feature-{{ feature_name | replace(from="_", to="-") }}{% endif %}
+    {% endfor %}
 {% endfor %}
 {% endmacro classes %}
 


MTRNord (approved):
Seems to fix the immediate issue.

I just wonder where Widgets are coming from as no client actually seems to advertise support it and its not anymore on the feature list(s) in the client details. But I think that doesnt block this PR.

HarHarLinks:
> Seems to fix the immediate issue.

It completely mismatches the actual list of features when you open a client though?

HarHarLinks:
https://github.com/matrix-org/matrix.org/blob/054aba632cc8e679c24283c5e1d5f7d2bbfce7d2/templates/ecosystem/clients.html#L155-L186

MTRNord:
> > Seems to fix the immediate issue.
> 
> It completely mismatches the actual list of features when you open a client though?

Huh? Going through the filters and clicking a random client did lead me to the filtered for thing. Thats what I meant there. I did notice that it lists widgets which isnt on the other list though. But it fixed filtering for the rest?

weeman1337:
I think I have decrypted what @HarHarLinks means. The filter list is not in sync with the date in the client files. I would say this can be a follow-up issue.

This PR fixes the broken filter.

HarHarLinks:
Is a fix a fix if afterwards the behavior is still broken?

weeman1337:
> Is a fix a fix if afterwards the behavior is still broken?

It is less broken than before. At least the existing filters on the client side are working again.

weeman1337:
https://github.com/matrix-org/matrix.org/issues/3512

PR #3534, opened 2026-07-14
Improve /support and /foundation/members designs

MTRNord:
### Description

More of an experiment trying to align it with all the other changes I proposed. Not sure if it is actually the right direction or not. For sure needs further feedback. I would love a proper masonry layout for example but I seem to not have css support for that :/

### Related issues

I guess #2198 

### Role

Website & Content WG

### Timeline

Lets take our time on this one and discuss before thinking about timelines :) 

### Signoff


MTRNord:
~~Cloudflare doesnt like to comment I guess: https://mtrnord-members-support-rede.matrix-website.pages.dev/~~

See https://github.com/matrix-org/matrix.org/pull/3534#issuecomment-4970822020

HarHarLinks:
maybe /support should also have the downward arrow ui hint from /try-matrix?

image
iirc the membership contracts guarantees platinum to have a bigger logo on the page than gold than silver

MTRNord:
> maybe /support should also have the downward arrow ui hint from /try-matrix?

Hm good idea

> image iirc the membership contracts guarantees platinum to have a bigger logo on the page than gold than silver

Hm ok I forgot about that. It just looked weird if i kept the original sizes in this setup :thinking: since it became extremely noticeable (though plat and gold were the same size in the old design it seems? and only the rest was smaller than those 2)

Not yet sure what to do about that. Hm.

HarHarLinks:
> though plat and gold were the same size in the old design it seems? and only the rest was smaller than those 2

then that was probably correct

templates/membership.html
```diff
@@ -35,7 +39,8 @@ <h3>Individual</h3>
                     <li>Free <a href="/homeserver/pricing">premium account</a> on Matrix.org</li>
                 </ul>
             </div>
-            <div id="most-popular" class="tier-card">
+            <div id="most-popular" class="tier-card tier-card--group-start tier-card--silver">
+                <p class="tier-card-eyebrow">For organisations</p>
```
HarHarLinks, line 43: have you tried how this would look following this approach?
image

HarHarLinks:
https://github.com/matrix-org/website-wg/issues/17 is also a thing

sass/_support.scss
```diff
@@ -1,37 +1,10 @@
...
-}
+// /support/ now reuses the shared .step-flow layout (_try-matrix.scss)
+// instead of its own grid. The rest of this file (.membership tiers,
+// #supporters wall) belongs to /foundation/members/ (membership.html).
+//
+// Breadcrumb nav (macros/breadcrumb.html) is styled by the shared
+// _breadcrumb.scss partial - see there for the .page-header (dark) context
+// both pages use.
 
 .membership {
```
HarHarLinks, line 9: the membership part of the style should perhaps split into a new members.scss

static/js/supporters-masonry.js
HarHarLinks, line 1: members-masonry.js?

sass/_try-matrix.scss
HarHarLinks, line 1: maybe the .step-flow $.colored should use #eee of the current support page rather than #fafafa? the latter is indistinguishable from white to me

templates/support.html
```diff
@@ -53,17 +74,19 @@ <h2>Make a donation</h2>
...
     </div>
 
-    <div id="working-group">
-        <div class="content">
-            <h2>Join a working group</h2>
-            <div>
+    <div class="step-background" id="working-group">
+        <div class="step no-image">
+            <div class="instructions">
+                <h2>Join a working group</h2>
```
HarHarLinks, line 89:
```suggestion
                <h2>Join a Working Group</h2>
```

style guide defines this spelling in the meantime. pls also update the text!!

sass/_base.scss
```diff
@@ -18,6 +18,31 @@
...
+    // darker/warmer so the two don't read as the same grey. These plain
+    // colours are for the (thin, non-text) frame border only; the flag is a
+    // solid fill with white text, which needs its own darker "-badge" pair
+    // to clear AA (the plain colours are too light/desaturated for that).
+    --color-tier-platinum: #b8c4cc;
+    --color-tier-platinum-badge: #5b6b74;
+    --color-tier-gold: #c9a227;
+    --color-tier-gold-badge: #8a6d1c;
+    --color-tier-silver: #7c7f83;
+    --color-tier-silver-badge: #5c5f63;
```
HarHarLinks, line 45: maybe check if you can reuse the precise colors or at least inspiration from the member badges on /branding and or check in with jordan

templates/membership.html
HarHarLinks: but it would need to stay how it currently works or an entirely new solution for the mobile view

HarHarLinks:
requesting review from andy to confirm whatever we do here matches the membership contract requirements

sass/_support.scss
MTRNord: Done

static/js/members-masonry.js
MTRNord: done

templates/support.html
MTRNord: fixed

templates/membership.html
MTRNord: Didnt consider this at all yet.

sass/_try-matrix.scss
MTRNord: I can but it breaks try-matrix for me 
grafik

So maybe a followup pr instead?

sass/_base.scss
MTRNord: switched to the badge colors now. I hope thats better :D

templates/support.html
```diff
@@ -1,22 +1,44 @@
...
-            <div>
+<div class="step-flow">
+    <div class="step-background quickstart">
+        <div class="step no-image">
+            <div class="instructions quickstart-instructions">
+                <h2>Ready to help right now?</h2>
+                <p>
+                    A donation is the fastest way to make a difference today.
+                    Prefer a different way to pitch in? The options below cover
+                    membership, working groups, and merch too.
```
HarHarLinks:
```suggestion
                    membership, Working Groups, and merch too.
```

templates/support.html
```diff
@@ -53,19 +77,21 @@ <h2>Make a donation</h2>
...
     </div>
 
-    <div id="working-group">
-        <div class="content">
-            <h2>Join a working group</h2>
-            <div>
+    <div class="step-background" id="working-group">
+        <div class="step no-image">
+            <div class="instructions">
+                <h2>Join a working Group</h2>
```
HarHarLinks:
```suggestion
                <h2>Join a Working Group</h2>
```

templates/support.html
MTRNord: my grep failed me apparently /o\

HarHarLinks:
How about?
image

But that still leaves the "missing" button for ecosystem and associate members or their perks

MTRNord:
> How about? image
> 
> But that still leaves the "missing" button for ecosystem and associate members or their perks

I think for the missing sections/cards I kinda need to know the rules. Then we can probably open a second row or something.

HarHarLinks:
Screenshot_20260817-223534
A bit tight on mobile. The hero itself also gets a bit in the way.

But the oddest feeling part is missing visual separation between the sections (flip the bgcolor to start at uneven?). The title alignments being different is more obvious on mobile too. Maybe that's ok?
Screenshot_20260817-223921

sass/_support.scss
```diff
@@ -1,161 +1,160 @@
...
+// Membership contracts specify Platinum logos must render larger than
+// Gold, which must render larger than Silver - a graduated step, not just
+// a flat "top 3 tiers are bigger" split.
+#platinum-supporters .supporters-card {
+    --supporters-logo-size: 190px;
+}
+
+#gold-supporters .supporters-card {
+    --supporters-logo-size: 165px;
+}
```
HarHarLinks: > > though plat and gold were the same size in the old design it seems? and only the rest was smaller than those 2
> 
> then that was probably correct

sass/_membership.scss
```diff
@@ -0,0 +1,149 @@
...
+    .tiers-row {
+        display: grid;
+        grid-template-columns: repeat(4, 15rem);
+        gap: 1rem;
+        // These 4 columns are wider than the viewport on most screens, so
+        // this scrolls sideways (the "scroll to most-popular" script in
+        // membership.html relies on scrollLeft); it was previously
+        // overflow-y, which didn't let that horizontal scroll happen at all.
+        overflow-x: auto;
+        scroll-snap-type: x proximity;
```
HarHarLinks, line 122: on mobile, the initial view on this places the individual members cleanly out of view to the left so the chance to discover it is near 0.

while it could help i'm not sure if snapping to the middle is the solution. it might require some thinking outside the box to solve this properly.

sass/_support.scss
```diff
@@ -1,161 +1,160 @@
...
+        // Every tile fills its column - the flex-basis weighting above only
+        // applies to the un-enhanced single-row layout.
+        .supporters-section {
+            flex: none;
+            width: 100%;
         }
     }
 
-    .representatives {
+    .supporters-section {
```
HarHarLinks, line 111: I wish the frames for each section would be sticky on mobile or something similar so i can tell more easily where i am

sass/_support.scss
andypiper: I think the language about logo sizes no longer exists in the "v2" membership agreement, but I've asked for clarification in case I misunderstood something. Just to say that I'm looking into this.

sass/_support.scss
MTRNord: its now sticky on desktop and mobile :)

sass/_membership.scss
MTRNord: I added scroll indicators now. I think that solves it.

sass/_support.scss
MTRNord: fixed

MTRNord:
> A bit tight on mobile. The hero itself also gets a bit in the way.

Maybe better now?

> But the oddest feeling part is missing visual separation between the sections (flip the bgcolor to start at uneven?). The title alignments being different is more obvious on mobile too. Maybe that's ok?

I flipped it now. Possibly a bit better. I think the title is fine as the first one is somewhat of a CTA and the rest are section headings?

MTRNord:
> How about? image
> 
> But that still leaves the "missing" button for ecosystem and associate members or their perks

I moved the buttons to more sensible places now I think. As said before I need more info for the perks of the other things for them to be added I guess.

sass/_support.scss
HarHarLinks: neat!

templates/membership.html
```diff
@@ -5,11 +5,67 @@
...
+        // native anchor jump scrolls the *page* to bring the target into
+        // view vertically too - not what we want here, we only want the
+        // row itself to scroll sideways. With JS available, do that scroll
+        // manually instead and skip the native jump.
+        document.querySelectorAll('.tiers-scroll-hint').forEach((hint) => {
+            hint.addEventListener('click', (e) => {
+                let target = document.querySelector(hint.getAttribute('href'));
+                if (!target) return;
+                e.preventDefault();
+                tiersRow.scrollTo({ left: target.offsetLeft - 16, behavior: 'smooth' });
```
HarHarLinks, line 66: https://github.com/user-attachments/assets/e7028885-43ac-4396-b9b7-0f1e04ab3c3c

shouldn't this scroll 1 by one instead of to the end?

sass/_base.scss
```diff
@@ -18,6 +18,31 @@
...
+    // Kept distinct from the older --borders-color/--color-text-light pair
+    // above rather than merged into them, since those are already used
+    // with their own (slightly different) shades elsewhere.
+    --color-border: #d2d2d2;
+    --color-text-muted: #555;
+    --color-bg-tint: #fafafa;
+    --card-radius: 16px;
+    --card-radius-sm: 8px;
+
+    // The top 3 "become a member" tiers are literally named after materials
```
HarHarLinks, line 32: we have reference colors for these and all other tiers at https://github.com/matrix-org/matrix.org/tree/main/static/branding/membership

content/foundation/members/members.toml
```diff
@@ -22,6 +22,7 @@ logo = "connect2x.png"
 name = "DINUM"
 website = "https://www.numerique.gouv.fr/"
 logo = "dinum.svg"
+white_bg = true
```
HarHarLinks, line 25: what is the purpose of this?

templates/membership.html
MTRNord: hm I guess thats possible too :thinking:

sass/_base.scss
MTRNord: That was fixed earlier already. Comment needs fixing/removal though.

content/foundation/members/members.toml
MTRNord: It activates some mixing css for those images that have a white background. Which makes that white bg disappear.

I think its not super noticeable anymore now that we dont have a slightly colored bg on the cards anymore (we did have hover bg at some point. But dont anymore. and there it would have been a white box with grey bg. with the mix that is added it did remove that white box and applied the bg as you would expect)

I think its still sensible to keep for people using dark mode plugins that change the background. It just adds `mix-blend-mode: multiply;` when the image isnt an isolated logo without background/transparency in the image.

HarHarLinks: > activates some mixing css

I couldn't find that

sass/_base.scss
HarHarLinks: i don't see a green ecosystem members frame nor purple associates

content/foundation/members/members.toml
MTRNord: it should be the flag that adds or removes the `white-bg-logo` class on images. I see it in the deployment so I am pretty sure its doing its job

HarHarLinks: I have finally found the class in support.scss

sass/_membership.scss
```diff
@@ -0,0 +1,249 @@
+.membership {
```
weeman1337, line 1: And another huuuge bunch of specific CSS. We should better think more in a framework direction for CTAs, Grids, etc. in my opinion. But we can clean this up later.

weeman1337 (commented):
The buttons should be CTAs = Call To Action. The current proposal does not contain a „call“. They can simple be in the box they belong to 🤷  See the attached screenshot with a sketch. We can think about to find a shorter text for the buttons, or at least for the join as individual button. If you want a bit more variety in the buttons, they can get the tier colour for example. 

I also think, that scrolling is really bad, because it hides content. Maybe this can be done for showcases etc. where the content is not important. For the membership I would say it should be a grid that breaks.

Also shouldn't there a link from the support page to the membership page?

Image

andypiper:
re: request for my review. 

The documentation on membership tier benefits, list logo placement perks as: Platinum large logo (front page + sponsor page), Gold medium logo (sponsor page), Silver small logo (sponsor page).

(where "sponsor page" is as written in prospectus; we can interpret as members page)

So, we must maintain those agreements and arrangements with any improvements.

Technically the existing page has both Platinum and Gold tiers with the same sized logos (and neither of the two members have commented on that!), and the revisions proposed do not change those sizes, so from that perspective this seems OK to me.

HarHarLinks:
> Platinum large logo (front page + sponsor page), Gold medium logo (sponsor page), Silver small logo (sponsor page).

if this is indeed the case (or any other updates in that are), somebody (you?) will need to task the WG with it

andypiper:
As I said - I think the update in this PR reflects the current site status 👍🏻

sass/_base.scss
MTRNord: fixed

PR #3518, opened 2026-07-09
Ecosystem Improvements Part 2

MTRNord:
### Description

This essentially is a followup on #3172 and #3239 which makes the ecosystem much closer to the try-matrix design language.

Its a massive redesign in comparison to #3172 hence it is separate.

### Role

Individual or Website & Content WG

### Timeline

Any time but hopefully less than 6 months ;)

### Signoff

Please [sign off](https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md) your individual commits.

PR #3536, opened 2026-07-15
Add a glossary to the Docs

### LINKED ISSUE #2262: Move https://github.com/matrix-org/glossary into docs?
It might make sense to move https://github.com/matrix-org/glossary (and the repo it mentions) inside the docs matrix.org page instead.
It seems like a logical place people would be looking for something like this outside of spec itself.

(cc @turt2live anything from your or I guess SCT side against doing this?)

MTRNord:
### Description

This fixes #2262 while hopefully not ending up in another unmaintained FAQ :)

grafik

### Related issues

#2262

(also mildly related to #3534 and #3518 since this is cherry picked from the same local branch I have for all this work)

### Role

Website & Content WG

### Timeline

Whenever

### Signoff

### DIFF
--- a/content/branding.md
+++ b/content/branding.md
@@ -42,4 +42,3 @@ The following table lists some terms that either are wordmarks of The Matrix.org
 | other organs | | Other organs of the Foundation are similarly capitalised, e.g. "Guardians", "Spec Core Team", "Trust & Safety". |
 | `#room:server.tld` | | Pronounced "[hash room on server dot tld](https://matrix.org/blog/wp-content/uploads/2014/09/Introducing-matrix.pdf)". |
 | `@user:server.tld` | | Pronounced "[at user on server dot tld](https://matrix.org/blog/wp-content/uploads/2014/09/Introducing-matrix.pdf)". |
-| [vodozemac](https://github.com/matrix-org/vodozemac) | | Named after [vodozémac](https://hjp.znanje.hr/index.php?show=search_by_id&id=f19vXxZ%2F), the Croatian word for "amphibian". Pronounced "[vo-do-ze-mats](https://www.youtube.com/watch?v=mjndO734kco&t=38)". |
--- a/content/docs/glossary/terms.toml
+++ b/content/docs/glossary/terms.toml
@@ -1,158 +1,134 @@
 # Glossary entries shown on /docs/glossary/.
 #
-# To add a term, append a new [[terms]] block below (alphabetical order isn't
-# required - the page sorts them itself). Only `term` and `definition` are
-# required.
+# Terms are grouped into arrays-of-tables named terms.<category>, one per
+# category. To add a term, append a new block under the right category below
+# (or start a new terms.<category> for a new one - the page picks it up and
+# labels/sorts it automatically). Only `term` and `definition` are required.
 #
-#   [[terms]]
+#   [[terms.core]]
 #   term = "Your term"
-#   category = "One of the categories used below, or a new one"
 #   aliases = ["Optional alternate names or abbreviations"]
 #   definition = "One or two plain-language sentences."
 #   learn_more = "/docs/some/relevant/page/#anchor"   # optional, internal or external
 
-[[terms]]
+[[terms.core]]
 term = "Homeserver"
-category = "Core concepts"
 aliases = ["HS"]
 definition = "A server that hosts Matrix user accounts and participates in the network. Every account belongs to exactly one homeserver, identified by that server's domain."
 learn_more = "/docs/matrix-concepts/elements-of-matrix/#homeserver"
 
-[[terms]]
+[[terms.core]]
 term = "Client"
-category = "Core concepts"
 definition = "An application people use to read and send messages on Matrix, such as Element. Clients talk to a homeserver, never directly to other clients."
 learn_more = "/docs/matrix-concepts/elements-of-matrix/#client"
 
-[[terms]]
+[[terms.core]]
 term = "Federation"
-category = "Core concepts"
 definition = "The process by which independently-run homeservers exchange room data with each other, so users on different servers can share rooms without depending on a central server."
 learn_more = "/docs/matrix-concepts/elements-of-matrix/"
 
-[[terms]]
+[[terms.core]]
 term = "Matrix ID"
-category = "Core concepts"
 aliases = ["MXID", "User ID"]
 definition = "A user's identifier, made up of a local part and the domain of their homeserver, written as `@username:example.com`. Pronounced \"at username on example dot com\"."
 learn_more = "/docs/matrix-concepts/elements-of-matrix/#homeserver"
 
-[[terms]]
+[[terms.core]]
 term = "Room"
-category = "Core concepts"
 definition = "A conversation, shared between the homeservers of everyone taking part. Each participating server keeps its own local copy of the room, which is kept in sync with the others."
 learn_more = "/docs/matrix-concepts/rooms_and_events/"
 
-[[terms]]
+[[terms.core]]
 term = "Room ID"
-category = "Core concepts"
 definition = "The internal, permanent identifier of a room, written as `!opaqueid:example.com`. Unlike a room's published address, it never changes for the lifetime of the room."
 learn_more = "/docs/matrix-concepts/rooms_and_events/"
 
-[[terms]]
+[[terms.core]]
 term = "Event"
-category = "Core concepts"
 definition = "A single timestamped, signed item in a room's history: a message, a membership change, a room setting being updated, and so on. Rooms are built up from a directed acyclic graph of events."
 learn_more = "/docs/matrix-concepts/rooms_and_events/"
 
-[[terms]]
+[[terms.core]]
 term = "State event"
-category = "Core concepts"
 definition = "An event that represents a piece of persistent room state (such as the room name, topic, or a member's membership) rather than a one-off message. Only the latest state event for a given type and key is considered current."
 learn_more = "/docs/matrix-concepts/rooms_and_events/"
 
-[[terms]]
+[[terms.core]]
 term = "Power level"
-category = "Core concepts"
 aliases = ["PL"]
 definition = "A per-user or per-action number in a room that determines what a user is allowed to do, such as sending certain event types, kicking, or banning."
 learn_more = "/docs/matrix-concepts/rooms_and_events/#administration-privileges"
 
-[[terms]]
+[[terms.core]]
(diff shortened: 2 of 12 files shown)


navigation.toml
```diff
@@ -28,9 +28,14 @@ title = "Categories"
 href = "/category/"
 
 [[header]]
+id = "docs"
 title = "Docs"
 href = "/docs/"
 
+[[header.children]]
```
HarHarLinks, line 35: spicy, at this point we should probably add all the docs sections (except legacy?)

templates/docs/glossary.html
HarHarLinks, line 1: a bit strange that this breaks with the UX of the rest of docs

templates/docs/glossary.html
```diff
@@ -0,0 +1,58 @@
...
+    <nav class="glossary-jump" aria-label="Jump to category">
+        {% for category, terms in grouped %}
+        <a href="#{{ category | slugify }}">{{ category }}</a>
+        {% endfor %}
+    </nav>
+
+    <div class="glossary-list">
+        {% for category, terms in grouped %}
+        <section class="glossary-category" id="{{ category | slugify }}">
+            <h2>{{ category }}</h2>
```
HarHarLinks: can you make the sections more visually obvious?

templates/docs/glossary.html
```diff
@@ -0,0 +1,58 @@
...
+            <dl>
+                {% for term in terms %}
+                <div class="glossary-term">
```
HarHarLinks: would probably be a great feature to be able to deeplink to each term

templates/docs/glossary.html
```diff
@@ -0,0 +1,58 @@
+{% extends "skel.html" %}
+{% import "macros/breadcrumb.html" as breadcrumb %}
+{% block head_extra %}
+<meta name="description" content="{{ section.extra.meta_description }}">
+{% endblock head_extra %}
+{% block content %}
+{% set data = load_data(path="content/docs/glossary/terms.toml") %}
+{% set grouped = data.terms | sort(attribute="term") | group_by(attribute="category") %}
```
HarHarLinks: ordering of the categories? bit surprising that T&S is on top.

content/docs/glossary/terms.toml
```diff
@@ -0,0 +1,158 @@
...
+#   [[terms]]
+#   term = "Your term"
+#   category = "One of the categories used below, or a new one"
+#   aliases = ["Optional alternate names or abbreviations"]
+#   definition = "One or two plain-language sentences."
+#   learn_more = "/docs/some/relevant/page/#anchor"   # optional, internal or external
+
+[[terms]]
+term = "Homeserver"
+category = "Core concepts"
```
HarHarLinks: why is the toml flattened instead of following the actual structure, i.e. `[[terms.core]]`, etc

content/docs/glossary/terms.toml
```diff
@@ -0,0 +1,158 @@
...
+aliases = ["GB"]
+definition = "The organ of The Matrix.org Foundation through which the ecosystem participates in its governance, made up of representatives elected or appointed from across the community."
+learn_more = "/foundation/governing-board/"
+
+[[terms]]
+term = "Spec Core Team"
+category = "Governance"
+aliases = ["SCT"]
+definition = "The team responsible for stewarding the Matrix specification itself, including reviewing and merging Matrix Spec Changes (MSCs)."
+
```
HarHarLinks, line 135:
```suggestion
definition = "The team responsible for stewarding the Matrix specification itself, including reviewing and merging Matrix Spec Changes (MSCs)."
learn_more = "/foundation/about/#the-spec-core-team"

```

content/docs/glossary/terms.toml
```diff
@@ -0,0 +1,158 @@
...
+[[terms]]
+term = "Federation"
+category = "Core concepts"
+definition = "The process by which independently-run homeservers exchange room data with each other, so users on different servers can share rooms without depending on a central server."
+learn_more = "/docs/matrix-concepts/elements-of-matrix/"
+
+[[terms]]
+term = "Matrix ID"
+category = "Core concepts"
+aliases = ["MXID", "User ID"]
```
HarHarLinks:
```suggestion
term = "Matrix user ID"
category = "Core concepts"
aliases = ["Matrix ID", "MXID", "User ID"]
```

content/docs/glossary/terms.toml
```diff
@@ -0,0 +1,158 @@
...
+definition = "A user's identifier, made up of a local part and the domain of their homeserver, written as `@username:example.com`. Pronounced \"at username on example dot com\"."
```
HarHarLinks, line 33: can get removed from branding then

content/docs/glossary/terms.toml
HarHarLinks, line 1: olm and megolm are mentioned under e2ee, maybe that is ok.

migrate vodozemac from /branding to here. maybe enough to mention libolm as part of that, being deprecated and all?

templates/docs/glossary.html
HarHarLinks, line 1: /branding should probably inherit this layout. can be a followup pr

MTRNord: It just doesnt feel like a true doc page for me I think. Its more of an independent page than a documentation guide. But maybe I am wrong in that.

MTRNord: Added the sidebar again now.

templates/docs/glossary.html
MTRNord: I tried. Not sure if its really a change though. Not sure what would help.

sass/_glossary.scss
```diff
@@ -0,0 +1,175 @@
...
+        font-size: 1.05rem;
+    }
+
+    dd {
+        margin: 0.35rem 0 0;
+        color: var(--color-text-muted);
+    }
+
+    // Only reveal the "#" permalink on hover/focus - it's a bonus for
+    // sighted mouse/keyboard users linking to a term, not something that
```
HarHarLinks, line 151: rip mobile users

templates/docs/glossary.html
```diff
@@ -0,0 +1,68 @@
...
+                {% for term in data.terms[key] | sort(attribute="term") %}
+                {% set term_slug = term.term | slugify %}
+                {% set term_id = "term-" ~ term_slug %}
+                <div class="glossary-term" id="{{ term_id }}">
+                    <dt>
+                        {{ term.term }}
+                        {% if term.aliases %}
+                        <span class="glossary-aliases">({{ term.aliases | join(sep=", ") }})</span>
+                        {% endif %}
+                        <a href="#{{ term_id }}" class="glossary-term-anchor" aria-label="Link to {{ term.term }}">#</a>
```
HarHarLinks, line 53: this is a problem everywhere, but jumping to this anchor needs an offset considering the navbar etc:
image

MTRNord: uhhhh it does have one :thinking:

Screensizes strike again...

content/branding.md
```diff
@@ -42,4 +42,3 @@ The following table lists some terms that either are wordmarks of The Matrix.org
 | other organs | | Other organs of the Foundation are similarly capitalised, e.g. "Guardians", "Spec Core Team", "Trust & Safety". |
 | `#room:server.tld` | | Pronounced "[hash room on server dot tld](https://matrix.org/blog/wp-content/uploads/2014/09/Introducing-matrix.pdf)". |
 | `@user:server.tld` | | Pronounced "[at user on server dot tld](https://matrix.org/blog/wp-content/uploads/2014/09/Introducing-matrix.pdf)". |
```
HarHarLinks, line 44: MXID is already transferred, room alias can be too?

sass/_glossary.scss
```diff
@@ -0,0 +1,175 @@
...
+
+    h2 {
+        position: sticky;
+        // Stick just below the fixed site header, not at the very top of the
+        // viewport, otherwise it's hidden behind the header and the first
+        // term ends up looking like the top of the section instead.
+        top: var(--navbar-height);
+        margin: 0;
+        padding: 0.5rem 0;
+        background-color: var(--color-bg-tint);
```
HarHarLinks, line 110: hm no i'm not convinced. that's not what i was looking for :/

content/docs/glossary/terms.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[[terms.core]]
+term = "Application Service"
+aliases = ["AppService"]
+definition = "A privileged piece of software, registered with a homeserver, that can puppet many virtual users and rooms at once. Bridges and some advanced bots are built as Application Services."
+learn_more = "https://spec.matrix.org/latest/application-service-api/"
+
+[[terms.core]]
+term = "Bridge"
+definition = "A form of Application Service that relays messages between Matrix and a third-party chat platform (for example IRC, Slack, or Discord), so users on either side can talk to each other."
```
HarHarLinks, line 70: oooh i think we have definitions of more details like puppeting and portalling and stuff somewhere. probably terms that are worth appearing here even if the details are only behind a link

HarHarLinks: i think bridges being appservices is strictly speaking also incorrect. a bridge is a component that unpacks one message completely, incl possible decryption, and then sends it to another network. a protocol converter is a component that only converts the protocol around a message without need to decrypt. something slide 25 https://docs.google.com/presentation/d/1hxRlyJ6SD3zXb2MPGmfigB0s9Yj4dsWBMtctNPv06IA/edit?usp=sharing (yay obscure documents)

sass/_glossary.scss
```diff
@@ -0,0 +1,175 @@
...
+        border-bottom: 1px solid var(--color-border);
+        font-size: 1.1rem;
```
HarHarLinks: ?

sass/_glossary.scss
HarHarLinks: maybe just reveal this always while in mobile view. make sure the touch area is big enough.

templates/docs/glossary.html
HarHarLinks: i could imagine trying to scroll it to the middle would be a valid workaround? if that is even possible.

it could also be an idea to put a highlighting frame something around the currently selected term.

content/docs/glossary/terms.toml
HarHarLinks: We also currently only put chat bridges into /ecosystem/bridges. Other definitions would include e.g. hookshot in bridges, which we keep under integrations.

PR #3545, opened 2026-07-21
Add two new T&S R&D WG members

andybalaam:
### Description

Add two new T&S R&D WG members

### Role

As chair of the WG.


andybalaam:
(James Bot's discourse invite is still in progress but they are on the way)

HarHarLinks (approved):
Speaking as Website & Content WG, lgtm

Issue #3135, opened 2026-01-05
Enhancement: Introduce REUSE for the license declaration of code (and other) files


awtj8o81ryywg793:
Since it appears that we are increasingly using REUSE to declare the licensing of image files (see #2706), I think we should use REUSE to do the same for files that contain code too. Thoughts?

HarHarLinks:
yeah, this is the current proposal to solve the deeper issue of https://github.com/matrix-org/matrix.org/issues/2706#issuecomment-2994527921. I would basically call it decided in favor of reuse tooling at this point, just the licenses for some files remain to be clarified. I think it would be okay to start implementing REUSE bit by bit.

awtj8o81ryywg793:
Quoted from a [previous comment](https://github.com/matrix-org/matrix.org/issues/3135#issuecomment-3711705774):

> yeah, this is the current proposal to solve the deeper issue of [#2706 (comment)](https://github.com/matrix-org/matrix.org/issues/2706#issuecomment-2994527921). I would basically call it decided in favor of reuse tooling at this point, just the licenses for some files remain to be clarified. I think it would be okay to start implementing REUSE bit by bit.

Brilliant. I fully support this decision. I will start implementing REUSE in my PRs going forward.

HarHarLinks:
Let's reopen and document the decision to at least adopt reuse gradually on new files before calling it done. Ideally we can add a CI that enforces the reuse linter on files added or touched in a PR.

PR #3566, opened 2026-07-29
Refresh Famedly logo to updated branding on members page.

andypiper:
### Description

Replaced the Famedly logo with an updated version that matches their current branding.

### Related issues

n/a

### Role

Matrix.org Foundation

### Timeline

When possible 🙏🏻 

### Signoff

👍🏻


HarHarLinks (commented):
lgtm, but please check if we can add license info via https://github.com/matrix-org/matrix.org/issues/3135

andypiper:
Is the suggestion to add the license info into the file itself (presumably as a comment in the SVG), or would we do this via another path?

HarHarLinks:
we have been using the .license file approach for svgs so far, e.g. https://github.com/matrix-org/matrix.org/blob/main/static/images/element-creators-of-matrix--light.svg.license

we need to get wording for this use case from legal, unless we have it already somewhere i forgot, and add a license file for that, similar to https://github.com/matrix-org/matrix.org/blob/main/LICENSES/LicenseRef-ElementLogo.txt

HarHarLinks (approved):
I'm somewhat surprised this is a license specifically for famedly rather than members in general, but technically works for me.

PR #3583, opened 2026-08-10
harden: the cohort-3d in cohort-3d.html

anupamme:
## Summary
Harden input handling in `static/~matthew/cohorts/cohort-3d.html` (flagged by multi_agent_ai).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `static/~matthew/cohorts/cohort-3d.html:22` |
| **Assessment** | Defensive hardening |

**Description**: The cohort-3d.html page reads the URL query string directly and concatenates it into a fetch() URL without validation. An attacker can supply arbitrary path segments to influence which resource is fetched, potentially accessing unintended files on the same origin.

## Changes
- `static/~matthew/cohorts/cohort-3d.html`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/static/~matthew/cohorts/cohort-3d.html
+++ b/static/~matthew/cohorts/cohort-3d.html
@@ -20,7 +20,8 @@ body {
 
     // Called when the Visualization API is loaded.
     function drawVisualization() {
-        var query = window.location.search.substring(1) || "ios";
+        var rawQuery = window.location.search.substring(1) || "ios";
+        var query = /^[a-zA-Z0-9_-]+$/.test(rawQuery) ? rawQuery : "ios";
         fetch("./cohort-3d-" + query + ".csv").then((res)=>{
             res.text().then((body)=>{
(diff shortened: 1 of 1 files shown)


MTRNord:
Rejecting on basis of our AI Policy in https://github.com/matrix-org/matrix.org/blob/main/CONTRIBUTING.md#ai-policy

Issue #3434, opened 2026-05-25
Document criteria for adding, updating, and removing ecosystem projects


awtj8o81ryywg793:
Given that we’ve decided on various aspects of better documenting the criteria needed to have a project added, updated, or removed from the website in two separate meetings (23/07/26 and 04/06/26), let’s do the work to put those decisions into action. See the listed sub-issues for each particular action or area of concern.

The criteria documented here are meant to be additive or corrective to the current requirements, rather than be a complete replacement.

> [!IMPORTANT]
> The criteria for this policy are very much a work in progress. If you’re looking to provide feedback, please remember that this comment is a living document. Quote the specific section you want to discuss, as it may be edited or removed.
>
> Whilst all parts of the criteria list are open to feedback, those labelled as **To-do** are especially so.

## Adding a Project

While other contributors, other than a project’s maintainer(s), can submit a request to add a project, it’s advised that maintainers do so, as they’re in a better position to answer any questions the Website & Content Working Group may have.

### Criteria:
- Submitter(s) must be able to evidence that the project is maintained (e.g., updating dependencies, bug fixes, new functionality, etc.).
  - Projects must exist for a minimal amount of time before being eligible to be listed.
    - **To-do:** Decide how long a project should exist before we deem it eligible to be listed. Perhaps three months?
- Submitter(s) must demonstrate relevance or value of the project to the wider Matrix community.
- Submitter(s) must include screenshots or videos to demonstrate that claims regarding support for certain functionality are correct.
- Submitter(s) must justify why they think the chosen maturity level (e.g., `Stable`, `Beta`, `Alpha`, or  `Obsolete`) is appropriate.
- Submitter(s) must include a `.licence` file alongside the project’s logo, which includes the information required for REUSE compliance.
  - Licence identifiers other than those documented on SPDX are allowed.
  - A copy of the licence should be made available in `LICENSES/`, if a copy is not already available.

#### Forked Projects:

- Submitter(s) must demonstrate significant changes to the folk of the original project.
  - This is because we want to avoid listing low-effort re-skins that don't contribute much to the overall ecosystem, and may soon become abandoned.
- Submitter(s) must ensure that the project has a separate identity (e.g., a different name, logo, etc.) from the project they forked from.

### Criteria:

- Submitter(s) must include screenshots or videos to demonstrate that claims regarding support for certain functionality are correct.
[...]

PR #3635, opened 2026-09-13
Add matrix-kdeconnect-sms to the SMS bridges

by-architect:
### Description

Adds `matrix-kdeconnect-sms` to `content/ecosystem/bridges/sms/bridges.toml`.

It bridges SMS through [KDE Connect](https://kdeconnect.kde.org/), using a phone that is already paired rather than a modem, a carrier API or a second handset. One room per SMS thread, with a ghost user per phone number. TypeScript, built on `matrix-appservice-bridge`, Apache-2.0.

Repo: https://github.com/by-architect/SmsKdeConnectMatrixBridge

Marked **Alpha** deliberately: it is new and has so far only run against a single homeserver.

The `supports` flags are set to what the code does today rather than what is planned:

- `mms = false` — attachments are decoded off the wire but not yet uploaded to Matrix
- `group_chats = false` — group threads currently collapse to the first address
- reactions, replies, typing notifications and read receipts are unimplemented

Two notes for review:

- I used the same `[bridges.supports]` keys as the existing entries in this file (`sms`, `mms`, `rcs`, `group_chats`, `history_from_android`) rather than the generic set in `CONTENT.md`, since the template notes that these vary by network. Happy to switch if you would rather they were uniform.
- No `room` key, as there is no support room for it yet. Every other entry in the file has one — glad to create one and add it if that is expected for listing.

### Related issues

None.

### Role

Contributing as an individual. I am the author and maintainer of the bridge being listed, so this is a self-submission.

### Timeline

No deadline — review whenever it suits the WG.

### Signoff

The single commit is signed off under the DCO, and the DCO check is passing.


awtj8o81ryywg793:
Whilst we appreciate your contribution, I’m afraid that, given the project’s limited history, we’re likely to wait at least three months before considering whether to list it, as we want to ensure that we are not listing throwaway projects.

We have yet to document this aspect of the listing criteria, but we have ongoing work aimed at doing so. See: https://github.com/matrix-org/matrix.org/issues/3434

by-architect:
Waiting for your decisions then.

PR #3638, opened 2026-09-14
Add matrix-google-classroom to the integrations

by-architect:
### Description

Adds `matrix-google-classroom` to `content/ecosystem/integrations/integrations.toml`.

It posts Google Classroom announcements, assignments and materials into a Matrix room per course, each from a ghost user standing for its author. TypeScript, built on `matrix-appservice-bridge`, Apache-2.0.

Repo: https://github.com/by-architect/GoogleClassroomMatrixBridge

**Listed as an integration rather than a bridge, deliberately.** The Google Classroom API exposes no comments endpoint at all — stream comments and private comments, which are the conversation in Classroom, are not reachable through it. So this is strictly one-way: nothing said in a course arrives beyond the posted stream, and nothing typed in Matrix goes back. Calling it a bridge would promise a round trip that cannot exist.

It is new, and currently running against one homeserver and two live courses.

### Related issues

None.

### Role

Contributing as an individual. I am the author and maintainer of the project being listed, so this is a self-submission.

### Timeline

No deadline — whenever suits the WG.

### Signoff

The single commit is signed off under the DCO.

No `room` key: there is no support room for it yet. I note that 19 of the 27 existing entries carry one and 8 do not, so I have followed the latter — happy to add one if it is expected.


awtj8o81ryywg793:
Whilst we appreciate your contribution, I’m afraid that, given the project’s limited history, we’re likely to wait at least three months before considering whether to list it, as we want to ensure that we are not listing throwaway projects.

We have yet to document this aspect of the listing criteria, but we have ongoing work aimed at doing so. See: https://github.com/matrix-org/matrix.org/issues/3434

PR #3640, opened 2026-09-17
Add a way to display events more prominently in TWIM posts

### LINKED ISSUE #3639: Consider adding a calendar table like WeeklyOSM to TWIM posts
It might be nice to have a continuous table of upcoming events listed on the TWIM posts. This would be an ADDITION to the existing twim posts announcing them.

See https://weeklyosm.eu/#wn842_upcoming_events for what I mean.

I believe they explicitly don't do recurring events but rather do every entry as a single entry. That probably works for us as well for most.

It likely also should be anchored to the post's publish date so it doesn't get confusing with old TWIM editions when we rebuild the page the week after.

It also likely should be some global file to allow a) to maybe have a calendar somewhere in a dedicated page b) to track far away events in every TWIM until it happens without needing to set it up every week and c) it might allow more easy creation of some ics url. Though to me ics support would be something we can also add later.

MTRNord:
### Description

Fixes #3639

### Related issues

#3639

### Role

Website & Content WG Hat

### Timeline

N/A

### Signoff

Commits are signed off.


MTRNord:
Branch Preview URL:	https://mtrnord-events-calendar.matrix-website.pages.dev/blog/2026/09/11/this-week-in-matrix-2026-09-11/

(scroll to bottom)

content/events/events.toml
```diff
@@ -0,0 +1,36 @@
...
+url = "https://conference.matrix.org/"
+date = "2026-10-20"
+end_date = "2026-10-23"
+location = "Malmö, Sweden"
+
+[[event]]
+name = "Matrix Stammtisch (CCCAC)"
+url = "https://matrix.to/#/#matrix-stammtisch:ccc.ac"
+date = "2026-09-14"
+location = "CCCAC"
```
HarHarLinks, line 17: repeating this manually for the second tuesday of every month is a lot of redundancy

content/events/events.toml
```diff
@@ -0,0 +1,36 @@
...
+name = "Matrix Stammtisch Karlsruhe"
+url = "https://matrix.to/#/#stammtisch-karlsruhe:alsace.hair"
+date = "2026-09-15"
+location = "Fachschaft ETIT, Karlsruhe"
+
+[[event]]
+name = "Matrix Stammtisch Freiburg"
+url = "https://blog.freiburg.social/event/matrix-stammtisch-freiburg-6/"
+date = "2026-09-17"
+location = "Blauer Fuchs, Freiburg"
```
HarHarLinks, line 29: this one however is only pseudo-regular and tends to get fresh URLs at least sometimes

content/events/events.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[[event]]
+name = "Matrix Stammtisch Dresden"
+url = "https://dresden.datenpunks.de/events/matrix/2026-09-20-matrix-stammtisch/"
+date = "2026-09-19"
+location = "Datenspuren, Dresden"
```
HarHarLinks, line 35: and this one is even less regular, including the venue changing

HarHarLinks (commented):
current data source: https://matrix-community.events/more.html

how would this list be maintained/submitted for TWIM? how to inform the TWIM room readers?

content/events/events.toml
MTRNord: I kinda oriented myself at https://osmcal.org/documentation/ which states:

> ### Why is there no support for repeated/periodic events?
> 
> Because we don't want to list events that potentially do not exist anymore because their initial creators forgot to remove them. Our alternative is the 'Repeat Event' functionality which you can reach from the event's detail page and allows you to copy an existing event.

I do see your point though. But also ics repetition stuff is quite complex. Like you got the whole vacation, calendar, timezone and whatever shenanigans to think about once you want repeating events. Which gets complex really fast.

MTRNord:
> how would this list be maintained/submitted for TWIM? how to inform the TWIM room readers?

The current assumption is pretty pragmatic: Its just added by the author when someone does an events post. The normal events posts we have wouldnt get removed. this is more meant as an addition and an overview. Possibly with an extra page eventually(?). I still would like to see TWIM entries too to spike engagement.

PR #3641, opened 2026-09-17
improve image size in blog posts and add lightbox support to them.

MTRNord:
### Description

This is an attempt to make twim images "flow" better. I am not entirely sure if this is ideal yet. But basically it groups adjacent images into 2 rows now, adds a lightbox (if js is available. Doesnt work in pure static mode sadly) and adds a bit of basic styling.

### Related issues

N/A

### Role

Website & Content WG

### Timeline

Whenever

### Signoff

Done in commits


HarHarLinks:
this PR seems to have more comments than code. without looking, this smells like an issue in itself.

MTRNord:
> this PR seems to have more comments than code. without looking, this smells like an issue in itself.

Cleaned that up.

PR #3642, opened 2026-09-17
Add a TWIM callout

MTRNord:
### Description

We learned that TWIG does this as well and maybe this helps to get more engagement with TWIM in the future and an easier way for people to find the twim-guide :) 

### Related issues

N/A

### Role

Website & Content WG

### Timeline

N/A

### Signoff


MTRNord:
Branch Preview URL:	https://mtrnord-twim-cta.matrix-website.pages.dev/

templates/post.html
```diff
@@ -72,27 +73,26 @@ <h1>{{ page.title }}</h1>
...
+                plausible_event="BlogPostCtaSupportUs"
+            ) }}
         </div>
         {% if page.toc %}
         <aside>
+            {% if page.taxonomies.category and "This Week in Matrix" in page.taxonomies.category %}
+            {{ callout::callout(
+                heading="Got Matrix news?",
+                body="<p>Share your project, release or event in next week's digest.</p>",
+                cta_text="Send your notes",
```
awtj8o81ryywg793: Given that the [TWIM guide](https://matrix.org/twim-guide/) (currently) states that:

> Copying the list of your commit messages of this week is easy, but not what TWIM is about. **Ideally, we prefer posts which are a few paragraphs long outlining some highlights over a long detailed list.** This makes your announcement more digestable to humans. You may of course still link to the changelog.

I’m not sure whether "notes" would be the best term to use here. Maybe it is fine, but I feel that it suggests short, more changelog-like posts rather than something more substantial, as suggested by the guide. I wonder whether something like "Send your update" or "Share an update" might work better? I’m not sure. Notes are not always short, so perhaps it is fine after all.

templates/taxonomy_single.html
```diff
@@ -19,6 +20,17 @@ <h2>
             {% endif %}
         </header>
 
+        {% if term.name == "This Week in Matrix" %}
+        {{ callout::callout(
+            heading="Got Matrix news?",
+            body="<p>Share your project, release or event in next week's digest.</p>",
+            cta_text="Send your notes",
```
awtj8o81ryywg793: See: https://github.com/matrix-org/matrix.org/pull/3642/changes#r4039769900

HarHarLinks:
the wording here would also benefit the homepage to be more engaging
image

templates/post.html
MTRNord: Replaced it in [cde6a9a](https://github.com/matrix-org/matrix.org/pull/3642/commits/cde6a9a4bced08fbeb2714e95543b507ceb1f557)

MTRNord:
> the wording here would also benefit the homepage to be more engaging image

Updated it to match the other button Text. Happy for more ideas though beyond that if needed.

